PR #18, opened 2018-08-10
fix: Remove 72 chars limit on body description 🔥

kosssi:
We would like limit max length of each line, not all body part
We have to wait  to be introduced to commitlint

[See this PR](https://github.com/marionebl/commitlint/pull/400)

### DIFF
--- a/packages/commitlint-config/index.spec.js
+++ b/packages/commitlint-config/index.spec.js
@@ -2,8 +2,8 @@ import lint from '@commitlint/lint'
 import cozyConfig from '.'
 
 const rules = cozyConfig.rules
-const TOO_LONG =
-  'This is an invalid sentence since it has too much character !!!!!!!!!!!!!'
+// const TOO_LONG =
+//  'This is an invalid sentence since it has too much character !!!!!!!!!!!!!'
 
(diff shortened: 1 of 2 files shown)


packages/commitlint-config/index.spec.js
```diff
@@ -2,8 +2,8 @@ import lint from '@commitlint/lint'
 import cozyConfig from '.'
 
 const rules = cozyConfig.rules
-const TOO_LONG =
-  'This is an invalid sentence since it has too much character !!!!!!!!!!!!!'
+// const TOO_LONG =
+//  'This is an invalid sentence since it has too much character !!!!!!!!!!!!!'
```
enguerran: I would suggest not commenting on the code because we never know why it was commented on.

packages/commitlint-config/index.spec.js
```diff
@@ -69,4 +70,5 @@ describe('Commitlint Config Cozy', () => {
       }
     })
   })
+  */
```
enguerran: I would suggest deleting the code block instead of commenting on it because we never know why it is commented on.

PR #21, opened 2018-08-16
Realtime: Add `domain` and `secure` to config

gregorylegarec:
~~BREAKING CHANGE: `url` is no more used in config and must be replaced by `domain` and `secure` attributes~~

~~Two informations was passed with the `url` attribute: the security protocol and the cozy domain. But they're was a flaw with url like `http://cozy.tools:8080`. The unsecure protocol was detected correctly, but the domain was not parsed and has `undefined` value.~~

Eventually, this PR just adresses the issue caused by a bad RegExp matching.

### DIFF
--- a/packages/realtime/src/index.js
+++ b/packages/realtime/src/index.js
@@ -49,10 +49,12 @@ function getWebsocketProtocol(url) {
 }
 
 function getDomainFromUrl(url) {
-  const domainRegexp = new RegExp(`^https?:/{2}(.*)$`)
-  const m = url.match(domainRegexp)
-  if (m) {
-    return m[1]
+  try {
+    return new URL(url).host
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.warn(`Cannot get domain from URL : ${error.message}`)
+    return null
   }
 }
 
@@ -68,6 +70,26 @@ function keepAlive(socket, interval, message) {
   return socket
 }
 
+const isRequired = [attr => !!attr, 'is required']
+const isString = [str => typeof str === 'string', 'should be a string']
+
+const validate = types => obj => {
+  for (const [attr, rules] of Object.entries(types)) {
+    for (const [validator, message] of rules) {
+      if (!validator(obj[attr])) {
+        throw new Error(`${attr} ${message}.`)
+      }
+    }
+  }
+}
+
+const configTypes = {
+  token: [isRequired, isString],
+  url: [isRequired, isString]
+}
+
+const validateConfig = validate(configTypes)
+
 async function connectWebSocket(
   config,
   onmessage,
@@ -75,6 +97,7 @@ async function connectWebSocket(
   numRetries,
   retryDelay
 ) {
+  validateConfig(config)
   return new Promise((resolve, reject) => {
     const protocol = getWebsocketProtocol(config.url)
     const domain = getDomainFromUrl(config.url)


gregorylegarec:
[[STO] Initialisation du realtime échoue](https://trello.com/c/j5oi8c1J/1171-sto-initialisation-du-realtime-%C3%A9choue)

kosssi:
@gregorylegarec By default I prefer that it's in secure mode

packages/realtime/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cozy-realtime",
-  "version": "1.1.2",
+  "version": "2.0.0",
```
kosssi: You do not have to change the version numbers ;)

ptbrowne:
> The unsecure protocol was detected correctly, but the domain was not parsed and has undefined value.

Why the API change to resolve for this bug ? Why not parse the domain correctly ? We are accustomed to use the URL property to configure libraries communicating with Cozy.

gregorylegarec:
@ptbrowne The thing is we discussed with @kosssi and agreed to update the way realtime was initialized, following my proposition.

Why changing ? For at least two reasons:
* The stack send us `domain` and `token`, I think we should keep them unchanged, having to add a `http(s)` protocol to the domain is pretty much useless, especially for the realtime where this protocol is just tested with a regexp to use `ws(s)` instead (Actually, that's the current regexp who fails to match the domain an causes the current bug.
* In my opinion, it is always a code smell to pass more than one information in a string, and it's exactly what we are doing here: by adding a protocol to the `domain` given by the stack to make the `url` variable, we are juste telling our library : "Hey, look at this protocol to determine if you have to use SSL and yes also this the domain". I would prefer to make everything explicit, so that's why I would prefer to add a `secure` property.

After your comment I discussed with @kosssi and we agreed it is eventually too soon to make all those changes, so those two points should be adressed in a dedicated issue about how are we initializing all ou libraries. I will create it in a few minutes.

So for now I will update this PR to just fix the behaviour which make the realtime unusable in a local environment.

ptbrowne:
Thanks Greg for your detailed explanation. I understand better now. Cool that the breaking change will be done in another issue and that first the bug will be fixed.

Concerning the API : I think we should not have to pass `secure`. This makes the API larger on the application side and it is a burden on the user to decide whether to use `secure` or not. The underlying library can choose whether or not to do SSL based on the `window.location.protocol`.

Concerning the first bug : Parsing URLs with a RegExp is a bad idea, it has already bitten us once for the intents when the stack started to put hash fragments in its URLs. The URL API seems to be a good choice here.

image

gregorylegarec:
@ptbrowne Totally agree with an internal use of `window.location.protocal`, that was how the first version of the realtime lib was proceeding back in the days. BUT I have a mix of both `window.location.protocol` and `secure` attribute in my mind, but I will detail it later to discuss about it.

I thank about the `URL` API but I think it is not available in Node for desktop and mobile apps using realtime.

ptbrowne:
The URL API is available in NodeJS 8 : https://nodejs.org/docs/latest-v8.x/api/url.html

gregorylegarec:
Nice, so let's go.

gregorylegarec:
@ptbrowne Ok code has been updated. I added a type check for `url` and `token` parameter because I have been stuck an hour because of an Object token (`cozy.client._token` being an object).

packages/realtime/src/index.js
```diff
@@ -75,6 +77,15 @@ async function connectWebSocket(
...
   retryDelay
 ) {
+  const expectedStringAttributes = ['domain', 'token']
+  expectedStringAttributes.forEach(attribute => {
+    if (typeof config[attribute] !== 'string') {
+      throw new Error(
+        `Configuration attribute ${attribute} should be a string.`
+      )
+    }
+  })
```
ptbrowne: Minor: for this kind of code I usually go in a PropTypes manner.

```js
const isString = [
  str => typeof str === 'string',
  'should be a string'
]
const configTypes = {
  domain: isString,
  token: isString
}
const validate = config => obj => {
	for (const [attr, [validator, message]] of Object.items(configTypes)) {
      if (!validator(obj[attr])) {
        throw new Error('Attribute ${attribute} ${message}')
      }
    }
    return true
}
const validateConfig = validate(configTypes)
```

gregorylegarec: I used this pattern to validate, thanks. I let you merge if it is ok.

PR #22, opened 2018-08-17
Add watch task for realtime

gregorylegarec:
Simple watch using only babel, based on build task.


kosssi:
@gregorylegarec it's better to add --watch on build command:
`"watch": "yarn run build --watch",`

gregorylegarec:
@kosssi Done!

PR #25, opened 2018-08-22
feat: add doctypes

ptbrowne:
This brings the code from cozy-doctypes for working with doctypes
that are shared between several codebases here.

### DIFF
--- a/packages/doctypes/package.json
+++ b/packages/doctypes/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "cozy-doctypes",
+  "version": "1.0.1-beta.6",
+  "description": "Definitions for Cozy models",
+  "main": "src/index.js",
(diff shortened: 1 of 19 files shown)


packages/doctypes/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "cozy-doctypes",
+  "version": "1.0.1-beta.6",
```
kosssi: On ne passerait pas en version stable ?

ptbrowne: done

packages/doctypes/src/Document.js
```diff
@@ -0,0 +1,208 @@
+const { omit, pick } = require('lodash')
```
kosssi: On importe finalement tout `lodash` ici

ptbrowne: done

PR #31, opened 2018-08-28
Better options check

gregorylegarec:
The main idea is to be clearer when an options like `appType` or `appSlug` is missing.

This PR adds a dedicated error message for attributes retrieved from manifest file.

Also it checks that the appType has the correct value (`webapp` or `konnector`).

### DIFF
--- a/packages/cozy-app-publish/lib/prepublish.js
+++ b/packages/cozy-app-publish/lib/prepublish.js
@@ -30,23 +30,43 @@ const sanitize = options => {
   }
 }
 
+const isRequiredFromManifest = manifestAttribute => [
+  field => typeof field !== 'undefined',
+  () => `Property ${manifestAttribute} must be defined in manifest.`
+]
+
+const isRequired = [
+  field => typeof field !== 'undefined',
+  key => `Option ${key} is required.`
+]
+
+const isOneOf = values => [
+  field => values.includes(field),
+  key => `${key} should be one of the following values: ${values.join(', ')}`
+]
+
+const optionsTypes = {
+  appBuildUrl: [isRequired],
+  appSlug: [isRequiredFromManifest('slug')],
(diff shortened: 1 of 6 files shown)


packages/cozy-app-publish/lib/prepublish.js
```diff
@@ -30,23 +30,43 @@ const sanitize = options => {
...
+
+const isOneOf = values => [
+  field => values.includes(field),
+  key => `${key} should be one of the following values: ${values.join(', ')}`
+]
+
+const optionsTypes = {
+  appBuildUrl: [isRequired],
+  appSlug: [isRequiredFromManifest('slug')],
+  appType: [isRequiredFromManifest('type'), isOneOf(['webapp', 'konnector'])],
```
CPatchane: ❤️

packages/cozy-app-publish/lib/prepublish.js
```diff
@@ -30,23 +30,43 @@ const sanitize = options => {
...
-  ].forEach(option => {
-    if (typeof options[option] === 'undefined') {
-      throw new Error(`${option} cannot be undefined`)
-    }
-  })
+  for (const option in optionsTypes) {
+    const validators = optionsTypes[option]
+    validators.forEach(validator => {
+      if (!validator[0](options[option])) {
+        throw new Error(validator[1](option))
```
y-lohse: Just curious why you used an array here rather than an object with named properties? I feel like `validator.test` and `validator.error` or something might be more explicit?

gregorylegarec: I was also wondering myself. Truth is @ptbrowne suggested me this kind of code a few days ago, with arrays and I kept this logic.

Also, as I made this kind of code two times in a week, I suggest that we could create some kind of `cozy-validators` lib.

PR #34, opened 2018-08-29
Try/catch publish()

gregorylegarec:
The call to publish() was not encapsulated in a try/catch, generating warning about unhandled Promise rejections.


gregorylegarec:
Fixed by https://github.com/cozy/cozy-libs/pull/50

PR #36, opened 2018-08-29
chore: Remove eslint auto fix

ptbrowne:
The auto fix can send Travis + Lerna into a loop if some files are
commited to master with linting errors. Since the --fix does not warn
that there are linting errors, Travis will report green even in the
Pull request : linting errors can be integrated into
master without any warning.

Since --fix changes files, Lerna thinks something has changed and
publishes the application. The publishing of the application will
result in a commit, which will in turn launch a new Travis job
and the whole story starts again.


gregorylegarec:
Very interesting bug !

PR #41, opened 2018-09-10
feat: Create cozy-app renovate config

y-lohse:
Once published to npm, this package should allow all our apps to use or at least extend the same renovate configuration.
[Here are the docs](https://renovatebot.com/docs/config-presets/) about how this works.

The configuration itself is very basic right now — it just prevents travis to be swamped during working hours. We'll be happy to consider more individual rules once the first version is published.

### DIFF
--- a/packages/renovate-config-cozy-app/package.json
+++ b/packages/renovate-config-cozy-app/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "renovate-config-cozy-app",
+  "version": "0.0.1",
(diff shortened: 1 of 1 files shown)


y-lohse:
[Créer une config renovate commune à toutes les apps cozy](https://trello.com/c/JDFGOMML/106-cr%C3%A9er-une-config-renovate-commune-%C3%A0-toutes-les-apps-cozy)

y-lohse:
Let's do this!

PR #53, opened 2018-09-19
Use shift in cozy-release only if remaining args

CPatchane:
Travis shell throws errors if we use `shift` without remaining arguments.
This PR check before using it.


gregorylegarec:
Thanks @CPatchane 💇

PR #58, opened 2018-09-20
Update shebang to use bash

CPatchane:
Without that, Travis CI will use the `dash` shell from ubuntu. 
So we can force using `bash` with this shebang


gregorylegarec:
Well spotted @CPatchane 👌 (ping @doubleface)

PR #61, opened 2018-09-21
Change semver parsing

gregorylegarec:
The previous command does not work well in Ubuntu and it makes Travis fail.

### DIFF
--- a/packages/cozy-release/scripts/cozy-release.sh
+++ b/packages/cozy-release/scripts/cozy-release.sh
@@ -59,10 +59,9 @@ read_current_version() {
 
 compute_next_version() {
   type=${2:-minor}
-  IFS='.' read -r -a array <<< "$1"
-  major=${array[0]}
-  minor=${array[1]}
-  patch=${array[2]}
+  major=`echo $1 | awk -F '.' '{print $1}'`
+  minor=`echo $1 | awk -F '.' '{print $2}'`
+  patch=`echo $1 | awk -F '.' '{print $3}'`
 
   case "$type" in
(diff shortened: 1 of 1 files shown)


CPatchane:
Thanks for that! We also have to change at [line 77](https://github.com/cozy/cozy-libs/blob/master/packages/cozy-release/scripts/cozy-release.sh#L77)

packages/cozy-release/scripts/cozy-release.sh
```diff
@@ -59,10 +59,11 @@ read_current_version() {
...
   type=${2:-minor}
-  IFS='.' read -r -a array <<< "$1"
-  major=${array[0]}
-  minor=${array[1]}
-  patch=${array[2]}
+  # from https://github.com/cloudflare/semver_bash/blob/master/semver.sh
+  local RE='[^0-9]*\([0-9]*\)[.]\([0-9]*\)[.]\([0-9]*\)\([0-9A-Za-z-]*\)'
+  major=`echo $1 | sed -e "s#$RE#\1#"`
+  minor=`echo $1 | sed -e "s#$RE#\2#"`
+  patch=`echo $1 | sed -e "s#$RE#\3#"`
```
CPatchane: What about:
```
major=`echo $1 | awk -F '.' '{print $1}'
minor=`echo $1 | awk -F '.' '{print $2}'
patch=`echo $1 | awk -F '.' '{print $3}'
```

No need to have a regular expression now I think.

gregorylegarec: Great, thanks !

Issue #42, opened 2018-09-11
[meta] Unifying folder names


y-lohse:
In this repo, [the names of the folders for packages](https://github.com/cozy/cozy-libs/tree/master/packages) sometime have a `cozy-` prefix, and sometimes not. Let's pick one solution.

Vote on this comment with:

- 🎉 if you **don't want** a prefix (eg. `realtime`)
- ❤️ if you **want** a prefix (eg. `cozy-realtime`)

Feel free to argue in the comments. Vote closes in a couple of days.

y-lohse:
As far as I can tell, this change will have no functional impact, the package names on npm will remain the same, etc. So it's just esthetics, and I don't really care one way or the other, but I feel like we have enough `cozy-` prefixes everywhere as it stands, and this one doesn't bring any benefits.

CPatchane:
In general, I prefer things being explicit so `NPM slug/name = folder name` as personal preference

gregorylegarec:
Same opnion as @CPatchane, it's clearer to find a npm package under its real name.

edas:
Well, NPM is migrating with the `@scope/name` instead of `scope-name`

y-lohse:
Sorry that's a separate issue :p we're not talking about npm naming here, just the folder names in the repo.

y-lohse:
We add prefixes everywhere! Thanks!

PR #86, opened 2018-09-25
Use same package name as NPM name

CPatchane:
As decided [here](https://github.com/cozy/cozy-libs/issues/42) we now use the same package name as the matched NPM name.


y-lohse:
Thank you!

PR #88, opened 2018-09-25
[commitlint] Raise max commit message length

y-lohse:
This issue brought up in a meeting recently, but I'm formally proposing we raise the limit for commit message titles to 72 chars instead of 50.

From what I can see (in [this article](https://chris.beams.io/posts/git-commit/#limit-50) and in various git tools):

- 50 character is the desired length
- 72 is the hard limit after which things gets truncated

Since `commitlint` is rejecting commits that don't match the configuration it's enforcing a *hard* limit, and should be 72 and not 50. But we as humans should still try to shoot for 50.

Does that sound ok? Especially pinging people who mentioned this issue @CPatchane @kosssi @enguerran 

edit: Don't mind Travis, I'll update the tests if/when we get consensus on this.

### DIFF
--- a/packages/commitlint-config-cozy/index.js
+++ b/packages/commitlint-config-cozy/index.js
@@ -3,7 +3,7 @@ module.exports = {
     'body-leading-blank': [1, 'always'],
     'body-max-line-length': [2, 'always', 72],
     'footer-leading-blank': [1, 'always'],
-    'header-max-length': [2, 'always', 50],
+    'header-max-length': [2, 'always', 72],
     'scope-case': [0, 'always', 'lower-case'],
     'subject-case': [
       2,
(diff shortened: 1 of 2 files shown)


ptbrowne:
> what is the always option for?
what is the 2 option for?

> Level [0..2]: 0 disables the rule. For 1 it will be considered a warning for 2 an error.
Applicable always|never: never inverts the rule.
Value: value to use for this rule.

From: https://marionebl.github.io/commitlint/#/reference-rules

I wonder why they did not simply put "warning" and "error" for the levels.


OK for the change !

y-lohse:
Alright, then I think we can have a warning at 50 and a an error at 72! I'll make the changes.

y-lohse:
Well sadly we can only have one configuration for a rule, so I put the hard limit at 72. Please keep them under 50 anyway <3

kosssi:
thanks @y-lohse

PR #102, opened 2018-10-02
Update BankTransaction::reconciliate algorithm

drazik:
We had a duplicates bug in some specific cases. We decided to use the same algorithm whichever the transactions are fetched by a « old » connector, or a newly configured one.

So now we use the algorithm based on the date everytime:

* We get the date of the most recent transaction already saved
* We create all the newly fetched transactions that are after this date
* For each newly fetched transaction that is before this date, we check if some of them don't exist in the DB yet (by comparing their `{ amount, originalLabel, date }` triplet)

This way, we don't need to know if the connector is a new one or not, we always do the same work.

Before merging this PR, I think we must try it on a real world case, even if I added some unit test to add more confidence.

### DIFF
--- a/packages/cozy-doctypes/src/BankTransaction.js
+++ b/packages/cozy-doctypes/src/BankTransaction.js
@@ -67,11 +67,11 @@ class Transaction extends Document {
 
   /**
    * Get the descriptive (and almost uniq) identifier of a transaction
-   * @param {object} transaction - The transaction (containing at least amount, label and date)
+   * @param {object} transaction - The transaction (containing at least amount, originalLabel and date)
    * @returns {object}
    */
   getIdentifier() {
-    return `${this.amount}-${this.label}-${this.date}`
+    return `${this.amount}-${this.originalLabel}-${this.date}`
   }
 
   /**
--- a/packages/cozy-doctypes/src/BankTransaction.js
+++ b/packages/cozy-doctypes/src/BankTransaction.js
@@ -47,11 +47,16 @@ class Transaction extends Document {
     }
   }
 
-  static reconciliate(remoteTransactions, localTransactions, options) {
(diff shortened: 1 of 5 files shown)


packages/cozy-doctypes/src/BankTransaction.js
```diff
@@ -44,29 +47,110 @@ class Transaction extends Document {
...
+    }
+  }
+
+  /**
+   * Get the descriptive (and almost uniq) identifier of a transaction
+   * @param {object} transaction - The transaction (containing at least amount, originalLabel and date)
+   * @returns {object}
+   */
+  getIdentifier() {
+    return `${this.amount}-${this.originalLabel}-${this.date}`
```
kosssi: je ne sais pas si ça peut poser un souci mais ton identifier peut commencer par un "-"

packages/cozy-doctypes/src/BankTransaction.js
```diff
@@ -44,29 +47,110 @@ class Transaction extends Document {
...
+
+  /**
+   * Get transactions that should be present in the stack but are not
+   * The transactions are checked using triplet {label, amount, date}
+   * @param {array} transactionsToCheck
+   * @param {array} stackTransactions
+   * @returns {array}
+   */
+  static getMissedTransactions(transactionsToCheck, stackTransactions) {
+    const missedTransactions = transactionsToCheck
```
kosssi: const identifiers = stackTransactions.map(t => getIdentiers(t))

packages/cozy-doctypes/src/BankTransaction.js
drazik: C'est une string quelconque qui sera comparée à une string aussi, donc je pense pas que ça puisse poser un soucis.

PR #103, opened 2018-10-03
cozy-release: Fix git pull

gregorylegarec:
Syntax was wrong. I did not detect this because I was using the globally installed yarn version. Sorry about that.


CPatchane:
Maybe we could add `node bin/cozy-release.js` as test script of the `cozy-release` package, at least to run the script in Travis during CI, what do you think?

PR #106, opened 2018-10-09
[renovate] Handling version ranges

y-lohse:
Regardless of what's currently in the PR, how do we want to handle version ranges?

[Here are our options](https://renovatebot.com/docs/configuration-options/#rangestrategy). Explain what you prefer and why, and we'll implement whatever makes the most sense.

### DIFF
--- a/packages/renovate-config-cozy-app/package.json
+++ b/packages/renovate-config-cozy-app/package.json
@@ -7,6 +7,7 @@
       "extends": [
         "config:base"
       ],
+      "rangeStrategy": "pin",
       "packageRules": [
         {
           "excludePackageNames": [


y-lohse:
In drive/photo/contacts, we've been pinning dependencies and it has worked well for us.

The upside of pinned dependencies is that everyone gets the same code output every time. With version ranges, you maybe have a different version on your local machine than what travis will use, which can lead to hard to spot problems.

The downside is that we get a PR for every single new version of a dependency, even minor changes. This can be mitigated with scheduling, and a combination of tests and automerges.

So I'd be in favor of pinning versions, but the `auto` option of renovate also seems sensible.

y-lohse:
[Créer une config renovate commune à toutes les apps cozy](https://trello.com/c/JDFGOMML/106-cr%C3%A9er-une-config-renovate-commune-%C3%A0-toutes-les-apps-cozy)

PR #112, opened 2018-10-11
Add new options to Transaction::reconciliate

drazik:
Add `onCatchupNonExistentTransaction`, `onCatchupMoreNew` and `onCatchupLessNew` so we can track the following events in Piwik in the Linxo connector:

* We fetch a transaction that is before the date split, but it doesn't exist in the Cozy
* We fetch a transaction that is before the date split, but for the same `{ amount, date, label }` triplet we have less transactions in the cozy than we have fetched
* We fetch a transaction that is before the date split, but for the same `{ amount, date, label }` triplet we have more transactions in the cozy than we have fetched

### DIFF
--- a/packages/cozy-doctypes/src/BankTransaction.js
+++ b/packages/cozy-doctypes/src/BankTransaction.js
@@ -157,7 +157,7 @@ class Transaction extends Document {
       )
 
       if (missedTransactions.length > 0) {
-        log('info', `Saving ${missedTransactions}.length`)
+        log('info', `Saving ${missedTransactions.length}`)
       } else {
         log('info', 'No missed transactions to catch-up')
       }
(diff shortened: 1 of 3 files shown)


packages/cozy-doctypes/src/BankTransaction.js
```diff
@@ -99,14 +104,30 @@ class Transaction extends Document {
...
+        }
+
         continue
       }
+
+      if (newTransactions.length < existingTransactions.length) {
+        if (typeof options.onCatchupLessNew === 'function') {
+          options.onCatchupLessNew()
+        }
+      }
```
ptbrowne: I dont' find the names of the functions very telling :/ 

Shouldn't `onCatchupLessNew` and `onCatchupMoreNew` have `Transaction` also ?

* Maybe the callbacks could take the current identifier ?

Catchup is not the right verb I think (Catch up == rattraper son retard): `onFetchTransaction(newTransactions, existingTransactions)` and then it is the responsibility of the caller to discriminate and send the right piwik events

ptbrowne: Cool :) `onMissedTransactionFound` seems clearer too :)

PR #114, opened 2018-10-11
Add --yes option

gregorylegarec:
--no-confirm option allows to bypass prompt by automatically confirming actions.

It should be use with care, in scripting context for example.


doubleface:
It works well but I personnaly find the --yes option more convential

PR #141, opened 2018-11-21
Commitlint docs

kosssi:
This PR add husky on dependency, it's already essential for commitlint and update documentation

### DIFF
--- a/packages/commitlint-config-cozy/README.md
+++ b/packages/commitlint-config-cozy/README.md
@@ -14,15 +14,16 @@
 ```json
 "commitlint": {
   "extends": [
-    "./packages/commitlint-config"
+    "cozy"
   ]
 }
 ```
-- Add verification during a commit on your `package.json` ([see an example](https://github.com/cozy/cozy-banks/blob/85572b6827cdaa45c1ed44d6922829ba6480b3c9/package.json#L65)):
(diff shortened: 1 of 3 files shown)


packages/commitlint-config-cozy/README.md
```diff
@@ -14,15 +14,16 @@
 ```json
 "commitlint": {
   "extends": [
-    "./packages/commitlint-config"
+    "cozy"
   ]
 }
 ```
-- Add verification during a commit on your `package.json` ([see an example](https://github.com/cozy/cozy-banks/blob/85572b6827cdaa45c1ed44d6922829ba6480b3c9/package.json#L65)):
+- Add verification during a commit with husky (1.2.0) on your `package.json` ([see an example](https://github.com/cozy/cozy-libs/blob/ea325a4ea2b5bf0067875f625b5ad0a5b320e7e9/package.json#L24-L28)):
```
drazik: Is there a reason for specifying the husky version?

kosssi: yes with 0.14.3 is not the same configuration ;)

drazik: OK so I propose this, since this configuration format is available since v1.0.0:

```suggestion
- Add verification during a commit with husky (> 1.0.0) on your `package.json` ([see an example](https://github.com/cozy/cozy-libs/blob/ea325a4ea2b5bf0067875f625b5ad0a5b320e7e9/package.json#L24-L28)):
```

PR #150, opened 2018-11-28
feat(Harvest): init cozy-harvest

sizvix:
First step to initialise cozy-harvest-lib

### DIFF
--- a/packages/cozy-harvest/README.md
+++ b/packages/cozy-harvest/README.md
@@ -1,0 +1,15 @@
+Harvest
+=====
+
+Cozy-Harvest will bring the logic, modules and components used in harvest applications.
+
+Harvest applications will be apps which will allow to configure konnectors, setup authentification informations and check for synchronized data (And much more in the future).
+Each harvest application will be associated with only one konnector.
+
+This first version is just an "Hello world" version aimed at initializing the lib and make it available in NPM
+
+### Usage
+
+```js
+import Harvest from 'cozy-harvest'
+```
--- a/packages/cozy-harvest/package.json
+++ b/packages/cozy-harvest/package.json
@@ -1,0 +1,32 @@
+{
+  "name": "cozy-harvest",
+  "version": "0.1.0",
+  "description": "Provides logic, modules and components for Cozy's harvest applications.",
+  "main": "dist/index.js",
+  "author": "Cozy",
+  "license": "MIT",
+  "homepage": "https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest/README.md",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cozy/cozy-libs.git"
+  },
+  "bugs": {
+    "url": "https://github.com/cozy/cozy-libs/issues"
+  },
+  "scripts": {
+    "test": "jest --verbose --coverage",
+    "build": "babel src -d dist",
+    "prepublishOnly": "yarn build",
+    "watch": "yarn build --watch"
+  },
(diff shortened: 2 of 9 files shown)


sizvix:
[(𝐒) [APR] Création d'une librairie](https://trello.com/c/fBQjTy6O/539-%F0%9D%90%92-apr-cr%C3%A9ation-dune-librairie)

packages/cozy-harvest/README.md
```diff
@@ -0,0 +1,10 @@
+Harvest
+=====
+
+Library to render a presentation for datas & configurations konnectors
```
gregorylegarec: I think we can be more explicit, what do you think about:
> Cozy-Harvest will bring the logic, modules and components used in harvest application.
> This first version is just an "Hello World" version aimed to initialize the lib and make it available in NPM.

(My english is far to be perfect so maybe I made some typo.)

packages/cozy-harvest/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "cozy-harvest",
+  "version": "0.1.0",
+  "description": "Harvest library for auto-apps in Cozy",
```
gregorylegarec: Same here, I think we can provide a better description:

> Provides logic, modules and components for Cozy's harvest applications.

What do you think ?

packages/cozy-harvest/README.md
sizvix: Maybe one word about what is "harvest application" ?
Harvest application => generic presentation for collected data ?

gregorylegarec: You're right, maybe :

> Cozy-Harvest will bring the logic, modules and components used in harvest application.
>
> Harvest applications will be apps which will allow to configure konnectors, setup authentification informations and check for synchronized data (And much more in the future).
> Each harvest application will be associated to only one konnector.
>
> This first version is just an "Hello World" version aimed to initialize the lib and make it available in NPM.

packages/cozy-harvest/src/Harvest.spec.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import Harvest from './Harvest'
+import renderer from 'react-test-renderer'
```
gregorylegarec: We are using Enzyme for shallow rendering, see https://github.com/cozy/cozy-store/blob/master/test/components/appsLoading.spec.js#L6 for example

packages/cozy-harvest/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "test": "jest --verbose --coverage",
+    "build": "babel src -d dist",
+    "prepublishOnly": "yarn build",
+    "watch": "yarn build --watch"
+  },
+  "devDependencies": {
+    "@babel/cli": "7.1.2",
+    "babel-preset-cozy-app": "^1.2.1",
+    "react": "16.6.1",
+    "react-final-form": "3.6.7",
```
gregorylegarec: I could be wrong but it seems that `react-final-form` is not needed in this initial version.

packages/cozy-harvest/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "react-test-renderer": "16.6.1",
```
gregorylegarec: We are using `enzyme` to render components in our tests, see further comment

gregorylegarec:
Only a few remarks :)

gregorylegarec:
Thanks @sizvix ! 💯

PR #154, opened 2018-11-29
Get all

ptbrowne:
Fetch a batch of documents by id.


kosssi:
can you add a test?

packages/cozy-doctypes/src/Document.js
```diff
@@ -328,6 +328,31 @@ class Document {
     }
     return result
   }
+
+  /**
+   * Fetch in one request a batch of documents by id.
+   * @param  {String[]} ids - Ids of documents to fetch
+   * @return {Promise} - Promise resolving to an array of documents, unfound document are filtered
```
Crash--: Sure? It seems that we return  an object?

ptbrowne: The error case returned an object yes. Fixed.

PR #177, opened 2018-12-05
test(flags): Add tests

drazik:
I am going to implement node compatible flags, but there were no tests so I didn't feel comfortable with it.


Crash--:
Trop cool! Merci

PR #180, opened 2018-12-06
Add Application

gregorylegarec:
The goal of this PR is to make the Application doctype's schema available
from anywhere. We need it to instanciate CozyClient at least in Home,
Store, and the Bar.

The idea is to just write :

```js
import Application from 'cozy-doctypes/Application'
// ...
  schema: {
    app: Application.schema
  }
}
```

### DIFF
--- a/packages/cozy-doctypes/src/Application.js
+++ b/packages/cozy-doctypes/src/Application.js
@@ -1,0 +1,10 @@
+const Document = require('./Document')
+
+class Application extends Document {}
+
+Application.schema = {
+  doctype: 'io.cozy.apps',
+  attributes: {}
+}
+
+module.exports = Application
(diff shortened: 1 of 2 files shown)


packages/cozy-doctypes/src/Application.js
```diff
@@ -0,0 +1,14 @@
...
+const schema = {
+  doctype: 'io.cozy.apps',
+  attributes: {}
+}
+
+class Application extends Document {
+  static get schema() {
+    return schema
+  }
+}
```
ptbrowne: Instead of using a static getter, why not simply : 

```
Application.schema = {
  doctype: 'io.cozy.apps'
  attributes: {}
}
```

PR #181, opened 2018-12-10
fix(flags): Node and browser detection

drazik:
Using `global` was not enough in all cases to detect if we are in a node process or not.


packages/cozy-flags/src/index.js
```diff
@@ -1,6 +1,8 @@
 export { default as FlagSwitcher } from './browser/FlagSwitcher'
 
-const flag = global
+const isNode = global && global.process && global.process.title === 'node'
```
Crash--: well seen!

PR #182, opened 2018-12-10
Fix`[: missing ']'` issue

CPatchane:
When we use operator in a if statement we have to wrap it into double brackets
See https://stackoverflow.com/a/35281824


gregorylegarec:
Thanks @CPatchane 👍

PR #186, opened 2018-12-11
fix(flags): Check if window exists

drazik:
I've gone way too far to check if we are running in a browser or in node. This is much simpler and less error prone.


Crash--:
But we can, in our node program, create a global `window` var, no? It's also error prone?

drazik:
You're right.

I checked [detect-node](https://github.com/iliakan/detect-node) code. It seems robust. In a browser, even if we create a global `process`, it returns `false`. I think we should use that actually.

drazik:
Yep `detect-node` is small, it adresses only one problem and it has no dependency so I think we can go for it.

PR #187, opened 2018-12-12
Add sanitizer

sizvix:
Add function to sanitize Manifest before send it to the form generation

### DIFF
--- a/packages/cozy-harvest-lib/package.json
+++ b/packages/cozy-harvest-lib/package.json
@@ -19,6 +19,9 @@
     "prepublishOnly": "yarn build",
     "watch": "yarn build --watch"
   },
+  "dependencies": {
+    "lodash": "^4.17.11"
+  },
   "devDependencies": {
     "@babel/cli": "7.2.3",
     "babel-jest": "23.6.0",
--- a/packages/cozy-harvest-lib/src/Harvest.spec.jsx
+++ b/packages/cozy-harvest-lib/src/Harvest.spec.jsx
@@ -6,7 +6,7 @@ import Adapter from 'enzyme-adapter-react-16'
 configure({ adapter: new Adapter() })
 
 describe('Harvest', () => {
-  it('shoul render a div', () => {
+  it('should render a div', () => {
     const component = shallow(<Harvest />).getElement()
     expect(component).toMatchSnapshot()
   })
--- a/packages/cozy-harvest-lib/src/Manifest.js
+++ b/packages/cozy-harvest-lib/src/Manifest.js
@@ -1,0 +1,73 @@
+export const legacyLoginFields = [
+  'login',
+  'identifier',
+  'new_identifier',
+  'email'
+]
+
+export const roleShouldBeRequired = ['identifier', 'password']
+
+export const legacyEncryptedFields = [
+  'secret',
+  'dob',
+  'code',
+  'answer',
+  'access_token',
+  'refresh_token',
+  'appSecret'
+]
+
+export const Manifest = {
+  sanitize: manifest => {
+    let sanitized = JSON.parse(JSON.stringify(manifest))
+    if (!sanitized.fields) return sanitized
+
+    sanitized = Manifest.sanitizeIdentifier(sanitized)
+    sanitized = Manifest.sanitizeRequired(sanitized)
+    sanitized = Manifest.sanitizeEncrypted(sanitized)
+
+    return sanitized
+  },
+
+  sanitizeIdentifier: sanitized => {
+    let hasIdentifier = false
+    for (let fieldName in sanitized.fields)
+      if (sanitized.fields[fieldName].role === 'identifier') {
+        if (hasIdentifier) delete sanitized.fields[fieldName].role
+        else hasIdentifier = true
+      }
+    if (hasIdentifier) return sanitized
+
+    for (let name of legacyLoginFields)
+      if (sanitized.fields[name]) {
+        sanitized.fields[name].role = 'identifier'
+        return sanitized
+      }
+
+    for (let fieldName in sanitized.fields)
+      if (sanitized.fields[fieldName].type !== 'password') {
+        sanitized.fields[fieldName].role = 'identifier'
+        return sanitized
+      }
+
+    return sanitized
+  },
+
+  sanitizeRequired: sanitized => {
+    for (let fieldName in sanitized.fields)
+      if (typeof sanitized.fields[fieldName].required !== 'boolean')
+        sanitized.fields[fieldName].required = true
+
+    return sanitized
+  },
+
+  sanitizeEncrypted: sanitized => {
+    for (let fieldName in sanitized.fields)
+      if (legacyEncryptedFields.includes(fieldName))
+        sanitized.fields[fieldName].encrypted = true
+
+    return sanitized
+  }
+}
+
+export default Manifest
--- a/packages/cozy-harvest-lib/src/Manifest.js
+++ b/packages/cozy-harvest-lib/src/Manifest.js
@@ -1,13 +1,55 @@
-export const legacyLoginFields = [
-  'login',
-  'identifier',
-  'new_identifier',
-  'email'
-]
+import _flow from 'lodash/flow'
+import _cloneDeep from 'lodash/cloneDeep'
 
-export const roleShouldBeRequired = ['identifier', 'password']
+const legacyLoginFields = ['login', 'identifier', 'new_identifier', 'email']
 
-export const legacyEncryptedFields = [
+/**
+ * Ensures that fields has at least one field with the role 'identifier'
+ * @param  {Object} [fields={}] Manifest fields
+ * @return {Object}             Sanitized manifest fields
+ */
+const sanitizeIdentifier = fields => {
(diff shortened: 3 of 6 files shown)


sizvix:
[(𝐒) [APR] Nettoyage du manifeste](https://trello.com/c/kahWjNS0/547-%F0%9D%90%92-apr-nettoyage-du-manifeste)

Crash--:
Tu devrais linter avec nos configs. (Et configurer ton vim pour...) 

J’imagine que le build foire a cause de ça.

packages/cozy-harvest/src/Manifest.spec.js
```diff
@@ -0,0 +1,92 @@
...
+      const input1 = {
+        fields: {
+          password: { type: 'password' },
+          plop: { type: 'text' }
+        }
+      }
+      const result = Manifest.sanitizeIdentifier(input1)
+      expect(result.fields.plop.role).toBe('identifier')
+    })
+
```
gregorylegarec: Another tests to insert here : 

```js
it('should set first non-password field as identifier', () => {
  const manifest = {
    fields: {
      username: {
          type: "text"
       },
      mail: {
         type: 'email'
      },
      password: {
          type: 'password'
      }
  }

  const result = Manifest.sanitize(manifest)
  expect(result.fields.username.role).toBe('identifier')
})

it('should set only one identifier', () => {
  const manifest = {
    fields: {
      username: {
          type: "text"
       },
      mail: {
         type: 'email'
      },
      password: {
         type: 'password'
      }
    }

    const result = Manifest.sanitize(manifest)
    const identifiers = Object.keys(result.fields).filter(fieldName => result.fields[fieldName] === 'identifier')
    expect(identifiers.length).toBe(1)
})

it('should keep identifier priority', () => {
  const manifest = {
    fields: {
      identifier: {
          type: "text"
       },
      username: {
         type: 'email'
      },
      login: {
         type: 'email'
      },
     password: {
         type: 'password'
      }
    }

    const result = Manifest.sanitize(manifest)
    const identifiers = Object.keys(result.fields).filter(fieldName => result.fields[fieldName] === 'identifier')
    expect(identifiers.length).toBe(1)
    expect(result.fields.login.role).toBe('identifier')
})
```

packages/cozy-harvest/src/Manifest.spec.js
```diff
@@ -0,0 +1,92 @@
...
+  describe('sanitizeIdentifier', () => {
+    it('should set role=identifier for login', () => {
+      const input1 = {
+        fields: {
+          login: { type: 'text' },
+          password: { type: 'password' }
+        }
+      }
+      const result = Manifest.sanitizeIdentifier(input1)
```
gregorylegarec: I think we should test only the whole sanitize method, and not the internal functions. By doing so we ensure that the sequence of call will keep the change and that another function will not overwrite some properties.

gregorylegarec:
Before merging, the first commit needs to be cleaned:

`git rebase -i HEAD^^`

edit `firstsha1`
pick `secondsha1`

remove merge syntax and then
`git add package.json`
`git rebase --continue`

fix conflicts and then
`git add package.json`
`git rebase --continue`

packages/cozy-harvest/src/Manifest.js
```diff
@@ -0,0 +1,63 @@
...
+    if (!manifest.fields) return manifest
+    manifest = Manifest.sanitizeIdentifier(manifest)
+    manifest = Manifest.sanitizeIsRequired(manifest)
+    manifest = Manifest.sanitizeEncrypted(manifest)
+    return manifest
+  },
+
+  sanitizeIdentifier: manifest => {
+    for (let fieldName in manifest.fields)
+      if (manifest.fields[fieldName].role === 'identifier') return manifest
```
gregorylegarec: I think it does not clean the case when several fields are marked as `identifier`.

packages/cozy-harvest/src/Manifest.js
```diff
@@ -0,0 +1,63 @@
...
+
+    for (let name of legacyLoginFields)
+      if (manifest.fields[name]) {
+        manifest.fields[name].role = 'identifier'
```
gregorylegarec: Same, it does not clean several fields marked as `identifiers`

This should need a test imho.

packages/cozy-harvest/src/Manifest.js
```diff
@@ -0,0 +1,63 @@
...
+        return manifest
+      }
+
+    for (let fieldName in manifest.fields)
+      if (fieldName !== 'password')
+        manifest.fields[fieldName].role = 'identifier'
```
gregorylegarec: If several fields are not passwords, all will be marked as `identifiers`.

This should need and additional test imho.

packages/cozy-harvest/src/Manifest.js
```diff
@@ -0,0 +1,63 @@
...
+
+    return manifest
+  },
+
+  sanitizeIsRequired: manifest => {
```
gregorylegarec: Nit : s/sanitizeIsRequired/sanitizeRequired/

packages/cozy-harvest/src/Manifest.js
```diff
@@ -0,0 +1,63 @@
...
+    for (let fieldName in manifest.fields)
+      if (typeof manifest.fields[fieldName].required != 'boolean')
```
gregorylegarec: I am surprised that the linter let us used not strict eveluation here. We shoud prefer `!==` over `!=`.

packages/cozy-harvest/src/Manifest.spec.js
```diff
@@ -0,0 +1,136 @@
+/* eslint-env jest */
+import { Manifest } from './Manifest'
+
+describe('Manifest', () => {
+  describe('sanitize', () => {
+    it('should remove "fields" if fields is null', () => {
+      const manifest = {
+        fields: null
```
gregorylegarec: What happen when manifest is `undefined`? I guess we should have the case of an undeclared `fields` property more often than a `fields` property having a `null` value, and the test should be better with `const manifest = {}`. What do you think ?

My bad : the test is right after 👍

packages/cozy-harvest/src/Manifest.spec.js
```diff
@@ -0,0 +1,136 @@
...
+        fields: {
+          password: { type: 'password' },
+          plop: { type: 'text' }
+        }
+      }
+      inputLegacy.fields[name] = { type: 'text' }
+      it('should set role=identifier to ' + name, () => {
+        const result = Manifest.sanitize(inputLegacy)
+        expect(result.fields[name].role).toBe('identifier')
+      })
```
gregorylegarec:
```js
//...
expect(result.fields.plop.role).not.toBe('identifier')
//...
```

packages/cozy-harvest/src/Manifest.spec.js
```diff
@@ -0,0 +1,136 @@
...
+        }
+      }
+      const result = Manifest.sanitize(manifest)
+      const identifiers = Object.keys(result.fields).filter(
+        name => result.fields[name].role === 'identifier'
+      )
+      expect(identifiers.length).toBe(1)
+    })
+
+    it('should keep the identifier pioripy', () => {
```
gregorylegarec: typo s/prioripy/priority/

packages/cozy-harvest/src/Manifest.js
```diff
@@ -0,0 +1,63 @@
...
+  'answer',
+  'access_token',
+  'refresh_token',
+  'appSecret'
+]
+
+export const Manifest = {
+  sanitize: manifest => {
+    if (!manifest.fields) return manifest
+    manifest = Manifest.sanitizeIdentifier(manifest)
```
gregorylegarec: As a good practice we should not mutate the manifest passed as argument and clone it instead : 
```js
const sanitized = { ...manifest }
sanitized.fields = sanitizedFields(sanitized.fields)
// ...

const sanitizedFields = fields => {
  const sanitized = { ...fields }
  sanitized = sanitizeIdentifier(sanitized)
  return sanitized
}

// etc...
```

With this pattern we can prevent any mutation on the original manifest parameter.

gregorylegarec:
Thanks @sizvix ! 🐻

packages/cozy-harvest/src/Manifest.js
```diff
@@ -17,59 +59,39 @@ export const legacyEncryptedFields = [
...
+ * @param  {Object} [fields={}] Manifest fields
+ * @return {Object}             Sanitized Manifest fields
+ */
+const sanitizeEncrypted = fields => {
+  const sanitized = _cloneDeep(fields)
+  for (let fieldName in sanitized)
+    sanitized[fieldName].encrypted =
+      typeof sanitized[fieldName].encrypted === 'undefined'
+        ? legacyEncryptedFields.includes(fieldName)
+        : !!sanitized[fieldName].encrypted
```
sizvix: Could be most readable :
```
if (typeof sanitized.fields[fieldName].required !== 'boolean')
  sanitized[fieldName].encrypted = legacyEncryptedFields.includes(fieldName)
```

PR #190, opened 2018-12-18
Handle 404 from registry

ptbrowne:
When publishing on a non-existent space, the error was not handled.

Also, do sha sum handling in NodeJS.

I find the code in manual a bit hard to understand since it deals with prompt and also with the fetch logic and error handling.

### DIFF
--- a/packages/cozy-app-publish/lib/prepublish.js
+++ b/packages/cozy-app-publish/lib/prepublish.js
@@ -1,6 +1,7 @@
 const runHooks = require('../utils/runhooks')
-const spawn = require('cross-spawn')
-
+const http = require('http')
+const https = require('https')
+const crypto = require('crypto')
 /**
  * Returns only expected value, avoid data injection by hook
  */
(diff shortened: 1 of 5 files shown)


packages/cozy-app-publish/lib/prepublish.js
```diff
@@ -71,30 +72,36 @@ const check = options => {
...
+    })
+
+    req.on('error', e => {
+      reject(e)
+    })
+  })
+
+const shasum = async options => {
+  const { appBuildUrl } = options
+  const shasum = await prepublish.shasum256FromURL(appBuildUrl)
```
gregorylegarec: Nit: I would call directly `shasum256FromURL` here. Is there a reason to use `prepublish` ?

gregorylegarec:
nit: In monorepos I think we should describe the related package in commit message. Example: `feat(publish): Something new`

packages/cozy-app-publish/lib/prepublish.js
ptbrowne: This is to be able to mock it in tests.

PR #191, opened 2018-12-18
Add new cozy-urls package

CPatchane:
Need this to simplify URLs handling in the Cozy-bar to get the best of both world (web and mobile)

I was not inspired for the name, but `cozy-urls` does the job I think.


packages/cozy-urls/src/cozy-urls.js
```diff
@@ -0,0 +1,16 @@
+import { getSSL, getUrlAndHost, setAppName } from './helpers'
+
+const getProps = (cozyURL, sslOption) => {
```
gregorylegarec: Default for `sslOption` should be `true` imho. What do you think ?

packages/cozy-urls/src/helpers.js
```diff
@@ -0,0 +1,80 @@
...
+
+// Optional
+// Allow to specificy which application or library is throwing the error message
+export const setAppName = name => {
+  APP_NAME = name
+}
+export const getAppName = () => APP_NAME || 'The app'
+
+export const getSSL = (cozyURL, ssl) => {
+  if (typeof ssl !== 'undefined' && ssl !== null) return ssl
```
gregorylegarec: We can only test `typeof ssl !== 'undefined'`, `null` being considered as a falsy value here.

packages/cozy-urls/src/helpers.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export const getUrlAndHost = (cozyURL, ssl) => {
+  let url
+  let host
+  const defaultProtocol = ssl ? 'https' : 'http'
+
+  try {
+    // only on mobile we get the full URL with the protocol
+    const parsedURL = new URL(cozyURL)
```
gregorylegarec: Repeated code from line 25

packages/cozy-urls/src/helpers.js
```diff
@@ -0,0 +1,80 @@
...
+    if (!VALID_PROTOCOLS.includes(parsedURL.protocol)) {
+      throw new Error(`Not valid protocol: ${parsedURL.protocol}`)
+    }
+    host = parsedURL.host
+    url = `${parsedURL.protocol}//${host}`
+    return { host, url }
+  } catch (error) {
+    if (isDevelopment()) {
```
gregorylegarec: I think we definetely should use `cozy-logger` to configure our logs browser-side, to avoid explicit tests like this one.

A task for another day however.

packages/cozy-urls/src/helpers.js
CPatchane: Ok I see. Let's do like this so 👍

packages/cozy-urls/src/cozy-urls.js
CPatchane: We can't do that here since we won't be able to pass `null` or `undefined`

kosssi:
If you can't merge it, I check with @CPatchane to make a new PR ;)

CPatchane:
@gregorylegarec, @kosssi did revamp a big part of this PR, feel free to review the main code again :)

packages/cozy-urls/src/cozy-urls.js
```diff
@@ -6,12 +6,17 @@ const ERROR_PROTOCOL_URL_INVALID = `[cozy-urls] Can't find protocol for`
 
 const DOMAIN_REGEX = '(((https?):)?/{2})?(([^.]*.)*[^./:]*)'
 
+let cozyURL
```
drazik: I think there is a potential issue with this on mobile app. If we use `getCozyURL`, then logout, this variable will still contain the old URL. If we login to another instance, the old URL will remain.

packages/cozy-urls/README.md
```diff
@@ -0,0 +1,121 @@
...
+#### `useSSL`
+
+This function return a boolean if the `protocol` is secure.
+
+Here is an example:
+
+```javascript
+import { useSSL } from 'cozy-urls'
+
+const protocolIsSecure = useSSL() // false
```
Crash--: I'm in favor of prefixing boolean variable by `is`. Something like `isUsingSSL` is, in my opinion, less confusing than `useSSL()`. To me `useSSL` will force the app to use SSL protocol

packages/cozy-urls/src/cozy-urls.js
```diff
@@ -0,0 +1,79 @@
...
+
+export const getCozyDomain = url => {
+  if (url === undefined) {
+    url = getCozyURL()
+  }
+
+  try {
+    const parsedURL = new URL(url)
+
+    return parsedURL.host
```
ptbrowne: `host` va renvoyer `a.b.c.d.fr` pour `https://a.b.c.d.fr`, est ce que c'est pas seulement `c.d.fr` que l'on veut ?

packages/cozy-urls/src/cozy-urls.js
```diff
@@ -0,0 +1,79 @@
...
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.warn(ERROR_DOMAIN_URL_INVALID, `'${url}'.`)
+    throw e
```
ptbrowne: Je ne trouve pas ca très utile de renvoyer les erreurs de la lib sous-jacente, et ca n'ira pas dans Sentry. Est ce que vous voyez un inconvénient à renvoyer `throw new Error(ERROR_DOMAIN_URL_INVALID + `: ${url}`)`

packages/cozy-urls/src/cozy-urls.js
kosssi: Si on est sur `https://banks-recette.mycozy.cloud` alors `getCozyURL()` va retourner `https://recette.mycozy.cloud`.
Si tu es sur `http://banks.cozy.tools:8080` alors `getCozyURL()` va retourner `http://cozy.tools:8080`.

C'est donc le résultat attendu ;)

packages/cozy-urls/src/cozy-urls.js
```diff
@@ -6,11 +6,7 @@ const ERROR_DOMAIN_NODE = `[cozy-url] COZY_URL variable isn't defined.`
 const ERROR_PROTOCOL_URL_INVALID = `[cozy-urls] Can't find protocol for`
 const ERROR_DOMAIN_URL_INVALID = `[cozy-urls] Can't find domain for`
 
-let cozyURL
-
 export const getBrowserCozyURL = () => {
-  if (cozyURL) return cozyURL
```
CPatchane: Maybe we can use localstorage here? Without that, each time we will want to get the CozyURL it will do a DOM request. Or maybe we have to explain that detail in the readme to not use it everywhere in the application code.

kosssi: We will talk about optimization when the library will exist

kosssi:
This PR can be closed, I created new PR on cozy-client https://github.com/cozy/cozy-client/pull/331

PR #201, opened 2019-01-02
Harvest: Handle I18n

gregorylegarec:
This PR add I18n handling in cozy-harvest-lib.

It use the cozy I18N wrapper to create a local scope around the AccountForm.

I18n phrases are updated with konnector manifest locales object.

The main <I18n /> wrapper is set directly into `index.js` file, to not interfer with unit tests and keep the component like AccountForm as pure as dumb as possible.

### DIFF
--- a/packages/cozy-harvest-lib/README.md
+++ b/packages/cozy-harvest-lib/README.md
@@ -42,6 +42,33 @@ Fields are declared by their name and can have several properties:
 }
 ```
 
+#### `locales`
+
+Object containing the local as specified in konnector's manifest.
+
+##### Example of `locales` object
+```js
+{
+  en: {
+    fields:
+      username: {
+        label: "Username"
+      }
+    }
+  },
+  fr: {
+    fields: {
+      username: {
+        label: "Nom d'utilisateur"
+      }
+    }
+  }
+}
+```
+Cozy-Harvest-Lib provides a defined range of field labels which do not need to be specified in the manifest. Those field labels are: `answer`, `birthdate`, `code`, `date`, `email`, `firstname`, `lastname`, `login`, `password`, `phone`.
+
+Any konnector can provide new locales for this field labels or add new ones. They will be loaded by the AccountForm component.
+
 ### AccountForm usage
 ```js
 <AccountForm fields={konnector.fields} locales={} />
--- a/packages/cozy-harvest-lib/package.json
+++ b/packages/cozy-harvest-lib/package.json
@@ -15,8 +15,10 @@
   },
   "scripts": {
     "build": "babel src -d dist --copy-files",
+    "prebuild": "yarn tx",
     "prepublishOnly": "yarn build",
     "test": "jest --verbose --coverage",
+    "tx": "tx pull --all || true",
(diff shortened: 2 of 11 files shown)


packages/cozy-harvest-lib/README.md
```diff
@@ -42,6 +42,33 @@ Fields are declared by their name and can have several properties:
...
+  fr: {
+    fields: {
+      username: {
+        label: "Nom d'utilisateur"
+      }
+    }
+  }
+}
+```
+Cozy-Harvest-Lib provides a defined range of field labels which not need to specified in the manifest. Those field labels are: `answer`, `birthdate`, `code`, `date`, `email`, `firstname`, `lastname`, `login`, `password`, `phone`.
```
CPatchane: I think it's `to be specified` here?

packages/cozy-harvest-lib/package.json
```diff
@@ -14,10 +14,12 @@
...
   "scripts": {
-    "test": "jest --verbose --coverage",
-    "build": "babel src -d dist",
+    "build": "babel src -d dist --copy-files",
+    "prebuild": "yarn tx",
     "prepublishOnly": "yarn build",
-    "watch": "yarn build --watch"
+    "test": "jest --verbose --coverage",
+    "tx": "tx pull --all || true",
+    "watch": "babel src -d dist --copy-files --watch"
```
CPatchane: No lint script?

packages/cozy-harvest-lib/src/components/AccountForm.jsx
```diff
@@ -2,29 +2,34 @@ import React, { PureComponent } from 'react'
...
 
 export class AccountField extends PureComponent {
   render() {
-    const { type } = this.props
+    const { name, t, type } = this.props
+    const label = t(`fields.${name}.label`)
     const fieldProps = {
+      ...this.props,
       className: 'u-m-0', // 0 margin
+      id: name,
```
CPatchane: Are we sure the field name will be unique at this point of code?

gregorylegarec: We're not, but it's not a priority at this point. We should even throw a kind of `ManifestError` if it's the case.

packages/cozy-harvest-lib/package.json
gregorylegarec: I will add the lint script in a dedicated PR if you don't mind

packages/cozy-harvest-lib/README.md
gregorylegarec: `which does not need to be specified`

packages/cozy-harvest-lib/src/components/AccountForm.jsx
sizvix: name need to be unique

gregorylegarec: Actually, as `fields` is a JS object and name is extracted from key, name will always be unique at this point.

By the way I removed this line.

PR #206, opened 2019-01-04
refactor(flags): Make flags at build time testable

drazik:
Refactor flags at build time enabling so it can be tested.


packages/cozy-flags/src/index.js
```diff
@@ -7,8 +7,20 @@ const flag = isNode
...
+ * Enables a list of flags
+ * @param {string[]} flagsToEnable
+ */
+export function enableFlags(flagsToEnable) {
+  if (!Array.isArray(flagsToEnable)) {
+    return
+  }
+
+  flagsToEnable.forEach(flagToEnable => flag(flagToEnable, true))
+}
```
ptbrowne: Could `enable` be attached on the exported `flag` function ? This enables namespacing for free `flags.enable` instead of `enableFlags`.

ptbrowne:
Thanks.

PR #210, opened 2019-01-07
Harvest: Add props labels to InputPassword

sizvix:
Inject I18n in Cozy-UI/Field to use InputPassword ( waiting I18n Cozy-UI is ok )


ptbrowne:
If I understand well, the application will have responsibility to have the translation strings ? I don't think this is something that we want. Why not provide the translation strings as part of the component, inside cozy-ui ?

gregorylegarec:
Thanks @sizvix 👍

PR #216, opened 2019-01-09
Use initialValues prop instead of account

gregorylegarec:
We aim to limit the logic related to account into accountForm, and deal only with values.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm.jsx
@@ -118,12 +118,11 @@ export class AccountForm extends PureComponent {
   }
 
   render() {
-    const { account, fields, oauth, t } = this.props
+    const { fields, initialValues, oauth, t } = this.props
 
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/components/AccountForm.jsx
```diff
@@ -105,20 +105,23 @@ export class AccountForm extends PureComponent {
...
     return (
       <Form
         initialValues={initialValues}
         // eslint-disable-next-line no-console
         onSubmit={v => console.log(v)}
         render={({ values }) => (
           <div>
-            <AccountFields manifestFields={sanitizedFields} t={t} />
+            <AccountFields
+              fillEncrypted={!!initialValues}
```
sizvix: `fillEncryted` if there is some values ?
Something like `hasValues` could be better, no ?

gregorylegarec: Functionnaly, `AccountField` can fill empty encrypted value, so it seems more logic to me explicit the feature.

PR #217, opened 2019-01-09
Harvest: Add default I18n placeholder

sizvix:
Add default placeholder based on the field name and with I18n

### DIFF
--- a/packages/cozy-harvest-lib/README.md
+++ b/packages/cozy-harvest-lib/README.md
@@ -69,6 +69,28 @@ Cozy-Harvest-Lib provides a defined range of field labels which do not need to b
 
 Any konnector can provide new locales for this field labels or add new ones. They will be loaded by the AccountForm component.
 
+#### `placeholder`
+
+We could defined a localized placeholder in the manifest
+
+They could be defined like this :
+```js
+{
+  ...
+  locales: {
+    [lang]: {
+      fields:
+        [fieldname]: {
+          placeholder: "the placeholder string"
+        }
+      }
+    },
+  },
+  ...
+}
+```
+
+
 ### AccountForm usage
 ```js
 <AccountForm fields={konnector.fields} locales={} />
@@ -77,3 +99,9 @@ Any konnector can provide new locales for this field labels or add new ones. The
 ## OAuthForm
 
 The OAuthForm is used for OAuth konnector and display a single "Connect" button.
+
+
+## Doc and example about manifests
+
+https://docs.cozy.io/en/tutorials/konnector/#the-manifest
+
--- a/packages/cozy-harvest-lib/src/components/AccountForm.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm.jsx
@@ -27,7 +27,7 @@ const predefinedLabels = [
 
 export class AccountField extends PureComponent {
   render() {
-    const { label, name, t, type } = this.props
+    const { label, placeholder, name, t, type } = this.props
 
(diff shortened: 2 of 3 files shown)


packages/cozy-harvest-lib/README.md
```diff
@@ -69,6 +69,26 @@ Cozy-Harvest-Lib provides a defined range of field labels which do not need to b
 
 Any konnector can provide new locales for this field labels or add new ones. They will be loaded by the AccountForm component.
 
+#### `default placeholder`
```
gregorylegarec: What do we want here is explain how the placeholder are working :

* No placeholder by default
* Placeholders may be specified in the locales object of the konnectors manifest : give an example

packages/cozy-harvest-lib/README.md
```diff
@@ -69,6 +69,28 @@ Cozy-Harvest-Lib provides a defined range of field labels which do not need to b
...
+We could defined a localized placeholder in the manifest
+
+They could be defined like this :
+```js
+{
+  ...
+  locales: {
+    𝘭𝘢𝘯𝘨: {
+      fields:
+        𝘧𝘪𝘦𝘭𝘥𝘯𝘢𝘮𝘦: {
```
CPatchane: What is this font? Is it normal?

packages/cozy-harvest-lib/README.md
```diff
@@ -77,3 +99,9 @@ Any konnector can provide new locales for this field labels or add new ones. The
 ## OAuthForm
 
 The OAuthForm is used for OAuth konnector and display a single "Connect" button.
+
+
+## Doc and example about manifests
+
+[https://docs.cozy.io/en/tutorials/konnector/#the-manifest]
```
CPatchane: Why put that into brackets here?

packages/cozy-harvest-lib/src/components/AccountForm.jsx
```diff
@@ -36,6 +36,9 @@ export class AccountField extends PureComponent {
       className: 'u-m-0', // 0 margin
       fullwidth: true,
       label: t(`fields.${localeKey}.label`, { _: name }),
+      placeholder: placeholder
+        ? placeholder
+        : t(`fields.${localeKey}.placeholder`, { _: '' }),
```
CPatchane: Maybe it could better on one line using `||`:

```js
       placeholder: placeholder || t(`fields.${localeKey}.placeholder`, { _: '' }),
```

gregorylegarec: For placeholders, we do not want to use predefined labels so we should use instead `t(\`fields.${name}.placeholder`, { _: '' })

packages/cozy-harvest-lib/README.md
sizvix: This is an italic font , because you can't set "italic markdown information" in "markdown code"

packages/cozy-harvest-lib/README.md
sizvix: ( confusion with some wiki markdown style ... ^^U )

packages/cozy-harvest-lib/README.md
gregorylegarec: Maybe

```js
{
  locales: {
    [lang]: {
      fields: {
        [fieldName]: {
```
is clearer ? It is JS-compatible as well. What do you think ?

gregorylegarec:
Thanks @sizvix 👍

PR #220, opened 2019-01-09
feat(babel): Update supported browsers list for Cozy platform

### LINKED ISSUE #215: [babel-preset] Old config for babel-preset-env
https://github.com/cozy/cozy-libs/blob/cb6591132bec425078c11ad784b6005208b57305/packages/babel-preset-cozy-app/index.js#L5-L14

We dropped support for IE11 and chrome 42 (https://paper.dropbox.com/doc/Arret-IE11-uc6IozUVabnpGDFKhr85O).

CPatchane:
Fix #215 

~~We wasn't previously transforming the polyfills with [`useBuiltIns`](https://babeljs.io/docs/en/babel-preset-env#usebuiltins) so we was required all the core-js I think, now it will be better handled by requiring only the ones needed for targeted browsers.~~

Edit: Only update the supported browsers list for now

### DIFF
--- a/packages/babel-preset-cozy-app/index.js
+++ b/packages/babel-preset-cozy-app/index.js
@@ -3,13 +3,19 @@
 const { declare } = require('@babel/helper-plugin-utils')
 
 const browserEnv = {
-  targets: {
-    chrome: 42,
-    ie: 10,
-    firefox: 40,
-    browsers: ['last 2 versions']
-  },
-  // don't transform polyfills
+  targets: [
+    'last 2 Chrome major versions',
+    'last 2 ChromeAndroid major versions',
+    'last 2 Firefox major versions',
+    'last 2 FirefoxAndroid major versions',
+    'last 2 Safari major versions',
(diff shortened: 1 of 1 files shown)


packages/babel-preset-cozy-app/index.js
```diff
@@ -3,13 +3,19 @@
...
+    'last 2 ChromeAndroid major versions',
+    'last 2 Firefox major versions',
+    'last 2 FirefoxAndroid major versions',
+    'last 2 Safari major versions',
+    'last 2 iOS major versions',
+    'last 2 Edge major versions',
+    'Firefox ESR',
+    '> 1% in FR',
+    'not dead',
+    'not ie <= 11'
```
drazik: Why not just `not ie`?

edas: I tried, and it wasn't recognized

drazik: OK thanks

PR #222, opened 2019-01-10
fix: Shasum was not following redirects

ptbrowne:
NodeJS standard `http` and `https` do not follow redirects. GitHub archive urls are redirections. We were shasumming the redirection page and not the archive. I changed to use `request` which automatically follow redirects (like a good boy 🐶 ).

For info `curl` does not follow redirects by default, it needs the `-L` option.

### DIFF
--- a/packages/cozy-app-publish/package.json
+++ b/packages/cozy-app-publish/package.json
@@ -6,9 +6,7 @@
   "license": "MIT",
   "scripts": {
     "test": "jest --verbose --coverage",
-    "test:manual": "jest test/manual.spec.js",
-    "test:publish": "jest test/publish.spec.js",
-    "test:travis": "jest test/travis.spec.js"
+    "test:integration": "env TEST_INTEGRATION=1 jest"
   },
   "bin": {
     "cozy-app-publish": "./cozy-app-publish.js"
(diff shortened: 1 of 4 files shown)


packages/cozy-app-publish/package.json
```diff
@@ -6,9 +6,7 @@
   "license": "MIT",
   "scripts": {
     "test": "jest --verbose --coverage",
-    "test:manual": "jest test/manual.spec.js",
-    "test:publish": "jest test/publish.spec.js",
-    "test:travis": "jest test/travis.spec.js"
```
ptbrowne: I did not know why there was that ? It does not seem so useful to me.

packages/cozy-app-publish/test/prepublish.spec.js
```diff
@@ -67,3 +70,13 @@ describe('Prepublish script', () => {
     expect(downcloudSpy).toHaveBeenCalled()
   })
 })
+
+if (process.env.TEST_INTEGRATION) {
```
ptbrowne: Behind a environment variable since it needs to download a file and is potentially long.

packages/cozy-app-publish/package.json
CPatchane: We was using it to test separately by we don't use them anymore, we run all tests on travis here

packages/cozy-app-publish/test/prepublish.spec.js
CPatchane: 👍

PR #225, opened 2019-01-10
Harvest: Handle default values specified in Manifest

gregorylegarec:
Inject default value when there is no initial value

### DIFF
--- a/packages/cozy-harvest-lib/src/Manifest.js
+++ b/packages/cozy-harvest-lib/src/Manifest.js
@@ -1,5 +1,40 @@
 import _flow from 'lodash/flow'
 import _cloneDeep from 'lodash/cloneDeep'
+import _mapValues from 'lodash/mapValues'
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/index.js
```diff
@@ -7,8 +7,10 @@ const dictRequire = lang => require(`./locales/${lang}.json`)
 
 export default Harvest
 
-export const AccountForm = (props, context) => (
-  <I18n dictRequire={dictRequire} lang={context.lang}>
-    <TranslatedAccountForm {...props} />
-  </I18n>
-)
+export const AccountForm = (props, context) => {
+  return (
```
sizvix: Why with `return` ?

gregorylegarec: Side effects of our test earlier this day.

PR #227, opened 2019-01-10
Harvest: Sanitize select props

gregorylegarec:
The goal is to avoid passing an empty string as value, as it make React-Select show no value selected.

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/fields.js
+++ b/packages/cozy-harvest-lib/src/helpers/fields.js
@@ -1,0 +1,32 @@
+/**
+ * Prepare props to pass to React-Select
+ * Options must be consistent and value cannot be ''
+ * @param  {props} props Props passed to Cozy-UI <Field /> component
+ * @return {[type]}       Sanitized props, ready to be passed to
+ */
+export const sanitizeSelectProps = props => {
+  const { options, value } = props
+  const sanitized = { ...props }
+
+  sanitized.options = options
+    ? options.map(option => ({
+        ...option,
+        // legacy
+        label: option.label || option.name
+      }))
+    : []
+
+  // We cannot pass an empty string as value to React-Select, or no value is
+  // never displayed in the select.
+  // `false` is an accepted value.
+  if (typeof value !== 'undefined') {
+    sanitized.value = sanitized.options.find(o => o.value === value)
+  }
+
+  sanitized.type = 'select'
+  return sanitized
+}
+
+export default {
+  sanitizeSelectProps
+}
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/test/helpers/fields.spec.js
```diff
@@ -0,0 +1,39 @@
...
+        options: [
+          { name: 'Option 1', value: '1', label: 'Option 1' },
+          { name: 'Option 2', value: '2', label: 'Option 2' },
+          { name: 'Option 3', value: '3', label: 'Option 3' }
+        ],
+        type: 'select'
+      })
+    })
+
+    it('should not allow empty string for value', () => {
```
sizvix: no test with a valid value ?

packages/cozy-harvest-lib/src/helpers/fields.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+export const sanitizeSelectProps = props => {
+  const { options, value } = props
+  const sanitized = { ...props }
+
+  sanitized.options = options
+    ? options.map(option => ({
+        ...option,
+        // legacy
+        label: option.name
```
CPatchane: For non legacy field we will overwrite the `label` prop here?

packages/cozy-harvest-lib/src/helpers/fields.js
```diff
@@ -0,0 +1,32 @@
...
+      }))
+    : []
+
+  // We cannot pass an empty string as value to React-Select, or no value is
+  // never displayed in the select.
+  // `false` is an accepted value.
+  if (typeof value !== 'undefined') {
+    sanitized.value = sanitized.options.find(o => o.value === value)
```
CPatchane: What if the value is not found here?

sizvix: just no sanitized.value ( which is ok )

packages/cozy-harvest-lib/src/helpers/fields.js
gregorylegarec: Well spotted, `!option.label` && option.name` should be better.

packages/cozy-harvest-lib/src/helpers/fields.js
gregorylegarec: `undefined` is a correct value for React-Select, it just works well 👌

packages/cozy-harvest-lib/src/helpers/fields.js
gregorylegarec: Update : `option.label || option.name` is faaaaar better.

PR #229, opened 2019-01-14
feat(valid): Add validation to enable submit button

sizvix:
The submit button is disabled while required fields aren't filled


gregorylegarec:
Could you add tests please ?

sizvix:
[(𝐗𝐒) [APR] Activer le bouton de validation lorsque les champs sont remplis](https://trello.com/c/ZQKbVtfa/760-%F0%9D%90%97%F0%9D%90%92-apr-activer-le-bouton-de-validation-lorsque-les-champs-sont-remplis)

gregorylegarec:
Also, the commit message should contain the lib name and be `feat(harvest): ...` as we are in a monorepo, it simplifies tree reading

packages/cozy-harvest-lib/test/components/__snapshots__/AccountForm.spec.js.snap
```diff
@@ -158,6 +158,52 @@ exports[`AccountForm AccountFields should render encrypted fields with placehold
...
+            "isThrow": false,
+            "value": "accountForm.submit.label",
+          },
+        ],
+      }
+    }
+  />
+  <DefaultButton
+    className="u-mt-2 u-mb-1-half"
+    disabled={false}
```
gregorylegarec: _AccountForm should have disabled button if there is required field empty_

Shouldn't `disabled` be `true` ?

packages/cozy-harvest-lib/src/components/AccountForm.jsx
```diff
@@ -118,30 +118,41 @@ export class AccountForm extends PureComponent {
...
+    if (props.oauth) return
+    this.sanitizedFields = Manifest.sanitizeFields(props.fields)
+    this.defaultValues = Manifest.defaultFieldsValues(this.sanitizedFields)
+  }
+
+  validate(vals) {
+    let errors = {}
+    for (let name in this.sanitizedFields)
+      if (this.sanitizedFields[name].required && !vals[name])
+        errors[name] = 'Required'
```
gregorylegarec: We never use plain english in our code. We should here:
* Use a constant to identify the error.
or
* At least use a translated locale.

packages/cozy-harvest-lib/src/components/AccountForm.jsx
```diff
@@ -118,30 +118,41 @@ export class AccountForm extends PureComponent {
     if (locales && lang) {
       extend(locales[lang])
     }
+
+    if (props.oauth) return
+    this.sanitizedFields = Manifest.sanitizeFields(props.fields)
```
gregorylegarec: As we are in a Pure Component, we should not store value as inner properties. If the props change, we cannot be sure that we will keep sanitized fields.

Keeping both `sanitizedFields` and `defaultValues` computed in render method will ensure idempotency.

It will avoid the conditional instruction l.122

packages/cozy-harvest-lib/src/components/AccountForm.jsx
```diff
@@ -118,30 +118,41 @@ export class AccountForm extends PureComponent {
     if (locales && lang) {
       extend(locales[lang])
     }
+
+    if (props.oauth) return
```
gregorylegarec: I assume that an error is thrown if we don't make an early return here.

Maybe we should prevent error from `sanitizeFields` instead ?

sizvix: In the previous version there was a redirection to OAuth before sanitize and that's true, fields aren't used and there is no fields for OAuth version ...
So I keep the same idea in first step.

PR #232, opened 2019-01-16
fix(RealTime): Right repo URL

Crash--:


packages/cozy-realtime/package.json
```diff
@@ -5,10 +5,10 @@
   "main": "dist/index.js",
   "author": "Cozy",
   "license": "MIT",
-  "homepage": "https://github.com/cozy/cozy-libs/blob/master/packages/realtime/README.md",
+  "homepage": "https://github.com/cozy/cozy-libs/blob/master/packages/cozy-realtime/README.md",
```
ptbrowne: Shoulnd't we have something like https://docs.cozy.io here ? For example when I compare to React: https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/package.json#L27

PR #234, opened 2019-01-16
Travis: Install Transifex

gregorylegarec:
Needed by Cozy-Harvest-Lib.

First attempt to make it work, first without password (because it's almost lost).


CPatchane:
I think you may need some addons: https://github.com/CPatchane/create-cozy-app/blob/master/packages/cozy-scripts/template/.travis.yml#L17-L24

Also you will need to put a transifexrc template https://github.com/CPatchane/create-cozy-app/blob/master/packages/cozy-scripts/template/app/.transifexrc.tpl with the username `cozy` but don't put the password since it will be versioned

Then you will need a secure TX_PASSWD in the travis config to do something like `- echo "password = $TX_PASSWD" >> ~/.transifexrc
`

PR #235, opened 2019-01-16
Harvest: Disable identifier with initial value

gregorylegarec:
When a field having role set to `identifier` has an initial value, it cannot be editable.


CPatchane:
The build is broken on Travis

PR #236, opened 2019-01-17
Harvest: Improve placeholder for encrypted fields mechanism

gregorylegarec:
Previous pattern was too ambiguous and was working almost by chance.

With this PR, every AccountField is now aware of its own encryption and if it
has an initial value.

Mapping field name with its encrypted name is done one level above, into AccountFields.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm.jsx
@@ -6,11 +6,13 @@ import Button from 'cozy-ui/react/Button'
 import { translate, extend } from 'cozy-ui/react/I18n'
 import Field from 'cozy-ui/react/Field'
 
-import { sanitizeSelectProps } from '../helpers/fields'
+import {
+  getEncryptedFieldName,
+  getFieldPlaceholder,
+  sanitizeSelectProps
+} from '../helpers/fields'
 import Manifest from '../Manifest'
 import OAuthForm from './OAuthForm'
-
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/components/AccountForm.jsx
```diff
@@ -42,7 +44,10 @@ export class AccountField extends PureComponent {
       disabled: !isEditable,
       fullwidth: true,
       label: t(`fields.${localeKey}.label`, { _: name }),
-      placeholder: placeholder || t(`fields.${name}.placeholder`, { _: '' }),
+      placeholder:
+        getPlaceholderForEncryptedFields(this.props) ||
+        placeholder ||
+        t(`fields.${name}.placeholder`, { _: '' }),
```
CPatchane: What about using only one method `getFieldPlaceholder(this.props)`?

gregorylegarec: That was my first thought, let me give it another try.

gregorylegarec: @CPatchane is that better ?

CPatchane: Yep 👍

gregorylegarec: @CPatchane and this ?

CPatchane: Much better 👍

PR #237, opened 2019-01-18
Harvest: Handle dates

gregorylegarec:
Date were ready in Cozy-UI but not yet handled in Cozy-Harvest-Lib.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm.jsx
@@ -50,6 +50,15 @@ export class AccountField extends PureComponent {
       showLabel: t('accountForm.password.show')
     }
     switch (type) {
+      case 'date':
+        return (
+          <Field
+            {...fieldProps}
+            placeholder={t(`fields.${name}.placeholder`, {
+              _: t(`default.dateFormat`).toLowerCase()
+            })}
+          />
+        )
       case 'dropdown':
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/components/AccountForm.jsx
```diff
@@ -50,8 +50,17 @@ export class AccountField extends PureComponent {
...
+        return (
+          <Field
+            {...fieldProps}
+            placeholder={t(`fields.${name}.placeholder`, {
+              _: t(`default.dateFormat`).toLowerCase()
+            })}
+          />
+        )
       case 'dropdown':
-        return <Field {...sanitizeSelectProps(fieldProps)} />
```
CPatchane: Is it related to this PR?

gregorylegarec: It isn't and I should have see it. Thanks !

packages/cozy-harvest-lib/test/components/AccountForm.spec.js
```diff
@@ -149,6 +149,12 @@ describe('AccountForm', () => {
       expect(component).toMatchSnapshot()
     })
 
+    it('render date', () => {
```
ptbrowne:
```suggestion
    it('renders a date', () => {
```

PR #249, opened 2019-01-22
Support suffixed account number

ptbrowne:
La banque postale accounts have letters at the end. We trim the account number only if its length is superior to 11.

Moved encryption command to cozy-doctypes package.json not to pollute the cozy-libs package.json.
Added decryption command.
I find `Makefile` a clear way to express dependencies between files (when it is simple).


packages/cozy-doctypes/src/banking/Makefile
```diff
@@ -0,0 +1,5 @@
+encrypted.tar.gz.gpg: __snapshots__/matching.spec.js.snap fixtures
```
kosssi: I prefer to have this command directly on `package.json` it's more easy to find all command on the same file

kosssi: After explanation with @ptbrowne it's a good idea. I did not know the usefulness of makefile to handle dependencies 👍

PR #251, opened 2019-01-23
[doctypes]: Matching ops, second pass with relaxed date constraint

ptbrowne:
Some BI operations have a date that is different from the Linxo one. To match those operations, we introduce a second pass of matching with only unmatched new transactions against unmatched existing transactions. This second pass can match two operations that have as most 3 days of delta. Within this time range, existing matching methods are used to match the operations.


packages/cozy-doctypes/src/banking/matching-transactions.js
```diff
@@ -58,18 +58,19 @@ const scoreLabel = (newTr, existingTr) => {
...
   const amountPoints = amountDiff === 0 ? methods.push('amount') && 100 : -1000
 
   const points = amountPoints + labelPoints
-  return {
-    op: existingTr,
-    points: points,
-    amountDiff,
-    methods
-  }
+  res.points = points
```
kosssi: you forgot `amountDiff`

ptbrowne: Yes I removed it on purpose. Good catch :)

PR #252, opened 2019-01-24
feat(doctypes): Match accounts from Axa

ptbrowne:
On Axa accounts the number is different from Linxo to BI, Linxo has 5 digits added to the number. I added a new matching method and test data to account for this fact.

Additionally, I added score matching for accounts to take care of accounts where the account number is very similar. For example CreditCard accounts where there is a main account and a differed debit account linked to 1 only "physical" banking account. Now we take into the type of the account as well as if the last digits of the account number if one of the account is a credit card account.

Credit card account tests have been added. All other tests remain green ;)


packages/cozy-doctypes/src/banking/matching-accounts.js
```diff
@@ -48,10 +48,35 @@ const normalizeAccountNumber = (number, iban) => {
   }
 }
 
+const eitherIncludes = (str1, str2) => {
```
kosssi: tu n'avais pas déjà codé cette fonction ?

ptbrowne: Si, juste à côté, pure lazyness, je corrige

PR #253, opened 2019-01-24
feat(doctypes): Register client statically on Document

drazik:
When registering the client on Document, we set it on a global variable. But in banks we have a case in which we register the client somewhere, then want to access it in a class that extends `Document`. Currently we can't to that since we don't have access to the global variable. With this PR, the client is also registered on the `Document` class, so we can access it via `this.cozyClient` in a subclass.

### DIFF
--- a/packages/cozy-doctypes/src/Document.js
+++ b/packages/cozy-doctypes/src/Document.js
@@ -13,8 +13,6 @@ const { parallelMap } = require('./utils')
 const log = require('cozy-logger').namespace('Document')
 const querystring = require('querystring')
 
-let cozyClient
-
 const DATABASE_DOES_NOT_EXIST = 'Database does not exist.'
 
 /**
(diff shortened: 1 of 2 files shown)


ptbrowne:
How does it work when the user disconnects on mobile ?

`Document` API is meant for NodeJS, not long running applications or webpages where the user can disconnect.

drazik:
Actually we are in NodeJS context and we don't want to change the client. We register the client at the "root" of our service: https://github.com/cozy/cozy-banks/blob/master/src/targets/services/onOperationOrBillCreate.js#L206. Then we want to use it in a part of the service: https://github.com/cozy/cozy-banks/blob/master/src/ducks/categorization/services.js#L83 and https://github.com/cozy/cozy-banks/blob/master/src/ducks/categorization/services.js#L285

But it fails because `this` doesn't have a `cozyClient` property, since it is not register on `Document` really.

ptbrowne:
Ah OK, sorry I did not understand that we were in a service.

Yes I can't see why not. Can you add a test showing that a derivated class can add a method accessing the client ? It might be better then to change the static methods of Document to use this.cozyClient instead of the global variable.

drazik:
`sendTransactions` should be static ;)

PR #265, opened 2019-01-31
feat: Remove reconciliationKey

ptbrowne:
`BankAccount:reconciliationKey` was used by the reconciliation
algorithm but is no longer used. It can safely be removed IMHO.


nono:
@ptbrowne there is a conflict to fix before merging this PR (friendly reminder)

PR #266, opened 2019-02-01
fix: Add Android 5 and iOS 10

Crash--:
We need to target Android 5 and iOS 10 for our mobile apps. 

- I have to target Android 4.4 since Android 5 doesn't exist because Webview and OS version are different since Android 5.
- I didn't find any query targeting Android 5 with a very old browser (ie no updated), we can consider targeting Samsung browser since it should follow Android Browser but I'm not sure this is the right way to do.

- I changed to `3 last majors version of iOS` because I think `2` is pretty dangerous specially when a new iOS version is released.

### DIFF
--- a/packages/babel-preset-cozy-app/index.js
+++ b/packages/babel-preset-cozy-app/index.js
@@ -1,17 +1,25 @@
 'use strict'
 
 const { declare } = require('@babel/helper-plugin-utils')
-
+/***
+ * We need to target Android 5 and at least iOS 10 for our mobile apps.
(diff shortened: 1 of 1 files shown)


Crash--:
This time the PR is ready to merge : 

- I have to target Android 4.4 since Android 5 doesn't exist because Webview and OS version are different since Android 5. 
- I didn't find any query targeting Android 5 with a very old browser (ie no updated), we can consider targeting Samsung browser since it should follow Android Browser but I'm not sure this is the right way to do.

nono:
@edas is it good for you?

edas:
LGTM

PR #269, opened 2019-02-05
Better realtime

### LINKED ISSUE #9: [realtime] Ajouter de la documentation pour `unsubscribe`

### LINKED ISSUE #143: [realtime] Multiple websocket are open
Cozy-Realtime open several sockets to the stack instead of only one.

This is due to race condition located in `getCozySocket`. If several calls to `realtime.subscribe` or `subscribeAll` are made in a row, each will generate a new call to `getCozySocket`. But as `getCozySocket` is asynchronous, the socket singleton will not be return and a new condition will be made.

Instead of the cozySocket, we should return here the promise of the cozySocket.

This fix need tests.

### LINKED ISSUE #144: [realtime] [refacto] Use async/await
Cozy guidelines tell us to use `async/await`. It will improve the readability of the realtime lib code.

### LINKED ISSUE #145: [realtime] Use correct endpoint for subscribe
The `subscribe` method should call the expected endpoint: https://docs.cozy.io/en/cozy-stack/realtime/#subscribe

Thanks @nono for spotting this.

### LINKED ISSUE #199: [realtime] Subscribe are not sent on reconnections
When the websocket closes (on a network error for example), the library tries to reconnect. If it is successful, it should also resend the `subscribe` from the previous calls. Currently, it just reopens the websocket and `auth` again. But as it does not resend `subscribe`, it's just a websocket that won't receive any events.

CPatchane:
Some oldest commits are now useless but I suggest to keep them to have the history.

* Handle correctly websocket retries with subscriptions sent again
* ~~Add max poll tries for socket ready status waiting~~ Use a promise to handle socket authenticating
* Add a lot of test, the code is now __full test covered__
* __BREAKING CHANGES__: Remove async usage since it is not necessary for realtime subscribing
* __BREAKING CHANGES__: Refactor API to have just a method `subscribe()` which takes a doc ID or not (subscribe all docs)

Fix #143 
Fix #144 
Fix #145 
Fix #199
Fix #9


packages/cozy-realtime/README.md
```diff
@@ -40,68 +40,60 @@ or
 
 ## API
 
-#### `subscribeAll(config, doctype, parse)`
+#### `subscribe(config, doctype, options)`
 
-This method allow you to subscribe to realtime for all documents of a provided doctype. Here are the parameters:
+This method is exactly the same working as the previous `subscribeAll()` function but to listen only one document. Here are the parameters:
```
ptbrowne: I would not compare this method with the former method since it is unlikely that the person reading the doc is familiar with the old one.

CPatchane: Oops I forgot to remove that, thanks 👍

CPatchane:
[[REAL] Corriger les soucis de perte de temps réel (store)](https://trello.com/c/VLUY6PQB/896-real-corriger-les-soucis-de-perte-de-temps-r%C3%A9el-store)

packages/cozy-realtime/src/index.js
```diff
@@ -13,28 +13,59 @@ const WEBSOCKET_STATE = {
...
+// listener key computing, according to doctype only or with doc id
+const LISTENER_KEY_SEPARATOR = '/' // safe since we can't have a '/' in a doctype
+const getListenerKey = (doctype, docId) =>
+  docId ? [doctype, docId].join(LISTENER_KEY_SEPARATOR) : doctype
+
+const getTypeAndIdFromListenerKey = listenerKey => {
+  const splitResult = listenerKey.split(LISTENER_KEY_SEPARATOR)
+  return {
+    doctype: splitResult[0],
+    docId: splitResult.length > 1 ? splitResult[1] : null
```
nono: I'd have say `splitResult.join(LISTENER_KEY_SEPARATOR)` instead of `splitResult[1]` (in case the docId has a `/`)

packages/cozy-realtime/src/index.js
```diff
@@ -99,228 +120,233 @@ const configTypes = {
...
+  socket.onopen = () => {
+    try {
+      socket.send(
+        JSON.stringify({
+          method: 'AUTH',
+          payload: options.token
+        })
+      )
+    } catch (error) {
+      throw error
```
nono: Why do you need a try/catch that rethrow the same error?

packages/cozy-realtime/src/index.js
```diff
@@ -13,28 +13,59 @@ const WEBSOCKET_STATE = {
...
+  return {
+    doctype: splitResult[0],
+    docId: splitResult.length > 1 ? splitResult[1] : null
+  }
+}
+
 // Send a subscribe message for the given doctype trough the given websocket, but
 // only if it is in a ready state. If not, retry a few milliseconds later.
-function subscribeWhenReady(doctype, socket) {
+const MAX_SOCKET_POLLS = 500 // to avoid infinite poling
```
gregorylegarec: s/poling/polling

packages/cozy-realtime/src/index.js
```diff
@@ -13,28 +13,59 @@ const WEBSOCKET_STATE = {
...
-      // eslint-disable-next-line no-console
       console.warn(`Cannot subscribe to doctype ${doctype}: ${error.message}`)
       throw error
     }
   } else {
-    setTimeout(() => {
-      subscribeWhenReady(doctype, socket)
-    }, 10)
+    // no retries remaining
+    if (!remainedTries) {
```
gregorylegarec: s/remained/remaining/

packages/cozy-realtime/src/index.js
```diff
@@ -99,228 +120,233 @@ const configTypes = {
...
+  try {
+    socket = connectWebSocket(
+      config,
+      onSocketMessage,
+      onSocketClose,
+      NUM_RETRIES,
+      RETRY_BASE_DELAY
+    )
+  } catch (error) {
+    throw error
```
gregorylegarec: Same remark as the one above made by @nono, we could avoid the try catch here if it is just used to throw the same error.

packages/cozy-realtime/src/index.js
```diff
@@ -99,228 +120,233 @@ const configTypes = {
...
+      if (
+        listeners.has(listenerKey) &&
+        listeners.get(listenerKey)[event] &&
+        listeners.get(listenerKey)[event].includes(listener)
+      ) {
+        listeners.set(listenerKey, {
+          ...listeners.get(listenerKey),
+          [event]: listeners.get(listenerKey)[event].filter(l => l !== listener)
+        })
+      }
```
nono: I have some issues to decipher modern JS. Just to be sure, those 10 lines of code does the same as this:

```js
const listenerKey = getListenerKey(doctype, docId)
const listeners = listeners.has(listenerKey) && listeners.get(listenerKey)[event]
if (listener && listeners) {
    listeners = listeners.filter(l => l !== listener)
}
```

Right?

packages/cozy-realtime/src/index.js
```diff
@@ -99,228 +120,233 @@ const configTypes = {
...
+      if (subscriptionsState.has(listenerKey)) {
```
nono: It looks wrong to me: if we had 2 subscriptions for the same doctype (or same doctype+docid), and we unsubcribe from the first, we should keep the `listenerKey` in `subscriptionsState` for the other subscription. I think we should check that `listeners.get(listenerKey)[event]` is empty before removing the key from `subscriptionsState`.

nono: Nope, it's more complicated than that: we should check that `listeners.get(listenerKey)` has empty arrays for all the possible events.

packages/cozy-realtime/src/index.js
```diff
@@ -99,228 +120,233 @@ const configTypes = {
...
-    const docListenerCurried = listener => {
-      return syncedDoc => {
-        if (syncedDoc._id === doc._id) {
-          listener(syncedDoc)
-        }
+      const listenerKey = getListenerKey(doctype, docId)
+
+      if (!listeners.has(listenerKey)) {
+        listeners.set(listenerKey, {})
+        subscribeWhenReady(doctype, socket, docId)
```
nono: If I understand correctly, if we setup a listener for `doctype=io.cozy.jobs & docId= 123` for both `updated` and `deleted` events, we will call `subscribeWhenReady` twice, and the stack will receive 2 times the message to subscribe for this document. Right?

packages/cozy-realtime/src/index.js
```diff
@@ -13,28 +13,59 @@ const WEBSOCKET_STATE = {
...
+}
+
 // Send a subscribe message for the given doctype trough the given websocket, but
 // only if it is in a ready state. If not, retry a few milliseconds later.
-function subscribeWhenReady(doctype, socket) {
+const MAX_SOCKET_POLLS = 500 // to avoid infinite poling
+export function subscribeWhenReady(
+  doctype,
+  socket,
+  docId,
```
nono: [nitpick] I'd have inversed the order of the two first arguments: `function subscribeWhenReady(socket, doctype, docId, remainedTries)` to keep `doctype` and `docId` together

packages/cozy-realtime/src/index.js
```diff
@@ -13,28 +13,59 @@ const WEBSOCKET_STATE = {
...
-    }, 10)
+    // no retries remaining
+    if (!remainedTries) {
+      const error = new Error('socket failed to connect')
+      console.warn(`Cannot subscribe to doctype ${doctype}: ${error.message}`)
+      throw error
+    } else {
+      setTimeout(() => {
+        subscribeWhenReady(doctype, socket, docId, --remainedTries)
+      }, 10)
```
nono: I don't like this polling. Can't we just to `socket.onOpen(() =>subscribeWhenReady(doctype, socket, docId))`?

packages/cozy-realtime/src/index.js
```diff
@@ -99,228 +120,233 @@ const configTypes = {
...
+    window.removeEventListener('beforeunload', windowUnloadHandler)
+    if (typeof onclose === 'function') onclose(event, numRetries, retryDelay)
+  }
+  socket.onerror = error => console.error(`WebSocket error: ${error.message}`)
 
-      resolve(socket)
+  if (isRetry && subscriptionsState.size) {
+    for (let listenerKey of subscriptionsState) {
+      const { doctype, docId } = getTypeAndIdFromListenerKey(listenerKey)
+      subscribeWhenReady(doctype, socket, docId)
```
nono: It's probably not very important, but if we are trying to subscribe to a doctype while retrying to connect, we will call `subscribeWhenReady` from both here and line 273.

packages/cozy-realtime/src/index.js
nono: Or maybe we can use a promise resolved after sending the authentication, to avoid sending the subscribe message before the auth message.

packages/cozy-realtime/src/index.js
```diff
@@ -99,228 +120,233 @@ const configTypes = {
...
-    }
+  const options = {
+    secure: config.url ? isSecureURL(config.url) : true,
+    ...config
+  }
 
-    const protocol = options.secure ? 'wss:' : 'ws:'
-    const domain = options.domain || getDomainFromUrl(options.url)
+  const protocol = options.secure ? 'wss:' : 'ws:'
+  const domain = options.domain || new URL(options.url).host
```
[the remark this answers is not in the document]
CPatchane: Yes, it's already checked before and I wasn't able to have a wrong formatted URL here in the tests

packages/cozy-realtime/src/index.js
```diff
@@ -99,228 +120,233 @@ const configTypes = {
...
-            } catch (error) {
-              // eslint-disable-next-line no-console
-              console.error(
-                `Unable to reconnect to realtime. Error: ${error.message}`
-              )
-            }
-          }, retryDelay)
-        }
-      }
+    if (listenerKey === payload.type) return
```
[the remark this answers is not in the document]
CPatchane: It can be true for specific documents like `io.cozy.apps` and `io.cozy.konnectors` like we discussed :/

packages/cozy-realtime/src/index.js
nono: But, if the value is an array, we don't need to set a new value, we can update the array in place:

```js
const m = new Map()
m.set('foo', [])
const foo = m.get('foo')
foo.push(1)
m.get('foo') // => [1]
```

nono: Oh, right, `filter` is not in place, my first snippet doesn't work.

My goal was not to suggest that you should change this code, just to understand what it does. And I think I'm OK now, so I will mark this thread as resolved.

packages/cozy-realtime/src/index.js
CPatchane: I added a promise to handle that properly, good point 👍

packages/cozy-realtime/src/index.js
CPatchane: Actually if there is already a listener registered for a doctype it won't. If for the previous subscription, we called unsubscribe, the listener will be removed and the new subscribe will send a new message as expected.

kosssi:
@CPatchane je ne comprends pas pourquoi tu m'as supprimé de [CODEOWNERS](https://github.com/cozy/cozy-libs/pull/269/files#diff-5daf978e60e03b3975b485d5f1b449dcL3) ?

CPatchane:
@kosssi Oh, sorry for that, I quickly supposed that was due to an out of date list since I don't see you in our maintainers/points of contact list, working on or using this library and I didn't got a review for this refactoring PR. So to not have too much people watching this lib, I thought I should remove it. But I was maybe wrong I can add you back if you wish, sorry again.

PR #270, opened 2019-02-06
Fix banque postale

ptbrowne:
Sometimes the number taken from the IBAN is not the
same as the old Linxo number


packages/cozy-doctypes/src/banking/matching-accounts.js
```diff
@@ -138,21 +138,32 @@ const normalizeAccount = account => {
...
+}
+
 const findMatch = (account, existingAccounts) => {
-  // IBAN
-  if (account.iban) {
-    const matchIBAN = findExactMatch('iban', account, existingAccounts)
-    if (matchIBAN && matchIBAN.match) {
-      return matchIBAN
+  // Start with exact attribute matches
+  for (let exactAttribute of exactMatchAttributes) {
```
kosssi: If you write `let` in `for` I search where you update the variable ;)

packages/cozy-doctypes/src/banking/matching-transactions.js
```diff
@@ -192,7 +192,9 @@ const matchTransactions = function*(newTrs, existingTrs) {
         maxDateDelta: delta * DAY
       }
     )) {
-      result.method = result.method + `-delta${delta}`
+      if (result.method) {
+        result.method = result.method + `-delta${delta}`
```
kosssi: You can write:

```suggestion
        result.method += `-delta${delta}`
```

PR #273, opened 2019-02-07
Harvest: Do not show OAuth button in edit form

gregorylegarec:
We should not show the Connect button for an existing OAuth account.


CPatchane:
Thank @gregorylegarec 🦄

PR #274, opened 2019-02-07
Harvest: Handle legacy locales for field label

gregorylegarec:
Some previous locales for field labels are now out of cozy-harvest-lib scope.

This PR got them back, identified as legacy, until the manifest of
concerned konnectors evolves.


CPatchane:
The build is broken :/

PR #281, opened 2019-02-12
feat: Use transpiled file only in browser context

ptbrowne:
1. Babel runtime was missing from the dependencies after the last update of babel-preset-cozy-app : since the list of supported of browsers has changed and now includes browsers that do not support async/await, babel-preset-env now uses the babel-transform-runtime to transpile the async/await. 

> Make sure you include @babel/runtime as a dependency.

https://babeljs.io/docs/en/babel-plugin-transform-runtime#regenerator-aliasing

2. The doctypes library is used in Node contexts, like services or konnectors. It is not necessary for node to use transpiled code. The solution is to use the `browser` field of the `package.json`. Check the [docs](https://docs.npmjs.com/files/package.json#browser).

- The `browser` field is used by webpack as the main entrypoint of the package.
- The `main` field continues to be used by node as the main entrypoint of the package.

Node users can use the non transpiled code and apps that run in a web-browser use the transpiled code 🎉


gregorylegarec:
Great 👍

edas:
👍 for me if we don't need any transpilation at all (are we sure of that?). We also could launch two different build/transpilation if really needed. I don't see a real problem with that last option.

Crash--:
main / browser, TIL!

Thanks 

👍

ptbrowne:
> 👍 for me if we don't need any transpilation at all (are we sure of that?).

Yes, the node code was running before without transpilation without  any issue.

PR #285, opened 2019-02-13
Handle legacy locale label

gregorylegarec:
`branchName` is still used by Bank konnectors


CPatchane:
Thanks @gregorylegarec ⛸

PR #287, opened 2019-02-15
feat: Add optional chaining to the babel preset

edas:
[Optional chaining](https://github.com/tc39/proposal-optional-chaining)
allows to chain potentially null components.

Instead of doing
```
if (window && window.location && window.location.protocol) {
  ssl = window.location.protocol
}
```
we could do
```
ssl = window?.location?.protocol
```

The spec is stage 1 only but [they seem to have reach consensus](https://github.com/tc39/proposal-optional-chaining/pull/75#issue-234782431)
and [the formal stage 2 decision should be in March](https://github.com/tc39/proposal-optional-chaining/issues/76#issuecomment-443435613)

There should not be any compatibility break


ptbrowne:
Why ?

PR #295, opened 2019-02-27
Harvest: Fix job promise fullfilment

gregorylegarec:
Promise was resolved on first job update, even if job
state was not done or errored. Handlers were triggered too early.

### DIFF
--- a/packages/cozy-harvest-lib/src/connections/triggers.js
+++ b/packages/cozy-harvest-lib/src/connections/triggers.js
@@ -3,6 +3,8 @@ import { subscribe } from 'cozy-realtime'
 const JOBS_DOCTYPE = 'io.cozy.jobs'
 const TRIGGERS_DOCTYPE = 'io.cozy.triggers'
 
+const JOB_END_STATES = ['done', 'errored']
+
 /**
  * Creates a trigger with given attributes
  * @param  {Object} client CozyClient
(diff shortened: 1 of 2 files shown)


Crash--:
It seems that tests are red

gregorylegarec:
My bad, should be ok now.

packages/cozy-harvest-lib/src/connections/triggers.js
```diff
@@ -57,7 +57,9 @@ const waitForLoginSuccess = async (
   return new Promise(resolve => {
     const resolveJob = () => resolve(job)
     setTimeout(resolveJob, expectedSuccessDelay)
-    jobSubscription.onUpdate(resolveJob)
+    jobSubscription.onUpdate(
+      job => ['done', 'errored'].includes(job.state) && resolveJob(job)
```
ptbrowne: const JOB_END_STATES = ['done', 'errored']

packages/cozy-harvest-lib/test/connections/triggers.spec.js
```diff
@@ -88,10 +88,14 @@ describe('Trigger mutations', () => {
...
+      state: 'done'
+    }
+
     beforeAll(() => {
       // Mock realtime to respond at 100ms
       realtime.subscribe.mockImplementation(() => ({
-        onUpdate: fn => setTimeout(() => fn(), jobSuccessResponseTime)
+        onUpdate: fn => setTimeout(() => fn(updatedJob), jobSuccessResponseTime)
       }))
     })
```
ptbrowne: Can you add a test showing that if the state is not done or errored, the promise is not resolved ?

PR #297, opened 2019-02-28
Harvest: Force folder_to_save for trigger

gregorylegarec:
It appeared that trigger still need a `message.folder_to_save` property.

This PR fixes this, but it comes with the full package:

* Folder fetching and/or creation
* Permission on folder for konnector
* Reference to files in konnector

The goal is to be iso with Cozy-Home, and it actually fixes
the issue we encountered.

I tried to make atomic commits, I did not start developing
with the idea to make so much commits. But good news they
are all pretty tiny and contains a lot of tests.

### DIFF
--- a/packages/cozy-harvest-lib/src/Manifest.js
+++ b/packages/cozy-harvest-lib/src/Manifest.js
@@ -3,6 +3,8 @@ import _cloneDeep from 'lodash/cloneDeep'
 import _mapValues from 'lodash/mapValues'
 import _pickBy from 'lodash/pickBy'
 
+const IDENTIFIER = 'identifier'
+
 /**
  * Returns a key/value object with field as key and default, if it exists in
  * fields parameter.
@@ -58,7 +60,7 @@ const sanitizeIdentifier = fields => {
   const sanitized = _cloneDeep(fields)
   let hasIdentifier = false
   for (let fieldName in sanitized)
-    if (sanitized[fieldName].role === 'identifier') {
+    if (sanitized[fieldName].role === IDENTIFIER) {
       if (hasIdentifier) delete sanitized[fieldName].role
       else hasIdentifier = true
     }
@@ -66,13 +68,13 @@ const sanitizeIdentifier = fields => {
 
   for (let name of legacyLoginFields)
     if (sanitized[name]) {
-      sanitized[name].role = 'identifier'
+      sanitized[name].role = IDENTIFIER
       return sanitized
     }
 
   for (let fieldName in sanitized)
     if (sanitized[fieldName].type !== 'password') {
-      sanitized[fieldName].role = 'identifier'
+      sanitized[fieldName].role = IDENTIFIER
       return sanitized
     }
 
--- a/packages/cozy-harvest-lib/src/Manifest.js
+++ b/packages/cozy-harvest-lib/src/Manifest.js
@@ -1,5 +1,6 @@
 import _flow from 'lodash/flow'
 import _cloneDeep from 'lodash/cloneDeep'
+import findKey from 'lodash/findKey'
 import _mapValues from 'lodash/mapValues'
 import _pickBy from 'lodash/pickBy'
 
@@ -37,6 +38,14 @@ const defaultFieldsValues = fields => {
     value => value.default
   )
 }
+
+/**
+ * Returns the key for the field having the role=identifier attribute
+ * @param  {Object} fields Konnector fields
+ * @return {[type]}        The key for the identifier field, example 'login'
+ */
+const getIdentifier = (fields = {}) =>
+  findKey(sanitizeIdentifier(fields), field => field.role === IDENTIFIER)
 
 /**
  * Ensures old fields are removed
@@ -146,6 +155,7 @@ export const sanitize = (manifest = {}) =>
 
 export default {
   defaultFieldsValues,
+  getIdentifier,
   sanitize,
   sanitizeFields
 }
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -7,10 +7,7 @@ import AccountCreator from './AccountCreator'
 import AccountEditor from './AccountEditor'
 import TriggerSuccessMessage from './TriggerSuccessMessage'
 import { triggersMutations } from '../connections/triggers'
-import {
-  buildKonnectorCron,
-  buildKonnectorTriggerAttributes
-} from '../helpers/triggers'
+import triggers from '../helpers/triggers'
 
 const IDLE = 'IDLE'
 const RUNNING = 'RUNNING'
@@ -61,10 +58,10 @@ export class TriggerManager extends Component {
     })
 
     const trigger = await createTrigger(
-      buildKonnectorTriggerAttributes({
+      triggers.buildAttributes({
         konnector,
         account,
-        cron: buildKonnectorCron(konnector)
+        cron: triggers.buildCron(konnector)
       })
     )
 
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -11,6 +11,7 @@ import { triggersMutations } from '../connections/triggers'
 import filesMutations from '../connections/files'
 import permissionsMutations from '../connections/permissions'
 import accounts from '../helpers/accounts'
+import cron from '../helpers/cron'
 import konnectors from '../helpers/konnectors'
 import { slugify } from '../helpers/slug'
 import triggers from '../helpers/triggers'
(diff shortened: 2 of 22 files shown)


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -61,10 +58,10 @@ export class TriggerManager extends Component {
     })
 
     const trigger = await createTrigger(
-      buildKonnectorTriggerAttributes({
+      triggers.buildAttributes({
         konnector,
         account,
-        cron: buildKonnectorCron(konnector)
+        cron: triggers.buildCron(konnector)
```
ptbrowne: Nit: Since we are on the subject of better names, what do you think of `cron.fromKonnector` ?

packages/cozy-harvest-lib/src/connections/files.js
```diff
@@ -0,0 +1,39 @@
...
+  return data
+}
+
+/**
+ * Retrieves a directory from its path
+ * @param  {Object}  client CozyClient
+ * @param  {string}  path   Directory path
+ * @return {Object}        Created io.cozy.files document
+ */
+export const fetchDirectoryByPath = async (client, path) => {
```
ptbrowne: Why not statDirectoryByPath ? Since you are calling statByPath below ?

packages/cozy-harvest-lib/src/connections/files.js
```diff
@@ -0,0 +1,39 @@
...
+  let response
+  try {
+    response = await client.collection(FILES_DOCTYPE).statByPath(path)
+  } catch (error) {
+    if (error && error.status === 404) return null
+    throw new Error(error.message)
+  }
+
+  return response.data
+}
```
ptbrowne: It seems that the two above methods have the same skeleton. I wonder if something like below could serve ?

```js
/* Decorates a function so that the first argument (client) becomes the collection */
const collectionMethod = doctype => collectionFn => (client, ...restArgs) => {
  const col = client.collection(doctype)
  return collectionFn(client, ...restArgs)
}

const fileMethod = collectionMethod(FILES_DOCTYPE)

export const createDirectoryByPath = fileMethod(async (files, path) => {
 const { data } = await files.createDirectoryByPath(path)
 return data
})

 export const softStatByPath = fileMethod(async (files, path) => {
  try {
    const { data } = files.statByPath(path)
    return data
  } catch (error) {
    if (error && error.status === 404) return null
    throw new Error(error.message)
  }
}
```

ptbrowne: I think it's worth refactoring those things since otherwise we are basically rewriting stackClient.

packages/cozy-harvest-lib/src/Manifest.js
```diff
@@ -38,6 +39,14 @@ const defaultFieldsValues = fields => {
   )
 }
 
+/**
+ * Returns the key for the field having the role=identifier attribute
+ * @param  {Object} fields Konnector fields
+ * @return {[type]}        The key for the identifier field, example 'login'
+ */
+const getIdentifier = (fields = {}) =>
+  findKey(sanitizeIdentifier(fields), field => field.role === IDENTIFIER)
```
ptbrowne: Is it the role of the getter to sanitize ? I think inputs should be sanitized/normalized as early as possible so that this method does not have to sanitize by itself.

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -11,7 +13,8 @@ export const prepareAccountData = (konnector, data) => {
   // For now we are just ensuring legacy
   return {
     auth: data,
-    account_type: konnector.slug
+    account_type: konnector.slug,
+    identifier: Manifest.getIdentifier(konnector.fields)
```
ptbrowne: sanitized here for example, or higher up

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -28,3 +40,9 @@ export const mergeAuth = (account, authData) => ({
   ...account,
   auth: merge(account.auth, authData)
 })
+
+export default {
+  getLabel,
+  prepareAccountData,
```
ptbrowne: s/prepareAccountData/prepareData/ . It is still unclear what prepareData does. What does it do ?

packages/cozy-harvest-lib/test/helpers/accounts.spec.js
```diff
@@ -1,4 +1,4 @@
-import { mergeAuth, prepareAccountData } from 'helpers/accounts'
+import { getLabel, mergeAuth, prepareAccountData } from 'helpers/accounts'
```
ptbrowne: It is unfortunate that it is not possible to do 

`import { getLabel, mergeAuth, prepareData } as accounts from 'helpers/accounts'`

It means we have two things that are not compatible

* Keeping the namespace (good for clarity, shorter names in implementation file)
* Importing only what we need (presumably good for tree shaking)

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -51,17 +56,35 @@ export class TriggerManager extends Component {
...
+    let folder
+    if (konnectors.needsFolder(konnector)) {
+      const path = `${t('default.baseDir')}/${konnector.name}/${slugify(
+        accounts.getLabel(account)
+      )}`
+
+      folder =
+        (await fetchDirectoryByPath(path)) ||
+        (await createDirectoryByPath(path))
+    }
```
ptbrowne: This can be extracted into a `ensureDirectoryForKonnector` method ? I think the need for `let` is a smell.

packages/cozy-harvest-lib/src/connections/files.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Add references to given files into given documents
+ * @param  {Object}  client    CozyClient
+ * @param  {Object}  document  Target document
+ * @param  {[type]}  files     Files to reference
+ * @return {Object}           Updated document
+ */
+export const addReferencesTo = async (client, document, files) => {
+  const { data } = client
```
Crash--: await ?

packages/cozy-harvest-lib/src/Manifest.js
gregorylegarec: I agree with sanitizing early, but here I am not sure. Actually getIdentifier could not return any identifier if fields parameter is not sanitized.

For now, fields are sanitized in AccountForm at rendering.

Maybe it indicates a bad conception here, our konnector should be sanitized in the TriggerManager. It would imply that our Manifest helper should better be a konnector helper/model.

But event there, I could not be sure that calling:

```js
KonnectorModel.fields.getIdentifier(konnector.fields)
```

will return me identifier if `konnectors.fields` are not sanitized.

packages/cozy-harvest-lib/src/helpers/accounts.js
gregorylegarec: I'll think about it. 👍

packages/cozy-harvest-lib/test/helpers/accounts.spec.js
gregorylegarec: I will namespace this, that the best solution I think.

packages/cozy-harvest-lib/src/connections/files.js
gregorylegarec: Oh, well spotted.

gregorylegarec:
Thanks guy for your review ❤️  ! I will add some changes and merge this.

packages/cozy-harvest-lib/src/connections/files.js
gregorylegarec: I think the `statBy` is pretty unexplicit, so I tend it to encapsulate it into functions with a better name. But you're right, let's be logic and keep this one.

ptbrowne: If your function wraps the statByPath to add a try/catch for 404, I think it is confusing to have a different name, a reader of the function has to forget cannot rely on its the knowledge of the underlying library.

To me, either we find the name of library not explicit and we send a PR there to deprecate the confusing name and use the new name, or we accept the name of the library and keep the coherence :)

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
gregorylegarec: I like the idea, but it's a little work, I will refactor it the next time I will implement some logic for files.

packages/cozy-harvest-lib/src/connections/files.js
gregorylegarec: The pattern is not repeated anymore, but it's a nice helper to keep for Cozy-Client and future Models I think. 👍

PR #302, opened 2019-03-04
feat: Add updatedByApps

ptbrowne:
Automatically add the slug of the application updating the
document when creating/updating the document.

Contrary to the docs [here](https://github.com/cozy/cozy-doctypes/#document-metadata),
`updatedByApps` is not a list but an object with slug -> update date. We think it
would be more useful and it's clearer than the convention "last app updating is first".

If this PR is accepted, I will update the docs there.

```
{
  cozyMetadata: {
   updatedByApps: {
     boursorama: new Date()
   }
  }
}
```

EDIT: Fixed example

### DIFF
--- a/packages/cozy-doctypes/src/Document.js
+++ b/packages/cozy-doctypes/src/Document.js
@@ -44,6 +44,20 @@ function sanitizeKey(key) {
   return key
 }
 
+function updateCreatedByApp(cozyMetadata, appSlug) {
+  if (!cozyMetadata.updatedByApps) {
+    cozyMetadata.updatedByApps = []
+  }
+  const now = new Date()
+  for (const appInfo of cozyMetadata.updatedByApps) {
+    if (appInfo.slug === appSlug) {
+      appInfo.date = now
+      return
+    }
+  }
+  cozyMetadata.updatedByApps.push({ slug: appSlug, date: now })
+}
+
 const withoutUndefined = x => omitBy(x, isUndefined)
 
 const flagForDeletion = x => Object.assign({}, x, { _deleted: true })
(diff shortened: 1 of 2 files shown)


cedricmessiant:
@ptbrowne don't you mean : 

```
{
  cozyMetadata: {
    updatedByApps: {
      boursorama: new Date()
    }
  }
}```

?

In this case, we will have to change `ensureCozyMetadata` in `cozy-client`: https://github.com/cozy/cozy-client/commit/f12574bb346aec23f47f283a89dbcf0c01e7d403

In this case, do we remove `updatedAt` metadata and let the consumer search for the app with the latest update date?

Do we also change `createdByApp` and `createdByAppVersion` to do something similar ?

```
createdByApp: {
  contacts: {
    date: '2018-01-22...'
    version: 2.0.0
  }
```

**Comment by edas** [comment]
LGTM but we first need to have a review in the different teams to look who is using the documented way today, and if an upgrade is doable there

**Comment by ptbrowne** [comment]
> Do we remove updatedAt metadata and let the consumer search for the app with the latest update date?

Mmh, I think keeping updatedAt as a simple attribute will let us search more efficiently with a request like `{ 'cozyMetadata.updatedAt' : { $gt: '2018-09-01' }}`

To go further, if we could support indexing with updatedApps, with a slight solution the proposed solution we could have : 

```
cozyMetadata: {
  updatedByApps: [
    { slug: 'boursorama', date: new Date() }
  ]
}
```

It would work with $elemMatch then.

 `{ 'cozyMetadata.updatedByApp' : { $elemMatch: $gt: '2018-09-01' }}}`

@nono Do you have any advice for indexing with Mango, would the solution with an array work well with Mango ?

cedricmessiant:
On Appvengers side, we will have to change stuff in Contacts, konnector-google and ACH (migration). For Contacts and konnector, it should be trivial once the change is done on cozy-client side: https://github.com/cozy/cozy-client/commit/f12574bb346aec23f47f283a89dbcf0c01e7d403

edas:
Vu sur zoom 

- contact est la seule app qui a du code correspondant, et il n'est pas encore en prod
- @cedricmessiant va mettre à jour la méthode de cozyClient
- la méthode de cozyClient ne va plus se baser sur le schéma mais va ajouter/mettre à jour les champs de façon fixe
- on utilise la dernière forme proposée par Patrick

PR #275, opened 2019-02-07
Harvest: Run trigger and konnector job

gregorylegarec:
The purpose of this PR is to handle Account creation and edition, trigger creation an launch.

**Error handling will be done our the next card.**

It is a rewriting of the existing mess in Cozy-Home.

It is based on the previously developed `<AccountForm />` component, which deals with displaying konnector fields properly (Btw I think we will rename it KonnectorForm or KonnectorFieldsForm in the near future).

The idea is to retrieve the data from the `<AccountForm />` and:
* Create or update an account
* Create a trigger in case of an account creation
* Launch the trigger and indicates to the parent component (located in Cozy-Home) that the login has succeed.

To do so, we use:

### AccountForm
Evolved to be now disabled on submitting.
Also enable button when editing if a required encrypted field is empty but got initial value (encrypted values are not displayed)
### AccountCreator
Encapsulates an `<AccountForm />`, creates an account and calls the expected handlers when it's done.
### AccountEditor
Encapsulates an `<AccountForm />`, update an account and calls the expected handlers when it's done.
### TriggerManager
Main component of this PR, it takes `konnector` as prop and renders either an `<AccountEditor />` if an account is passed as prop as well, or an `<AccountCreator />` if no account is passed. It also display the `<TriggerSuccessMessage />` component in case of account creation succeess.
### TriggerSuccessMessage
A really dumb component which just display an image with a `Done` button. I still wonder if it should be fully handled by Cozy-Home or if it has its place into `cozy-harvest-lib`. Time will tell.

### accounts
Most of them have been copied/pasted/adapted from Cozy-Home
### triggers
New mutations based on recette TriggerCollection in Cozy-Client

### accounts
Mainly preparing data and hydrating documents

### triggers
Preparing trigger and cron values. Cron stuff have been copied/pasted/refactored from Cozy-Home. The refacto here has been big.

### daytime
Helpers which allow us to generate randome day time, used for cron values.

PR #305, opened 2019-03-05
Harvest: Handle Errors during account, files and trigger mutations

gregorylegarec:
The goal of this PR is to catch and handle errors from konnector job, and regular mutations errors as well.

It starts with a few refactoring:
* Code/test simplifications
* AccountCreator and AccountEditor logics are now too close as AccountCreator needs to keep the created account in case of error, so are both merged into TriggerManager which now handle account creation and update.
* TriggerSuccessMessage disappears, as it is actually more the Cozy-Home's role to display the success message and picture.

Then we create:
* KonnectorJobError class, as we need a dedicated logic to handle errors from konnector jobs
* AccountFormError component, which display Errors and KonnectorJobErrors

At last, we catch errors in TriggerManager and handle them in AccountForm.

NB: During my last tests I realized that a case is not well covered. When a directory creation, permission creation, or reference addition fails, the account is now orphan and any update cannot launch the associated trigger. I will amend this PR with a fix that check directory and trigger existence and will create them if they does not exist. But this PR can yet be reviewed.


packages/cozy-harvest-lib/test/components/AccountForm.spec.js
```diff
@@ -101,39 +101,71 @@ describe('AccountForm', () => {
...
-  it('should have disabled button if there is required field empty', () => {
-    const fields = {
-      test: {
-        type: 'text'
+  describe('Submit Button', () => {
+    const getButtonDisabledValue = wrapper =>
+      wrapper
+        .dive()
+        .find('DefaultButton')
+        .props().disabled
```
ptbrowne: Nit: I think getButtonDisabledValue() does a bit too much stuff (finding the button, and checking a prop), I find it a bit cumbersome to pass wrapper also.

I like to keep a global wrapper variable that helper function use directly.

It is a matter of taste and not blocking at all.

```
let wrapper

const findButton = () => wrapper.dive().find('DefaultButton').props().disabled

it('should do stuff', () => {
  wrapper = shallow(...)
  const btn = findButton()
  expect(btn.props().disabled).toBe(true)
})
```

Why is it necessary to `dive()` ?

packages/cozy-harvest-lib/src/components/AccountCreator.jsx
```diff
@@ -36,10 +36,8 @@ export class AccountCreator extends PureComponent {
     const { account, konnector, submitting } = this.props
     return (
       <AccountForm
-        fields={konnector.fields}
+        konnector={konnector}
```
ptbrowne: 🎉 

I think it is much simpler mentally to pass around known objects and it avoids creating a lot of props.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -104,57 +96,75 @@ export class TriggerManager extends Component {
...
+    const { account, error, status } = this.state
+    const submitting = status === RUNNING || running
+
+    return (
+      <AccountForm
+        error={error}
         konnector={konnector}
-        onBeforeCreate={this.handleAccountMutation}
-        onCreateSuccess={this.handleAccountCreationSuccess}
+        initialValues={account && (account.auth || account.oauth)}
```
ptbrowne: Pass directly the account no ? Why is it ncessary to destructure here ?

packages/cozy-harvest-lib/src/connections/accounts.js
```diff
@@ -109,14 +109,27 @@ const updateAccount = async (client, account) => {
...
 /**
  * Get accounts mutations
  * @param  {Object} client CozyClient
  * @return {Object}        Object containing accounts mutations
  */
 export const accountsMutations = client => ({
   createAccount: createAccount.bind(null, client),
-  updateAccount: updateAccount.bind(null, client)
+  updateAccount: updateAccount.bind(null, client),
+  saveAccount: saveAccount.bind(null, client)
```
ptbrowne: Why is updateAccount necessary here ? saveAccount already does both creating and saving ?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -104,57 +96,75 @@ export class TriggerManager extends Component {
...
-    this.setState({
-      status: SUCCESS
-    })
+    return onSuccess(trigger)
   }
 
   render() {
-    const { account, konnector, onDone, running } = this.props
-    const { createdAccount, status, updatedAccount } = this.state
-    const succeed = createdAccount && [LOGGED, SUCCESS].includes(status)
```
ptbrowne: No more LOGGED or SUCCESS status ? Is this done higher up ?

packages/cozy-harvest-lib/src/helpers/permissions.js
```diff
@@ -0,0 +1,7 @@
+import intersection from 'lodash/intersection'
+
+export const build = (doc, verbs) => ({
+  type: doc._type,
+  values: [doc._id],
+  verbs: intersection(verbs, ['DELETE', 'GET', 'POST', 'PUT'])
+})
```
ptbrowne: I do not see where this is used ?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -4,9 +4,8 @@ import PropTypes from 'react-proptypes'
 import { withMutations } from 'cozy-client'
 import { translate } from 'cozy-ui/react/I18n'
```
ptbrowne: This commit (73178ab190a997dc851a5ac36d11a341d36c9d35) is a bit hard to follow and I am not sure I can properly review it.

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
```diff
@@ -189,6 +212,52 @@ describe('TriggerManager', () => {
       wrapper.instance().handleSubmit()
       expect(wrapper.props().submitting).toEqual(true)
     })
+
+    it('should call saveAccount without account', () => {
+      const wrapper = shallowWithoutAccount()
```
ptbrowne: Nit : I find that it would preferable if this function would not two things at once (shallow rendering and passing the right props).

```
const wrapper = shallow(<TriggerManager {...props.noAccount} />)
```

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
```diff
@@ -189,6 +212,52 @@ describe('TriggerManager', () => {
...
+        fixtures.account
+      )
+    })
+
+    it('should call saveAccount with account', () => {
+      const wrapper = shallowWithAccount()
+      wrapper.instance().handleSubmit(fixtures.data)
+      expect(saveAccountMock).toHaveBeenCalledWith(
+        fixtures.konnector,
+        fixtures.createdAccount
```
ptbrowne: I find a bit weird that saveAccount is called with first argument a konnector ?

packages/cozy-harvest-lib/src/components/AccountCreator.jsx
gregorylegarec: Totally agree, it was a matter of time :)

packages/cozy-harvest-lib/src/connections/accounts.js
gregorylegarec: Yep, I first removed them but then I put them back as it was easier to test them (i.e just keep their related tests).

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
gregorylegarec: Nope. The triggerManager is either `IDLE` or `RUNNING`.

The success is handler higher up, via `onLoginSuccess` and `onSuccess` hooks.

packages/cozy-harvest-lib/src/helpers/permissions.js
gregorylegarec: Oh well spotted, it has been commited by mistake. Should go in our future permission model instead. I'll remove this file from the PR right now.

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
gregorylegarec: `createAccount` is needing a konnector, to know if it needs to create an aggregator account first.

A workaround could be to pass this information as an options or directly into the account.

But you're right, I think it's a right thing to avoid passing a konnector at this point of our code.

I suggest, if you agree, to keep it for a future (little) refactoring.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -66,29 +66,35 @@ export class TriggerManager extends Component {
...
+          accounts.getLabel(account)
+        )}`
+
+        folder =
+          (await statDirectoryByPath(path)) ||
+          (await createDirectoryByPath(path))
+
+        await addPermission(konnector, konnectors.buildFolderPermission(folder))
+        await addReferencesTo(konnector, [folder])
+      }
```
ptbrowne: Could be extracted outside the component in a  ensureKonnectorHasFolderIfNeeded(konnector) method

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -66,29 +66,35 @@ export class TriggerManager extends Component {
...
+          cron: cron.fromKonnector(konnector),
+          folder,
+          konnector
+        })
+      )
+
+      return await this.launch(trigger)
+    } catch (error) {
+      return this.handleError(error)
     }
```
ptbrowne: I think this business logic should be outside the React component.

packages/cozy-harvest-lib/src/components/AccountFormError.jsx
```diff
@@ -0,0 +1,83 @@
...
+import PropTypes from 'react-proptypes'
+
+import { translate } from 'cozy-ui/react/I18n'
+
+import { KonnectorJobError } from '../helpers/konnectors'
+import Markdown from './Markdown'
+
+/**
+ * Component used to show an error related to account or to konnector job
+ * Regular error are treated in a common way, but for KonnectorJobErrors
```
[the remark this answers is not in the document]
gregorylegarec: classic ?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
gregorylegarec: We agree, the next step is to start using Models.

packages/cozy-harvest-lib/src/components/AccountFormError.jsx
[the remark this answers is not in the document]
ptbrowne: What do you mean by classic or common ? :D

packages/cozy-harvest-lib/src/components/AccountCreator.jsx
y-lohse: Funny seeing this after the other review :p It's ok if the object is *known*, but for me at least `konnector` is opaque and I have no idea what is or isn't in it, so i actually prefer being explicit about the props. It can also help avoid extra renders when used in PureComponents.

packages/cozy-harvest-lib/src/components/AccountForm.jsx
```diff
@@ -284,9 +286,7 @@ export class AccountForm extends PureComponent {
 
 AccountForm.propTypes = {
   account: PropTypes.object,
-  fields: PropTypes.object,
-  oauth: PropTypes.object,
-  locales: PropTypes.object,
+  konnector: PropTypes.object.isRequired,
```
y-lohse: But also if we're passing a full object, we can describe it's shape here :-)

packages/cozy-harvest-lib/src/components/AccountFormError.jsx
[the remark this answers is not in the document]
ptbrowne: Ah ok it's clearer to me now

packages/cozy-harvest-lib/src/components/AccountFormError.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+    const args = {
+      name: konnector.name || '',
+      link: konnector.vendor_link || ''
+    }
+
+    if (error instanceof KonnectorJobError) {
+      return t(getKonnectorJobKey(error.code), {
+        ...args,
+        _: t(getKonnectorJobKey(error.type), {
```
y-lohse: What is this `_` key?

packages/cozy-harvest-lib/src/components/AccountFormError.jsx
```diff
@@ -0,0 +1,81 @@
...
+    return error.message ? `${description} (${error.message})` : description
+  }
+
+  render() {
+    const errorDescription = this.getDescription()
+    /* eslint-disable-next-line no-console */
+    console.error(errorDescription)
+    return (
+      <div className="u-error">
+        <Markdown source={this.getDescription()} />
```
y-lohse: We can use `errorDescription ` here.

packages/cozy-harvest-lib/src/components/AccountForm.jsx
```diff
@@ -193,8 +201,9 @@ export class AccountForm extends PureComponent {
    * @param  {Object} formState See https://github.com/final-form/final-form#formstate
    * @return {Boolean}
    */
-  isSubmittable({ dirty, initialValues, valid }) {
-    return valid && !(initialValues && !dirty)
+  isSubmittable({ dirty, error, initialValues, valid }) {
+    const untouched = initialValues && !dirty
+    return error || (valid && !untouched)
```
y-lohse: So it's submittable when there is an error?

packages/cozy-harvest-lib/src/connections/triggers.js
```diff
@@ -56,12 +57,15 @@ const waitForLoginSuccess = async (
...
 
-  return new Promise(resolve => {
+  return new Promise((resolve, reject) => {
     setTimeout(() => resolve(job), expectedSuccessDelay)
-    jobSubscription.onUpdate(
-      realtimeJob =>
-        JOB_END_STATES.includes(realtimeJob.state) && resolve(realtimeJob)
-    )
+    jobSubscription.onUpdate(updatedJob => {
+      return (
```
y-lohse: Why are we using `return` here?

packages/cozy-harvest-lib/src/components/AccountFormError.jsx
gregorylegarec: It allows to specify a fallback value is the key does not exist in translation file.
See (Polyglot documentation)[https://github.com/airbnb/polyglot.js?files=1#polyglotprototypetkey-interpolationoptions].

Here we are using this feature on three levels 💪, it spares us [this kind of mess](https://github.com/cozy/cozy-home/blob/master/src/lib/konnectors.js#L99)

packages/cozy-harvest-lib/src/components/AccountFormError.jsx
gregorylegarec: Oops, my bad. Very recent change. Well spotted !

packages/cozy-harvest-lib/src/components/AccountForm.jsx
gregorylegarec: Yes, first version was not, but actually, if you have an error somewhere you should be allowed to retry, i.e. to resubmit form.

packages/cozy-harvest-lib/src/connections/triggers.js
gregorylegarec: 😱

packages/cozy-harvest-lib/src/components/AccountForm.jsx
gregorylegarec: [Same remark](https://github.com/cozy/cozy-libs/pull/275#discussion_r258502959) as @Crash--, you definitely should meet each other.

gregorylegarec:
Thanks guys for your precious feedback !

PR #307, opened 2019-03-05
Harvest: allow float in randomDayTime() helper

gregorylegarec:
Banking konnectors are declaring float values in their time_interval
attribute.

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/daytime.js
+++ b/packages/cozy-harvest-lib/src/helpers/daytime.js
@@ -1,8 +1,8 @@
-import isInteger from 'lodash/isInteger'
+import isNumber from 'lodash/isNumber'
 /**
  * Returns an hour of the day between two hours given in parameters
- * @type {integer} min minimal start hour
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/helpers/daytime.js
```diff
@@ -11,8 +11,8 @@ export const randomDayTime = (
   end = 1,
   randomize = (min, max) => Math.random() * (max - min) + min
 ) => {
-  if (!isInteger(start)) throw new Error('Parameter start must be an integer')
-  if (!isInteger(end)) throw new Error('Parameter end must be an integer')
+  if (!isNumber(start)) throw new Error('Parameter start must be an integer')
+  if (!isNumber(end)) throw new Error('Parameter end must be an integer')
```
nono: [nit] The error message should reflect that: s/an integer/a number/

gregorylegarec: Oh thanks, I am a little bit tired.

PR #308, opened 2019-03-06
feat: Add method to detect incoherent createdByApp

ptbrowne:
In the banking connector, we send events to Matomo if we detect
accounts with incoherent data. The `hasIncoherentCreatedByApp`
method was previously there, but it is better if it is shared.

### DIFF
--- a/packages/cozy-doctypes/src/banking/BankAccount.spec.js
+++ b/packages/cozy-doctypes/src/banking/BankAccount.spec.js
@@ -49,3 +49,43 @@ describe('deleteDuplicateBankAccountsWithNoOperations', () => {
     expect(res.map(x => x._id)).toEqual(['empty'])
   })
 })
+
+describe('incoherences', () => {
+  const incoherent = {
+    institutionLabel: "Caisse d'épargne",
+    cozyMetadata: {
+      createdByApp: 'boursorama'
+    }
+  }
+
+  const coherent = {
+    institutionLabel: "Caisse d'épargne",
+    cozyMetadata: {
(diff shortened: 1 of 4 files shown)


packages/cozy-doctypes/src/banking/slug-account.js
```diff
@@ -0,0 +1,24 @@
...
+    if (ilabelRx[0] === '/' && ilabelRx[ilabelRx.length - 1] === '/') {
+      return [new RegExp(ilabelRx.substr(1, ilabelRx.length - 2), 'i'), slug]
+    } else {
+      return [ilabelRx, slug]
+    }
+  }
+)
+
+export const getSlugFromInstitutionLabel = institutionLabel => {
+  for (let [rx, slug] of institutionLabelsCompiled) {
```
kosssi: it's not important but here it's `const [rx, slug]`

packages/cozy-doctypes/src/banking/BankAccount.spec.js
```diff
@@ -49,3 +49,32 @@ describe('deleteDuplicateBankAccountsWithNoOperations', () => {
...
+  }
+
+  const unknown = {
+    institutionLabel: 'Unknown bank',
+    cozyMetadata: {
+      createdByApp: 'unknown1337'
+    }
+  }
+
+  it('should detect when slug is not coherent with createdByApp', () => {
```
kosssi: maybe you can write 2 more test without `createdByApp` key and another without `cozyMetadata` key

PR #311, opened 2019-03-07
Harvest: Ensure trigger can be launched, even after creation error

gregorylegarec:
After a creation error, re-submitting the form was
throwing an error, because the trigger in TriggerManager
was not saved in state. This PR solves this point in
the second commit.

Then it ensures that the trigger exists on update, in the
case of an error between account creation and trigger
creation.

At last, it merge `handleAccountCreationSuccess` and
`handleAccountUpdateSucces` into a single one handler
as they are now identical.

**NB: This PR concern a logic part being still in `TriggerManager`, but which need to be extracted in a dedicated logic component (a Model for example).**

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -26,11 +26,6 @@ const RUNNING = 'RUNNING'
  * @type {Component}
  */
 export class TriggerManager extends Component {
-  state = {
-    account: null,
-    status: IDLE
-  }
-
   constructor(props) {
     super(props)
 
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -91,35 +88,13 @@ export class TriggerManager extends Component {
...
    */
-  async handleAccountCreationSuccess(account) {
-    this.setState({
-      account
-    })
-
-    try {
-      const trigger = await this.ensureTrigger()
-      return await this.launch(trigger)
-    } catch (error) {
```
gregorylegarec: The try catch disappears, as we can rely on the try/catch into `handleSubmit`

PR #316, opened 2019-03-07
Harvest: Give focus to identifier field

gregorylegarec:
Needing to have the `inputRef` prop in Cozy-UI Field component, as in https://github.com/cozy/cozy-ui/pull/858


packages/cozy-harvest-lib/src/helpers/fields.js
```diff
@@ -46,6 +46,8 @@ export const sanitizeSelectProps = props => {
   delete sanitized.onKeyUp
   // Not applicable on Selects
   delete sanitized.autoComplete
+  // Only for inputs
+  delete sanitized.inputRef
```
ptbrowne:
```
const sanitized = omit(props, 'onKeyUp', 'inputRef', 'autoComplete')
```

ptbrowne: In the description of the function, it's written react-select and then Field component ? I guess it should read "Prepare props to pass to Field component" ? Could we use Field.propTypes to only pass relevant properties ?

```
const propTypes = pick(props, Object.keys(Field.propTypes))
```

?

gregorylegarec: Oh nice, I love this approach, thanks !

gregorylegarec: Actually, what we do is:
Renders Field, which renders SelectBox, which renders ReactSelect.

Removing unusable props should be done in `<Field />`.

gregorylegarec: So, `<ReactSelect />` component does not expose propTypes, so we cannot consider them in `<SelectBox />` component, and in `<Field />` component neither.

The best solution I found to clean props we pass to `<SelectBox />` is here. https://github.com/cozy/cozy-ui/pull/863

gregorylegarec: I removed the `inputRef` deletion and will remote the other line in a dedicated PR.

PR #318, opened 2019-03-08
Harvest: Handle onKeyUp on whole AccountFields component

gregorylegarec:
~~The underlying component in our SelectBox (ReactSelect) does
not allow handler for KeyUp events. The cheapest way to handle
them is to wrap our SelectBox into a div element to capture
keyUp events.~~

Move the onKeyUp handler to the AccountFields component main container.

EDIT:

Why keyUp ? To prevent browser for filling password in Cozy-Home, the "best" solution we found so far is to:
* Set the password autocomplete property to `new-password`.
* Do not provide a form wrapper and use a div isntead.

Using a div is forcing us to implement some form native behaviour, [like submitting on `Enter`](https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/components/AccountForm.jsx#L209) key, and for this we need to handle the `onKeyUp` event.

We are aware that replacing a form with a div is an anti-pattern but we did not find another option, and we are using this hack since the beginning of Cozy-Home.

This will disappear as soon as we will have Harvest applications with their own domain, which will prevent a form to be filled with another service information.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm.jsx
@@ -45,7 +45,6 @@ export class AccountField extends PureComponent {
       initialValue,
       label,
       name,
-      onKeyUp,
       required,
       role,
       t,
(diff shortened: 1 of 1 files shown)


ptbrowne:
What is the usecase ? When do you need to listen onKeyUp on a select ? Is the onChange handler not sufficient ?

gregorylegarec:
I edited the description to precise our need, it's a long known issue un Cozy-Home.

react-select is providing only `onKeyDown` handler. See https://react-select.com/props and https://github.com/JedWatson/react-select/blob/770ba08c6fcaba5190b0df9dfd4e2da98a2fef46/src/Select.js#L212

gregorylegarec:
@ptbrowne updated whit handler on the whole AccountFields component.

PR #319, opened 2019-03-08
fix: Use cozy-realtime 2.0 in harvest

nono:
In the package.json of cozy-harvest-lib, we were saying to use cozy-realtime 1.2.8, but as we are in a monorepo, it was already cozy-realtime 2 that was used in fact.

### DIFF
--- a/packages/cozy-harvest-lib/package.json
+++ b/packages/cozy-harvest-lib/package.json
@@ -23,7 +23,7 @@
   },
   "dependencies": {
     "cozy-client": "6.3.0",
-    "cozy-realtime": "1.2.8",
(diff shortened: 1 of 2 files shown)


CPatchane:
@nono there are some conflicts and the build is broken, could you fix the tests? If the version didn't change I don't see why the build is broken by the way :/

nono:
I don't understand why the tests fail, and I don't have the time to dig that. Let's close this PR.

gregorylegarec:
@nono I think it's due to the little changes in the API.

Thanks by the way.

By the way I worked on my spare time on a revamp of the realtime API. I'll make a PR tonight.

PR #321, opened 2019-03-09
Babel preset app options for presets and plugins

ptbrowne:
In this PR, more flexibility is given to users of babel-preset-cozy-app to configure presets and plugins that babel-preset-cozy-apps uses.

## babel-preset-env

To use webpack tree shaking features, we need to have our modules shipped on npm using es6 imports.

For `babel-preset-env` not to transpile import/exports, we have to pass `modules: false` https://babeljs.io/docs/en/babel-preset-env#modules.

`babel-preset-env` was not configurable when using `babel-preset-cozy-app` so this PR adds an option `presetEnv` to the options of `babel-preset-cozy-app`.

## transform-runtime

I wanted to experiment with `helpers: true` and cozy-ui and babel-preset-cozy-app did not give enough flexibility on the options pased. `transformRuntime` options are passed directly to `transform-runtime` plugin (they are merged with the defaults `{ regenerator: true, helpers: false }`.

## validation

`babel-preset-cozy-app` options were validated but were supposed to all be Booleans, in contrast to `presetEnv`/`transformRuntime` which are object.s To be able to add these options and keep existing validation, validation has been refactored to use validator functions and a config objects specifying for each field the kind of validator it needs.

### DIFF
--- a/packages/babel-preset-cozy-app/README.md
+++ b/packages/babel-preset-cozy-app/README.md
@@ -90,6 +90,32 @@ By default, this babel preset uses [`babel-plugin-transform-runtime`](https://ba
 }
 ```
 
+#### Preset and plugin options
+
+You can have control on the options passed to `babel-preset-env` and `babel-plugin-transform-runtime`:
+
+* `presetEnv` will be passed to `babel-preset-env`
+* `transformRuntime` will be passed to `babel-plugin-transform-runtime`
+
+```json
+{
+  "presets": [
+    ["cozy-app", {
+      "presetEnv": { "modules": false },
+      "transformRuntime": { "helpers": true }
+    }]
+  ]
+}
+```
+
+In this case, we do not want `preset-env` to touch to `import/export` and want the inlined Babel helpers
+to be replaced by imports from `babel-runtime`.
+
+See the options on the official docs :
+
+https://babeljs.io/docs/en/babel-preset-env#modules
+https://babeljs.io/docs/en/babel-plugin-transform-runtime#helpers 
+
 ## Community
 
 ### What's Cozy?
--- a/packages/babel-preset-cozy-app/index.js
+++ b/packages/babel-preset-cozy-app/index.js
@@ -40,7 +40,7 @@ const optionConfigs = {
 const validators = mapValues(optionConfigs, x => x.validator)
 const defaultOptions = mapValues(optionConfigs, x => x.default)
 
-module.exports = declare((api, options) => {
+const mkConfig = (api, options) => {
   const presetOptions = {
     ...defaultOptions,
     ...options
@@ -97,4 +97,11 @@ module.exports = declare((api, options) => {
   }
   config.plugins = plugins
   return config
-})
+}
+
+module.exports = declare(mkConfig)
+
+if (require.main === module) {
+  // eslint-disable-next-line no-console
+  console.log(JSON.stringify(mkConfig(null, {}), null, 2))
+}
--- a/packages/babel-preset-cozy-app/index.js
+++ b/packages/babel-preset-cozy-app/index.js
@@ -2,7 +2,7 @@
 
 const { declare } = require('@babel/helper-plugin-utils')
 const browserslist = require('browserslist-config-cozy')
-const { validate, isOfType } = require('./validate')
+const { validate, isOfType, deprecated } = require('./validate')
 const mapValues = require('lodash/mapValues')
 const merge = require('lodash/merge')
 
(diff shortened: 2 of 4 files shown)


packages/babel-preset-cozy-app/index.js
```diff
@@ -2,57 +2,90 @@
 
 const { declare } = require('@babel/helper-plugin-utils')
 const browserslist = require('browserslist-config-cozy')
+const { validate, isOfType } = require('./validate')
+const merge = require('lodash/merge')
 
-const browserEnv = {
+const mapValues = (obj, iter) => {
```
kosssi: why you don't use lodash [`mapValues`](https://lodash.com/docs/4.17.11#mapValues)?

ptbrowne: Because `babel-preset-cozy-app` is a very small module, I did not want to introduce lodash here.

packages/babel-preset-cozy-app/index.js
```diff
@@ -2,57 +2,90 @@
...
+  const res = {}
+  for (let k of Object.keys(obj)) {
```
kosssi: `k` is a const ;)

packages/babel-preset-cozy-app/index.js
```diff
@@ -2,57 +2,90 @@
...
+const validators = mapValues(optionConfigs, x => x.validator)
+const defaultOptions = mapValues(optionConfigs, x => x.default)
+
+const mkConfig = (api, options) => {
+  const presetOptions = merge(defaultOptions, options)
+
+  try {
+    validate(presetOptions, validators)
+  } catch (e) {
+    e.message = `babel-preset-cozy-app : Config validation error : ${e.message}`
```
kosssi: 👍

packages/babel-preset-cozy-app/validate.js
```diff
@@ -0,0 +1,19 @@
+const isOfType = type => (value, key) => {
+  if (!(typeof value === type)) {
+    throw new Error(`Error: "${key}" must be a ${type}`)
+  }
+}
+
+const validate = (obj, validators) => {
+  for (let key of Object.keys(obj)) {
```
kosssi: `key` is a const ;)

packages/babel-preset-cozy-app/index.js
kosssi: but it's already [here](https://github.com/cozy/cozy-libs/pull/321/files#diff-f61cc66dd92dad64f006e20057c023ebR6)...

packages/babel-preset-cozy-app/index.js
```diff
@@ -2,57 +2,90 @@
 
 const { declare } = require('@babel/helper-plugin-utils')
 const browserslist = require('browserslist-config-cozy')
+const { validate, isOfType } = require('./validate')
+const merge = require('lodash/merge')
```
kosssi: you don't import lodash.

ptbrowne:
Thanks, @drazik I have added documentation [here](https://github.com/cozy/cozy-libs/pull/321/commits/803ec6db39cf50cfd23f3890cdc4ede0cb0699c1).

packages/babel-preset-cozy-app/README.md
```diff
@@ -90,6 +90,32 @@ By default, this babel preset uses [`babel-plugin-transform-runtime`](https://ba
 }
 ```
 
+#### Preset options
+
+You can have control on the options passed to `babel-preset-env` and `babel-plugin-transform-runtime`:
+
+* `presetEnv` will be passed to `babel-preset-env`
+* `transformRuntime` will be passed to `babel-plugin-transform-runtime`
```
CPatchane: `transformRuntime` is a plugin option which is different from "preset options" categorised here.
Also, we already have a `transformRegenerator` option above, so it will be deprecated? Shouldn't we remove the existed one or maybe merge the both? If we pass an object it will be used as the plugin options, if we pass booleans it will enable/disable with the default config. What do you think? It's just about having a clearer API and documentation

packages/babel-preset-cozy-app/index.js
```diff
@@ -94,7 +102,7 @@ const mkConfig = (api, options) => {
       }
     ]
   ]
-  if (!node && transformRegenerator) {
```
CPatchane: Should we provide a way to sill be able to disable transform runtime here?
[Cozy-client](https://github.com/cozy/cozy-client/blob/master/.babelrc#L2) (and maybe some other libraries) seem to disable it but maybe we should not have to disable it for browser environment

packages/babel-preset-cozy-app/validate.js
```diff
@@ -0,0 +1,57 @@
...
+
+const isFalse = (value, key) => {
+  if (value !== false) {
+    throw new Error(`"${key}" must be false`)
+  }
+}
+
+const either = (...validators) => (value, key) => {
+  let errors = []
+  for (let validator of validators) {
```
kosssi: a new `for (let` :D

PR #324, opened 2019-03-12
feat: Disable a transform that makes React slower

nono:
See https://github.com/facebook/create-react-app/pull/5278


CPatchane:
Good shot! 👍

PR #327, opened 2019-03-12
feat: Support tag prefix

ptbrowne:
Since we have to be able to publish beta versions of customized
banks, from the customization repository, without publishing
other components, we introduce here the notion of prefix for
tags.

When publishing from a monorepo, instead of tagging only
with a version `1.0.0`, you tag with a prefix `cozy-banks/1.0.0`
and then you tell cozy-app-publish to only consider tags
with this prefix `cozy-app-publish --tag-prefix cozy-banks`.

### DIFF
--- a/packages/cozy-app-publish/cozy-app-publish.js
+++ b/packages/cozy-app-publish/cozy-app-publish.js
@@ -59,6 +59,10 @@ const program = new commander.Command(pkg.name)
     '--postpublish <script-path>',
     'Hook to process parameters just after publishing, typically to deploy app'
   )
+  .option(
+    '--tag-prefix <tag-prefix>',
+    'When publishing from a monorepo, only consider tags with tagPrefix, ex: cozy-banks/1.0.1.'
+  )
   .option('--yes', 'Force confirmation when publishing manually')
   .option(
     '--registry-url <url>',
@@ -92,7 +96,8 @@ try {
     yes: program.yes,
     registryUrl: program.registryUrl,
     space: program.space,
-    verbose: program.verbose
+    verbose: program.verbose,
+    tagPrefix: program.tagPrefix
   }).catch(handleError)
 } catch (error) {
   handleError(error)
@@ -135,6 +140,7 @@ async function publishApp(cliOptions) {
       registryToken: cliOptions.token,
       registryUrl: cliOptions.registryUrl,
       spaceName: cliOptions.space,
+      tagPrefix: cliOptions.tagPrefix,
       verbose: cliOptions.verbose,
       yes: cliOptions.yes
     })
--- a/packages/cozy-app-publish/cozy-app-publish.js
+++ b/packages/cozy-app-publish/cozy-app-publish.js
@@ -5,7 +5,6 @@
 const commander = require('commander')
 const colorize = require('./utils/colorize')
 const scripts = require('./index')
-const pickBy = require('lodash/pickBy')
 const capitalize = require('lodash/capitalize')
 
 const pkg = require('./package.json')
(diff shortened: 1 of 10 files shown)


packages/cozy-app-publish/cozy-app-publish.js
```diff
@@ -107,45 +114,35 @@ function _getPublishMode() {
...
+
+  const publishFn = scripts[publishMode]
+
+  console.log()
+  console.log(
+    `${colorize.bold(capitalize(publishMode))} ${colorize.blue('publish mode')}`
+  )
+  console.log()
+  return publishFn(
+    pickBy({
```
CPatchane: I think we don't need to use a `pickBy` here, I prefer to have the property `undefined` than missing in the options object.

packages/cozy-app-publish/cozy-app-publish.js
```diff
@@ -107,45 +114,35 @@ function _getPublishMode() {
...
+  }
+
+  const publishFn = scripts[publishMode]
+
+  console.log()
+  console.log(
+    `${colorize.bold(capitalize(publishMode))} ${colorize.blue('publish mode')}`
+  )
+  console.log()
+  return publishFn(
```
y-lohse: We're now missing the `try/catch` block around this call (https://github.com/cozy/cozy-libs/pull/327/files#diff-18ba38e9e92d14c179d04b2d7403e1cbL127)

packages/cozy-app-publish/cozy-app-publish.js
ptbrowne: The thing is that if you pass `undefined`, default options are overriden by undefined and this is not what we want here.

packages/cozy-app-publish/cozy-app-publish.js
ptbrowne: Yes, this is already taken care of here https://github.com/cozy/cozy-libs/blob/985908611148b58507430c99d5325a99f61f2764/packages/cozy-app-publish/cozy-app-publish.js#L87

packages/cozy-app-publish/cozy-app-publish.js
CPatchane: Ok I didn't go it. So could you just explain that in a comment? It would be better for code reading for later I think.

packages/cozy-app-publish/cozy-app-publish.js
y-lohse: Ah yes, allright!

packages/cozy-app-publish/cozy-app-publish.js
ptbrowne: In fact, turns out, following another refactor, that it is not needed. https://github.com/cozy/cozy-libs/pull/327/commits/91907b11fe820059b7bb9f3d9800a0402f7150c5

Thanks

packages/cozy-app-publish/package.json
```diff
@@ -38,5 +38,8 @@
     "setupFiles": [
       "./test/jestLibs/setupJest.js"
     ]
+  },
+  "engines": {
+    "node": ">=6"
```
CPatchane: This will stop running the script or just warn and continue?

gregorylegarec:
LGTM 👌

(Just the usual reminder to indicate a scope in the commit messages, like `feat(publish): Something new`)

packages/cozy-app-publish/package.json
ptbrowne: Warn and continue. I think we should not add unnecessary checks in our code. This is the built-in way to do node version checks.

CPatchane: Ok, it was just for my culture, let's do it like this 👍

PR #347, opened 2019-03-19
Interapp: Export correctly as default module

CPatchane:
As said in the readme, we should be able to do `import Intents from 'cozy-interapp'`


ptbrowne:
Is it a breaking change ?

CPatchane:
It's used by banks and sandbox so it is a breaking change. I change my commit to notice it 👍

CPatchane:
@ptbrowne I added both to avoid a breaking change just for this

ptbrowne:
:shipit: :)

PR #348, opened 2019-03-19
No old reconciliation

ptbrowne:
Filter out new transactions that are too old when reconciliating.

### DIFF
--- a/packages/cozy-doctypes/src/banking/BankingReconciliator.spec.js
+++ b/packages/cozy-doctypes/src/banking/BankingReconciliator.spec.js
@@ -161,10 +161,10 @@ describe('banking reconciliator', () => {
     ]
     const fetchedTransactions = [
       {
-        amount: -200,
-        label: 'Debit 200',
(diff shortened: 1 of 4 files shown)


packages/cozy-doctypes/src/banking/BankingReconciliator.spec.js
```diff
@@ -161,10 +161,10 @@ describe('banking reconciliator', () => {
     ]
     const fetchedTransactions = [
       {
-        amount: -200,
-        label: 'Debit 200',
+        amount: -400,
+        label: 'Debit 400',
         vendorAccountId: 2,
-        date: '2018-06-01T00:00' // prior to split date and doesn't exist, saved
+        date: '2018-06-22T00:00' // prior to split date and doesn't exist, saved
```
ptbrowne: Had to change the date otherwise it was filtered out.
Had to change the amount and label otherwise it was reconciliated with an existing transaction.

PR #349, opened 2019-03-19
fix(cozy-authentication): Remove lock file

Crash--:
In a monorepo the lock file goes to the root folder. No need to have this one here 

Seems that @y-lohse says the same thing here : https://github.com/cozy/cozy-libs/pull/338#pullrequestreview-215016993


ptbrowne:
Oups, missed it, thanks !

PR #350, opened 2019-03-19
feat(app-publish/downcloud): Use buildDir option

drazik:
The downcloud pre-hook was using a hardcoded `./build` value, ignoring
the `--build-dir` options that can be passed to cozy-app-publish CLI.

So if I passed `--build-dir something/build`, the downcloud hook was looking at `./build`, which doesn't exist. Now it looks at `something/build`.


ptbrowne:
Looking good, maybe add a little test ?

ptbrowne:
LGTM.

PR #354, opened 2019-03-22
feat: Add an utility to open or redirect for a deeplink

Crash--:
I don't know how to test this part since it's already a bit hacky by itself. 

This code is currently served by the stack, but we plan to use it also in the bar.

### DIFF
--- a/packages/cozy-device-helper/src/deeplink.js
+++ b/packages/cozy-device-helper/src/deeplink.js
@@ -1,0 +1,106 @@
+/**
+ * This file is used to open the native app from a webapp
+ * if this native app is installed
+ *
+ * From a webapp, we don't have any clue if a native app is installed.
+ * The only way to know that, is to try to open the custom link
+ * (aka cozydrive://) and if nothing happens (no blur) we redirect to
+ * the callback
+ *
+ * Firefox tries to open custom link, so we need to create an iframe
+ * to detect if this is supported or not
+ */
+const _createHiddenIframe = (target, uri, randomId) => {
+  const iframe = document.createElement('iframe')
+  iframe.src = uri
+  iframe.id = `hiddenIframe_${randomId}`
+  iframe.style.display = 'none'
+  target.appendChild(iframe)
+  return iframe
+}
+
+const openUriWithHiddenFrame = (uri, failCb) => {
+  const randomId = Math.random()
+    .toString(36)
+    .replace(/[^a-z]+/g, '')
+    .substr(0, 5)
+  window.addEventListener('blur', onBlur)
+  let iframe = _createHiddenIframe(document.body, 'about:blank', randomId)
+
+  const timeout = setTimeout(function() {
+    failCb()
+    window.removeEventListener('blur', onBlur)
+    iframe.parentElement.removeChild(iframe)
+  }, 500)
+
+  function onBlur() {
+    clearTimeout(timeout)
+    window.removeEventListener('blur', onBlur)
+    iframe.parentElement.removeChild(iframe)
+  }
+
+  iframe.contentWindow.location.href = uri
+}
+
+const openUriWithTimeoutHack = (uri, failCb) => {
+  const timeout = setTimeout(function() {
+    failCb()
+    target.removeEventListener('blur', onBlur)
+  }, 500)
+
+  //handle page running in an iframe (blur must be registered with top level window)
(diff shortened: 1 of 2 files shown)


packages/cozy-device-helper/src/deeplink.js
```diff
@@ -0,0 +1,105 @@
...
+    failCb()
+    handler && handler.remove()
+  }, 500)
+
+  let iframe = document.getElementById('hiddenIframe')
+  if (!iframe) {
+    iframe = _createHiddenIframe(document.body, 'about:blank')
+  }
+
+  var handler = window.addEventListener('blur', onBlur)
```
y-lohse: Why use `var` here?

packages/cozy-device-helper/src/deeplink.js
```diff
@@ -0,0 +1,105 @@
...
+  return iframe
+}
+
+const openUriWithHiddenFrame = (uri, failCb) => {
+  const timeout = setTimeout(function() {
+    failCb()
+    handler && handler.remove()
+  }, 500)
+
+  let iframe = document.getElementById('hiddenIframe')
```
y-lohse: There's no risk if the iframe already exists? Wouldn't it be safer to create a new iframe every time?

packages/cozy-device-helper/src/deeplink.js
```diff
@@ -0,0 +1,105 @@
...
+  iframe.id = 'hiddenIframe'
+  iframe.style.display = 'none'
+  target.appendChild(iframe)
+  return iframe
+}
+
+const openUriWithHiddenFrame = (uri, failCb) => {
+  const timeout = setTimeout(function() {
+    failCb()
+    handler && handler.remove()
```
y-lohse: I didn't know `handler.remove` could be a thing? [The documentation says `addEventListener` returns nothing](https://developer.mozilla.org/fr/docs/Web/API/EventTarget/addEventListener). What's the difference with `window.removeEventListener('blur', onBlur)`?

packages/cozy-device-helper/src/deeplink.js
```diff
@@ -0,0 +1,105 @@
...
+
+const openUriWithMsLaunchUri = (uri, failCb) => {
+  navigator.msLaunchUri(uri, undefined, failCb)
+}
+
+const checkBrowser = () => {
+  const isOpera = !!window.opera || navigator.userAgent.indexOf(' OPR/') >= 0
+  const ua = navigator.userAgent.toLowerCase()
+  const isSafari =
+    (~ua.indexOf('safari') && !~ua.indexOf('chrome')) ||
```
y-lohse: TIL https://www.joezimjs.com/javascript/great-mystery-of-the-tilde/ but uh… still a bit confused by this.

packages/cozy-device-helper/src/deeplink.js
Crash--: Old code, is old. Thank you. Fixed

packages/cozy-device-helper/src/deeplink.js
Crash--: Added a randomId

packages/cozy-device-helper/src/deeplink.js
```diff
@@ -0,0 +1,108 @@
...
+    .replace(/[^a-z]+/g, '')
+    .substr(0, 5)
+  window.addEventListener('blur', onBlur)
+  const timeout = setTimeout(function() {
+    failCb()
+    window.removeEventListener('blur', onBlur)
+  }, 500)
+
+  let iframe = document.getElementById(`hiddenIframe_${randomId}`)
+  if (!iframe) {
```
y-lohse: I think we might as well remove the check entirely now that the id is random?

packages/cozy-device-helper/src/deeplink.js
```diff
@@ -0,0 +1,108 @@
...
+    iframe = _createHiddenIframe(document.body, 'about:blank', randomId)
+  }
+
+  function onBlur() {
+    clearTimeout(timeout)
+    window.removeEventListener('blur', onBlur)
+    iframe.parentElement.removeChild(iframe)
```
y-lohse: We remove the iframe here, but we don't do it in the `timeout`, is that on purpose?

Issue #10, opened 2018-08-08
[realtime] Initialize and instantiate


kosssi:
The lib need to be instanciated and/or initialized to pass `url` and `token` to it, to avoid passing a config object to every call.

If we create an init function, we should write a reset function, because on mobile application we can unlink account.

All discussion on cozy-realtime repository:

---

gregorylegarec:

Thanks for removing cozy-client-js 🔥.

I feel that it's a little bit constraining to pass url and token at every subscribe call. May be we could initialize the realtime once as we do for the other cozy clients. something like :

```
realtime.init(url, token)
// Maybe we should instantiate it ?
// ... later
const subscription = realtime.subscribe('io.cozy.files')
```

What do you think ?

----

sebn:

> I feel that it's a little bit constraining to pass url and token at every subscribe call. May be we could initialize the realtime once as we do for the other cozy clients.

👍 `url` and `token` always _travel_ together, so they should probably be part of the same _config_ object, so we may reuse it, don't need to remember params order, and don't need to change function calls in case _config_ gets a new field.

Although I would prefer one of the following solutions:

1. Just pass the config objet to function calls, e.g. `realtime.subscribe(config, 'io.cozy.files')` (may be a first, easier to implement step)
2. Instanciate the realtime, e.g. `const realtime = Realtime.init()`, so consumer code doesn't need to know whether the result is built using a class or a closure. Module vs variable names always bother me in js, hence the capital R, but any other name would be fine by me. (probably out of the scope of this PR, so we may introduce it in a subsequent one)

In both case, we don't expose functions until they can actually be used so we don't get unexpected errors when calling them (a designer once gave me this example: "Tea or coffee? Oh sorry we don't have any of them...").

ptbrowne:
+1 for initting realtime.

> In both case, we don't expose functions until they can actually be used so we don't get unexpected errors when calling them (a designer once gave me this example: "Tea or coffee? Oh sorry we don't have any of them...").

I am not sure I agree with this. If you do not expose the methods and try to call subscribe you will have

```
> realtime.subscribe()
undefined is not a function
```

IMHO, this is not as useful as

```
> realtime.subscribe()
Error: Realtime should be initialized with a token and URL before being used. Perhaps have you forgotten the `realtime.init({ token, url })` call ?
```

sebn:
> I think cozy-client-js does not have its method exposed until exposition, and it was difficult for me to know where the bug was when I called the methods and they were not available.

[...]

PR #355, opened 2019-03-26
Realtime: Expose new API and then refactor a little

gregorylegarec:
## Context
Two weeks ago I talked with @y-lohse about its need of renewing the OAuth token in cozy-realtime. He was arguing that it was not easy to implement a new feature in the existing codebase. I took a look on the code that was first written by myself and damn he was right; the code was a mess, with global variables everywhere, error-prone and subject to race conditions.

I immediately felt guilty and stay obsessed with this lib, I took on me to refactor it a little on my free time. I spent a few night in writing robust end to end tests and then used a class to refactor all the code.

Today @kosssi asked me about realtime and I talked to him about the state of the codebase and this little refacto. That's why I am making this PR now.

## What's changing
The previous API was based only on functions, and was looking like:

```js
import realtime from 'cozy-realtime'


const config = {
  domain: 'cozy.tools:8080',
  token: 'abcdef'
} 

const accountCreationHandler = account => { console.debug('Account has been created', account)}

const accountSubscription = realtime.subscribe(config, 'io.cozy.accounts')
accountSubscription.onCreate(accountCreationHandler)


//...

accountSubscription.unsubscribe()

```
This API is problematic in many ways:

* We must pass a `config` object at every `subscribe()` call
* We can register only one handler for an given event on a given doctype
* We can only unsubscribe to all events related to a doctype

The new API fixes all those problems, and proposes a realtime initialization as suggested in this issue https://github.com/cozy/cozy-libs/issues/10

```js
import { CozyRealtime } from 'cozy-realtime'


const realtime = new CozyRealtime({ url: 'http://cozy.tools:8080', token: 'abcdef')

const accountCreationHandler = account => { console.debug('Account has been created', account)}

realtime.subscribe({ type: 'io.cozy.accounts'}, 'created', accountCreationHandler)

const otherHandler = account => { console.debug(account._id) }
realtime.subscribe({ type: 'io.cozy.accounts'}, 'created', otherHandler)


//...

// Unsubscribe only one handler
realtime.unsubscribe({ type: 'io.cozy.accounts' }, 'created', accountCreationHandler)
```

## New features
This PR introduces also two callbacks to pass to CozyRealtime at initialisation:
* `onError`: Called when a realtime error occurs (can be handler by app)
* `onDisconnect`: Called when CozyRealtime fails to reconnect three times in a row

## Lighter alternative for this PR
If this PR is too big to merge now, as it is not prioritary in our backlogs, we can only merge the first two commits which are introducing the new API based on the legacy code.

## First feedback by @y-lohse 
* Provide graceful disconnection, to call in a `componentDidUnmount` for example
* Handle the case of `this._socketPromise` being `null` in `CozyRealtime._sendSubscribeMessage`

## TODO
- [ ]  Update docs
- [ ] Remove legacy code and release version 3.0
- [ ] Create websocket only when they are one subscription and close websocket on the last unsuscribe


CPatchane:
We have to add a `BREAKING CHANGE:` label somewhere in a commit description since there will be a new major version for this API change

gregorylegarec:
The only breaking change is a `console.warn()` when using realtime the old way, so I don't know if we have to indicates it now or when the legacy code will be removed.

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -0,0 +1,297 @@
...
+   * Returns an instance of CozyRealtime. Can be used instead of
+   * `new CozyRealtime`.
+   * @static
+   * @param  {Object} options Object containing domain, secure, token and url
+   * parameters
+   * @return {CozyRealtime}         CozyRealtime instance
+   */
+  static init(options) {
+    return new CozyRealtime(options)
+  }
```
ptbrowne: I think we should remove this. It is idiomatic javascript to write `new`, there is no advantage in having this method, and having several methods to do 1 thing is confusing IMHO.

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -0,0 +1,297 @@
...
+      secure: [isBoolean],
+      token: [isRequired, isString],
+      url: [isRequiredIfNo(['domain']), isURL]
+    }).validate({ domain, secure, token, url })
+
+    this._domain = domain || new URL(url).host
+
+    this._secure = url ? isSecureURL(url) : secure
+    this._token = token
+    this._url = url
```
ptbrowne: Is it possible not to cache `url` `token` and `secure` and to always go through a cozy-client ?

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -0,0 +1,297 @@
...
+  isString,
+  isURL
+} from './helpers/Validator'
+
+const MAX_RETRIES = 3
+const RETRY_BASE_DELAY = 1000
+
+export class CozyRealtime {
+  /**
+   * Log were subcribe messages sent are recored
```
```suggestion
   * Log were subscribe messages sent are recored
```

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -0,0 +1,297 @@
...
+    this._stopListenningUnload()
+
+    if (!event.wasClean) {
+      this._retry(event)
+    }
+  }
+
+  /**
+   * Handle a socket opening, send the authentification message to cozy
+   * stack, and a subribe message for every subscriptions that have been made.
```
```suggestion
   * stack, and a subscribe message for every subscriptions that have been made.
```

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -0,0 +1,297 @@
...
+  static init(options) {
+    return new CozyRealtime(options)
+  }
+
+  /**
+   * Remove the given handler from the list of handlers for given
+   * doctype/document and event.
+   *
+   * @param  {String}  type      Document doctype to subscribe to
+   * @param  {String}  id        Document id to subscribe to
```
drazik: You can also apply the JSDoc change here

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -0,0 +1,297 @@
...
+    this._socketPromise = null
+    this._stopListenningUnload()
+
+    if (!event.wasClean) {
+      this._retry(event)
+    }
+  }
+
+  /**
+   * Handle a socket opening, send the authentification message to cozy
```
drazik: « authentication »

packages/cozy-realtime/src/RealtimeSubscriptions.js
```diff
@@ -0,0 +1,129 @@
...
+    this._handlers[key] = this._handlers[key] || []
+    if (!this._handlers[key].includes(handler)) {
+      this._handlers[key].push(handler)
+    }
+  }
+
+  /**
+   * Call all the handlers related to the given selector and event name.
+   * @param  {[type]} type      Document doctype
+   * @param  {[type]} id        Document id
```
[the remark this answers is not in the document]
ptbrowne: `{[type]}` also

nono:
Ping @gregorylegarec: what is the status of this PR?

kosssi:
@nono I'm working on the next version of cozy-realtime, we put this topic aside but I'm on it now, the goal is to have a connection only if we launched a subscribe

gregorylegarec:
As @kosssi is working on the next steps of this PR, I suggest to close it.
@kosssi just remind to take comment from reviewers into account :)

PR #360, opened 2019-03-28
fix: Upgrade Auth

Crash--:
This is the up to date Auth module


packages/cozy-authentication/MobileRouter.jsx
```diff
@@ -0,0 +1,128 @@
+import React, { Component } from 'react'
+import { Router, withRouter } from 'react-router'
+import Proptypes from 'prop-types'
+import Authentication from './src/Authentication'
+import Revoked from './src/Revoked'
+import { logException } from 'drive/lib/reporter'
```
ptbrowne: drive ? How does it work ?

packages/cozy-authentication/MobileRouter.jsx
```diff
@@ -0,0 +1,133 @@
...
+  async doOnboardingLogin(receivedState, code, instanceDomain, history) {
+    const localState = await readState()
+    const localSecret = await readSecret()
+
+    try {
+      if (localState !== receivedState) {
+        throw new Error('States are not equals')
+      }
+
+      const clientInfo = await secretExchange(
```
ptbrowne: nit: I think it is better if function names begin with a verb.

packages/cozy-authentication/MobileRouter.jsx
```diff
@@ -0,0 +1,133 @@
...
+        client_id,
+        client_secret
+      } = clientInfo
+
+      if (
+        !(localSecret === onboarding_secret && localState === onboarding_state)
+      )
+        throw new Error('exchanged informations are not good')
+
+      const token = await getAccessToken(
```
ptbrowne: Nit: I tend to use get for synchronous functions and fetch for asynchronous ones.

packages/cozy-authentication/MobileRouter.jsx
```diff
@@ -0,0 +1,133 @@
...
+        url: addProtocolToURL(instanceDomain),
+        token,
+        clientInfo,
+        router: history
+      })
+    } catch (error) {
+      this.props.onLogout()
+    } finally {
+      clearState()
+      clearSecret()
```
ptbrowne: Nit: It seems state/secret live together, I think it would make more sense to manage them together.

```
const { state, secret } = getLocalOnboardingInfos()

clearOnboardingInfos()
```

Nit: `state` is too generic

packages/cozy-authentication/MobileRouter.jsx
```diff
@@ -0,0 +1,133 @@
...
+      appIcon,
+      onboarding,
+      onboardingInformations,
+      onException
+    } = this.props
+
+    if (!isAuthenticated) {
+      if (checkIfOnboardingLogin(onboardingInformations)) {
+        /* We need to hide() the ViewController since the ViewController is still active
+        when the application cames from background (specialy on iOS)
```
ptbrowne: s/specialy/only for/

Your comments start/end seem off

```
// Either use
// this

/* Or this
 * when using jsDoc
 */
```

packages/cozy-authentication/MobileRouter.jsx
```diff
@@ -0,0 +1,133 @@
...
+        */
+        if (window.SafariViewController) window.SafariViewController.hide()
+        const { code, state, cozy_url } = onboardingInformations
+        this.doOnboardingLogin(state, code, cozy_url, history)
+        //we return null since the previous method is async
+        return null
```
ptbrowne: It is not very clear here why we return null, don't we want to show a spinner (Opening login...) or something ? I have a bad feeling from the imperative code inside the render. What happens if `render` is called twice ?

packages/cozy-authentication/MobileRouter.jsx
```diff
@@ -0,0 +1,133 @@
...
+        />
+      )
+    } else {
+      return <Router history={history}>{appRoutes}</Router>
+    }
+  }
+}
+MobileRouter.defaultProps = {
+  onException: e => {
+    console.warn('exeception', e) //eslint-disable-line no-console
```
ptbrowne: Please use a proper logger from `minilog`

s/exeception/exception

packages/cozy-authentication/MobileRouter.spec.jsx
```diff
@@ -0,0 +1,274 @@
...
+      />
+    )
+    expect(app).toMatchSnapshot()
+  })
+
+  it('should render the auth screen when the onboarding has not started', () => {
+    const onboardingInformations = {
+      code: null,
+      state: null,
+      cozy_url: null
```
ptbrowne: It is unfortunate that the user has to deal with underscores :/ IMHO it would be better if the component did this on its own and what is exposed to the user is cozyURL.

packages/cozy-authentication/MobileRouter.spec.jsx
```diff
@@ -0,0 +1,274 @@
...
+    const app = shallow(
+      <MobileRouter
+        client={jest.fn()}
+        history={{}}
+        isAuthenticated={false}
+        isRevoked={false}
+        onAuthenticated={jest.fn()}
+        onLogout={jest.fn()}
+        onboarding={{}}
+        onboardingInformations={{ code: null }}
```
ptbrowne: When looking at this, I am in trouble to understand what is the difference between onboarding and onboardingInformations :/

packages/cozy-authentication/src/Authentication.jsx
```diff
@@ -42,9 +43,16 @@ class Authentication extends Component {
...
       const cozyClient = this.context.client
       const { client, token } = await register(cozyClient, url)
-      onComplete({
+
+      const destructuredToken = {
+        accessToken: token.accessToken,
+        refreshToken: token.refreshToken,
+        scope: token.scope,
+        tokenType: token.tokenType
+      }
```
ptbrowne: I don't understand, what does destructured mean here ?

packages/cozy-authentication/src/steps/ButtonLinkRegistration.jsx
```diff
@@ -1,32 +1,77 @@
...
+      softwareID,
+      softwareVersion,
+      clientURI,
+      logoURI,
+      policyURI,
+      scope
+    })
+    const url = `https://manager.cozycloud.cc/cozy/create?pk_campaign=drive-${getPlatform() ||
+      'browser'}&onboarding=${onboardingObject}`
+    this.setState({ url })
```
ptbrowne: It's weird to have a setState here when `generateURL` returns something.

It is easier to reason about code when functions either have side effects or are pure. A consequence is that a function should either return something or have a side effect but not both.

I do not really understand why we need to setState since `state.url` is used for the Button href but since we use onClick, it seems we do not need `href`.

packages/cozy-authentication/src/steps/ButtonLinkRegistration.jsx
```diff
@@ -1,32 +1,77 @@
...
+        className={className}
+        subtle={subtle}
+        type={type}
+      />
+    )
+  }
+}
+
+ButtonLinkRegistration.propTypes = {
+  onboarding: PropTypes.object.isRequired
```
ptbrowne: I think we can find a clearer variable name: `onboardingConfig` ?

packages/cozy-authentication/src/steps/ButtonLinkRegistration.jsx
```diff
@@ -1,32 +1,77 @@
...
+      clientName,
+      redirectURI,
+      softwareID,
+      softwareVersion,
+      clientURI,
+      logoURI,
+      policyURI,
+      scope
+    })
+    const url = `https://manager.cozycloud.cc/cozy/create?pk_campaign=drive-${getPlatform() ||
```
ptbrowne: `drive` here :/

I think the manager used should be configurable.

packages/cozy-authentication/src/utils/onboarding.js
```diff
@@ -0,0 +1,122 @@
...
+  if (!secret) {
+    secret = generateSecret()
+    await writeSecret(secret)
+  }
+  let state = await readState()
+
+  if (!state) {
+    state = generateState()
+    await writeState(state)
+  }
```
ptbrowne: Here we manage state and secret separately. Does it make sense ? It forces us to have dual methods for state/secret. Maybe they could be managed together. What happens if one exists and not the other. Is it something that is normal ? I think that if there is a discrepancy, there is a logout. If both variables form a bundle, I think it would be clearer if the code reflected that more clearly.

packages/cozy-authentication/src/utils/onboarding.js
```diff
@@ -0,0 +1,122 @@
...
+  return encodeURIComponent(JSON.stringify(oauthData))
+}
+
+export const addProtocolToURL = instanceDomain => {
+  return `https://${instanceDomain}`
+}
+export const secretExchange = (secret, instanceDomain, client) => {
+  const response = client.stackClient.fetchJSON(
+    'POST',
+    addProtocolToURL(instanceDomain) + '/auth/secret_exchange',
```
ptbrowne: Why do we have to do this manually (manage the protocol etc...) ? Because the client has not yet been configured with a cozy domain ?

packages/cozy-authentication/src/utils/onboarding.js
```diff
@@ -0,0 +1,122 @@
...
+  return response
+}
+
+export const getAccessToken = async (
+  { client_id, client_secret },
+  instanceDomain,
+  code,
+  client
+) => {
+  const body = `grant_type=authorization_code&code=${code}&client_id=${client_id}&client_secret=${client_secret}`
```
ptbrowne: Clearer when using `querystring`.

```
const body = querystring.stringify({
  grant_type: 'authorization_code',
  ....
})
```

packages/cozy-authentication/src/utils/onboarding.js
```diff
@@ -0,0 +1,122 @@
...
+    addProtocolToURL(instanceDomain) + '/auth/access_token',
+    body,
+    {
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded'
+      }
+    }
+  )
+  const token = await getTokenRequest.json()
+  if (getTokenRequest.status !== 200) {
```
ptbrowne: If the getTokenRequest is in 404, the json() will fail right ? I think the status check should be done before the json() call.

PR #361, opened 2019-03-28
Add a cozy-mjml package for email templating

nono:
Cozy-mjml is a new package for sharing MJML components between mail templates in several projects. Currently, I've just added a single custom component to show how it will work later.

### DIFF
--- a/packages/cozy-mjml/README.md
+++ b/packages/cozy-mjml/README.md
@@ -4,3 +4,55 @@ Cozy-MJML
 Cozy-MJML is used for emails. It brings some custom components for MJML that we
 want to share between several projects (cozy-stack, clouderery, notifications,
 etc.).
+
+## How it works?
+
+Cozy-MJML exposes a single binary that behave like mjml CLI. The goal is to
+replace the mjml binary by this one on the cloudery: it will allow to use our
+custom MJML components with mjml-rails, even if the mjml CLI doesn't support
+custom components and mjml-rails has some expectations (like the version number
+of the CLI tools). This binary will also be used by the cozy-stack to generate
+HTML mails: the stack will run it like a konnector or a service.
+
+I've had to put some hacks in the `src/hacks` directory and in the webpack
+config to make the bundled version works. I'm not proud of that, if you
+suggestion on how to improve that, please, let me know.
+
+You can test in command line the output of the given example by running this
+command:
+
+```sh
+$ node src/index.js example.mjml
+```
+
+The result will be output on stdout. If you prefer to test with the bundled
+version, you can do:
+
+```sh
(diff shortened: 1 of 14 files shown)


nono:
[(𝐌) [NUA] ETU j'ai des emails personnalisé en fonction de mon offre](https://trello.com/c/vXXMFYl5/1071-%F0%9D%90%8C-nua-etu-jai-des-emails-personnalis%C3%A9-en-fonction-de-mon-offre)

packages/cozy-mjml/src/hacks/datauri-index.js
```diff
@@ -0,0 +1,3 @@
+// This file replaces node_modules/datauri/index.js, because the shebang
+// confuses webpack.
+module.export = require('datauri/module')
```
ptbrowne: webpack has a shebang loader  : https://www.npmjs.com/package/shebang-loader

packages/cozy-mjml/src/components/index.js
```diff
@@ -0,0 +1,7 @@
+const core = require('mjml-core')
+
+const components = [require('./MJCozyTitle')]
+
+components.forEach(component => {
+  core.registerComponent(component)
+})
```
ptbrowne: I am not a fan of side-effects in index.js. I understand the usecase here but I think I would prefer to be explicit and export a `register` function.

packages/cozy-mjml/test/components.spec.js
```diff
@@ -0,0 +1,15 @@
+// Loads the MJML custom components, by side effect
+require('../src/components')
```
ptbrowne: Thus here, `const registerComponents = require('../src/components').register; registerComponents()` ?

packages/cozy-mjml/README.md
```diff
@@ -0,0 +1,58 @@
...
+etc.).
+
+## How it works?
+
+Cozy-MJML exposes a single binary that behave like mjml CLI. The goal is to
+replace the mjml binary by this one on the cloudery: it will allow to use our
+custom MJML components with mjml-rails, even if the mjml CLI doesn't support
+custom components and mjml-rails has some expectations (like the version number
+of the CLI tools). This binary will also be used by the cozy-stack to generate
+HTML mails: the stack will run it like a konnector or a service.
```
ptbrowne: Should the email service be an always running server ? I guess the time to generate an email would be then lower than when starting the node process from scratch every time.

nono: There are some trade-offs: having an always running server would be faster, but it makes things more complicated to deploy (for self-hosted in particular), monitor, debug. And we don't send that many mails that the performance would be noticeable.

PR #362, opened 2019-03-28
Initialize realtime with uri from CozyClient

gregorylegarec:
It seems that recent updates on CozyClient uri broke realtime in Cozy-Harvest.

Capture d’écran 2019-03-28 à 21 03 46


packages/cozy-harvest-lib/src/connections/triggers.js
```diff
@@ -50,8 +50,7 @@ const waitForLoginSuccess = async (
   const jobSubscription = await subscribe(
     {
       token: client.client.token.token,
```
CPatchane: something to change in the future, it seems very weird :s

gregorylegarec: See it like a song.

PR #364, opened 2019-04-01
WIP fix: Do not use localStorage if not available

ptbrowne:
Flags can be used on NodeJS where localStorage is not available.
Protect all places where localStorage is used with a typeof check.

### DIFF
--- a/packages/cozy-flags/src/browser/flag.js
+++ b/packages/cozy-flags/src/browser/flag.js
@@ -7,6 +7,9 @@ export const prefix = 'flag__'
 export const getKey = name => prefix + name
 
 const listFlagLocalStorage = () => {
+  if (typeof localStorage !== 'undefined') {
+    return []
+  }
   return Object.keys(localStorage)
     .filter(x => x.indexOf(prefix) === 0)
     .map(x => x.replace(prefix, ''))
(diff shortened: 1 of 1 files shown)


packages/cozy-flags/src/browser/flag.js
```diff
@@ -7,6 +7,9 @@ export const prefix = 'flag__'
 export const getKey = name => prefix + name
 
 const listFlagLocalStorage = () => {
+  if (typeof localStorage !== 'undefined') {
```
cedricmessiant: here, the condition should be `typeof localStorage === 'undefined'` right?

packages/cozy-flags/src/browser/flag.js
```diff
@@ -24,6 +27,9 @@ class FlagStore {
   }
 
   fillFromLocalStorage() {
+    if (typeof localStorage !== 'undefined') {
```
cedricmessiant: same here, the condition should be `typeof localStorage === 'undefined'`?

cedricmessiant:
It seems that we have tests covering this ;-) https://travis-ci.org/cozy/cozy-libs/builds/514161451?utm_source=github_status&utm_medium=notification

ptbrowne:
Yes I need to work on this, I pushed the PR because it was lingering on my desktop, sorry for the inconvenience, I added "WIP" ;)

nono:
Ping

ptbrowne:
Fixed in https://github.com/cozy/cozy-libs/pull/453

PR #370, opened 2019-04-04
Harvest: Make code React compliant

gregorylegarec:
* Translated TriggerManager was missing stuff like contextTypes definition.
* KeyUp event (for ENTER key on form) was not handler in React
* `aria-disabled=false` on Field was actually enabling the CSS style for aria-disabled component.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
@@ -36,7 +36,7 @@ export class AccountForm extends PureComponent {
 
   handleKeyUp(event, { dirty, form, initialValues, valid, values }) {
     if (
-      event.code === 'Enter' &&
+      event.key === 'Enter' &&
       this.isSubmittable({ dirty, initialValues, valid })
     ) {
       this.handleSubmit(values, form)
--- a/packages/cozy-harvest-lib/src/index.js
+++ b/packages/cozy-harvest-lib/src/index.js
@@ -1,18 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import I18n from 'cozy-ui/react/I18n'
-import TranslatedAccountForm from './components/AccountForm'
 import TranslatedTriggerManager from './components/TriggerManager'
 
 export { KonnectorJobError } from './helpers/konnectors'
 
 const dictRequire = lang => require(`./locales/${lang}.json`)
-
-export const AccountForm = (props, context) => (
-  <I18n dictRequire={dictRequire} lang={context.lang}>
-    <TranslatedAccountForm {...props} />
-  </I18n>
-)
 
 export const TriggerManager = (props, context) => (
   <I18n dictRequire={dictRequire} lang={context.lang}>
@@ -25,6 +18,5 @@ TriggerManager.contextTypes = {
 }
 
 export default {
-  AccountForm,
   TriggerManager
 }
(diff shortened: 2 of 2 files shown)


packages/cozy-harvest-lib/src/components/AccountForm/AccountField.jsx
```diff
@@ -52,7 +53,7 @@ export class AccountField extends PureComponent {
     const isEditable = !(role === ROLE_IDENTIFIER && initialValue)
 
     // Cozy-UI <Field /> props
-    const fieldProps = {
```
CPatchane: Why should we do that? Falsy props should correctly be handled :/

packages/cozy-harvest-lib/src/index.js
```diff
@@ -19,6 +20,10 @@ export const TriggerManager = (props, context) => (
   </I18n>
 )
 
+TriggerManager.contextTypes = {
+  lang: PropTypes.string.isRequired
+}
+
```
CPatchane: Not for AccountForm as well?

packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
```diff
@@ -35,8 +35,10 @@ export class AccountForm extends PureComponent {
   }
 
   handleKeyUp(event, { dirty, form, initialValues, valid, values }) {
+    const key = event.code || event.key
```
ptbrowne: Is it because the event is differnt between preact and react ?

If so i think this can be more explicit by doing

```
const keyFromPreactOrReactEvent = event => event.code || event.key

const key = keyFromPreactOrReactEvent(event)
```

This way we do not need a comment. It could be done in a separate compat.js file also. I am a bit surprised than preact and react do not have the same structure for events :/

CPatchane: Did you try `event.persist()`? React uses synthetic event but I don't know for preact

packages/cozy-harvest-lib/src/components/AccountForm/AccountField.jsx
gregorylegarec: I'll remove this, and actually count on Cozy-UI to not stylize component with `aria-disabled=false`. See https://github.com/cozy/cozy-ui/pull/894

packages/cozy-harvest-lib/src/index.js
gregorylegarec: Actually it is not used anymore, and should not. I'll remove it from being exposed here.

packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
gregorylegarec: Actually `event.code` [seems to be a standard](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code), but [not yet implemented in React](https://github.com/facebook/react/issues/14102). I will do a few more test to see what's the best property to rely on.

packages/cozy-harvest-lib/src/components/AccountForm/AccountField.jsx
CPatchane: Let's use the last cozy-ui and remove this so 👍

packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
gregorylegarec: `event.key` is working in both environment. Let's keep it and only it. 👌

PR #376, opened 2019-04-10
docs: Add some docs to cozy-release's README

cedricmessiant:
Add some info about deps, how to and some warnings about stuff I had to do when I used cozy-release for `cozy-contacts`.

It may be nice to have a way to create releases for apps that are still in beta (`0.x.y`, see the warning about patch).

### DIFF
--- a/packages/cozy-release/README.md
+++ b/packages/cozy-release/README.md
@@ -28,19 +28,37 @@
 
 # What's cozy-release?
 
-TODO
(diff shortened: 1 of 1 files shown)


gregorylegarec:
Oh thanks <3
That was on my todo

gregorylegarec:
Actually, in [our git workflow](https://docs.cozy.io/en/references/git-flow/), a patch for a beta is just a new beta, but I could be missing a point.

PR #382, opened 2019-04-14
WIP Harvest: Refactor to prepare 2FA

gregorylegarec:
This PR prepares the 2FA implementation by creating a dedicated class KonnectorJobWatcher.

This KonnectorJobWatcher should next be responsible a handling any konnector job event, including 2FA.

This will need a on2FA callback at initialisation, then it will be handled by TriggerManager to show 2FA form.

At this point TriggerManager could directly instantiate the KonnectorJobWatcher.

I let you update test and implement the rest of the logic @CPatchane. Based on what is already existing (mutations and components), it should be easy. Just remember we do not want logic inside components. Have fun !

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -10,7 +10,6 @@ import { triggersMutations } from '../connections/triggers'
 import filesMutations from '../connections/files'
 import permissionsMutations from '../connections/permissions'
 import accounts from '../helpers/accounts'
-import { KonnectorJobError } from '../helpers/konnectors'
 import cron from '../helpers/cron'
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -142,26 +153,14 @@ export class TriggerManager extends Component {
...
-    try {
-      job = await waitForLoginSuccess(await launchTrigger(trigger))
-    } catch (error) {
-      return this.handleError(new KonnectorJobError(error.message))
-    }
-
-    this.setState({
-      status: IDLE
+    watchKonnectorJob(await launchTrigger(trigger), {
+      onError: this.handleError,
```
gregorylegarec: We still have to bind it into cosntructor

PR #383, opened 2019-04-15
Cozy URL already has scheme

aeris:


packages/cozy-mjml/src/components/MJFooter.js
```diff
@@ -12,7 +12,7 @@ class MJFooter extends core.BodyComponent {
       return `
         <mj-section padding="16px 0">
           <mj-column vertical-align="middle">
-            <mj-button href="https://${instance}" background-color="#fff" font-size="14px" color="#297ef2" text-decoration="none" font-weight="bold" padding="0" inner-padding="4px 16px" vertical-align="middle" border-radius="16px">
+            <mj-button href="${instance}" background-color="#fff" font-size="14px" color="#297ef2" text-decoration="none" font-weight="bold" padding="0" inner-padding="4px 16px" vertical-align="middle" border-radius="16px">
```
GoOz: If `instance` brings the protocol with it then you have to strip it for the label just below.
And the thing is the stack's instance variable doesn't have the protocol with it. We have to make things work the same way, because this fix will break on the stack.
By doing so you are actually moving the problem, not fixing it.

aeris: You're not able to know the scheme here (local test are on http://, self-hosting is not enforced to https://, etc), so I think stack must carry the information too (config ?)

nono:
Let's use #386 instead

PR #385, opened 2019-04-16
Rename instance to fqdn and add an url parameter

maximemenager:
This allows to distinguish the instance fqdn for display and url for
hrefs


nono:
Let's use #386 instead

PR #386, opened 2019-04-16
feat(cozy-mjml): Use an URL with a scheme for MJ-Footer

nono:


nono:
The travis build is red, but it is only the same errors as we have on master: it is not related to this PR.

nono:
[Le lien vers le cozy dans les mails est cassé](https://trello.com/c/qHn96P0j/1265-le-lien-vers-le-cozy-dans-les-mails-est-cass%C3%A9)

PR #397, opened 2019-04-24
Add Contact and Group models

cedricmessiant:
Add Contact model in `cozy-doctypes` package. Most of this code comes from `cozy-drive`

### DIFF
--- a/packages/cozy-doctypes/src/contacts/Contact.js
+++ b/packages/cozy-doctypes/src/contacts/Contact.js
@@ -54,6 +54,75 @@ class Contact extends Document {
   }
 }
 
+const ContactShape = PropTypes.shape({
+  _id: PropTypes.string.isRequired,
+  _type: PropTypes.string.isRequired,
+  fullname: PropTypes.string,
+  name: PropTypes.shape({
+    givenName: PropTypes.string,
+    familyName: PropTypes.string,
+    additionalName: PropTypes.string,
+    namePrefix: PropTypes.string,
+    nameSuffix: PropTypes.string
+  }),
+  birthday: PropTypes.string,
+  note: PropTypes.string,
+  email: PropTypes.arrayOf(
+    PropTypes.shape({
+      address: PropTypes.string.isRequired,
+      label: PropTypes.string,
+      type: PropTypes.string,
+      primary: PropTypes.bool
+    })
+  ),
+  address: PropTypes.arrayOf(
+    PropTypes.shape({
+      street: PropTypes.string,
+      pobox: PropTypes.string,
+      city: PropTypes.string,
+      region: PropTypes.string,
+      postcode: PropTypes.string,
+      country: PropTypes.string,
+      type: PropTypes.string,
+      primary: PropTypes.bool,
+      label: PropTypes.string,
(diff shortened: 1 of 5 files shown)


packages/cozy-doctypes/src/contacts/Contact.js
```diff
@@ -0,0 +1,37 @@
...
+
+    return ['givenName', 'familyName']
+      .map(part => name[part] || '')
+      .map(name => name[0])
+      .join('')
+      .toUpperCase()
+  }
+
+  // TODO: sadly we have different versions of contacts' doctype to handle...
+  // A migration tool on the stack side is needed here
```
Crash--: Is it still relevant? 

If yes maybe we should add this task to our next migration plan

packages/cozy-doctypes/src/contacts/Contact.js
```diff
@@ -0,0 +1,37 @@
+const getPrimaryOrFirst = property => obj => {
+  if (!obj[property] || obj[property].length === 0) return ''
+  return obj[property].find(property => property.primary) || obj[property][0]
+}
+
+class Contact extends Document {
+  static getInitials(name) {
```
Crash--: should we take a contact as the argument instead of the name?

Crash--: Can you add a jsdoc here?

packages/cozy-doctypes/src/contacts/Contact.js
cedricmessiant: oui, c'est prévu dans la carte Migration !

packages/cozy-doctypes/src/contacts/Contact.js
cedricmessiant: I thought about taking a contact as argument but the code looks a bit odd. Maybe we should remove the string case from here since it is not related to contacts at all but it will leave some sort of `getInitials` function in drive...

packages/cozy-doctypes/src/contacts/Contact.js
```diff
@@ -0,0 +1,37 @@
...
+      ? getPrimaryOrFirst('email')(contact).address
+      : contact.email
+  }
+
+  static getPrimaryCozy(contact) {
+    return Array.isArray(contact.cozy)
+      ? getPrimaryOrFirst('cozy')(contact).url
+      : contact.url
+  }
+}
```
Crash--: we should also add getDisplayName ?

cedricmessiant: I thought about it but I think that `getDisplayName` has more to do with recipients than with contacts (sometimes recipient is a contact, sometimes it is the cozy "owner" and maybe sometimes something else, I take a look)

packages/cozy-doctypes/src/index.js
```diff
@@ -15,6 +16,7 @@ module.exports = {
   BankAccount,
   BankingReconciliator,
   BankTransaction,
+  Contact,
```
y-lohse: We don't want to expose groups ?

cedricmessiant: :+1:

PR #403, opened 2019-04-26
fix: Expose FlagSwitcher only in browser context

### LINKED ISSUE #402: [flags] localStorage is used on Node, which throws an error
The following error is thrown in the cozy-banks service using `cozy-flags@1.6.0`:

```
localStorage is not defined
```

I don't have the time to fix it now so I'm just downgrading cozy-flags in Banks. I will look at it asap if nobody have the time before me.

drazik:
Exposing FlagSwitcher on node context was causing an error in Node
context because it's importing the flags browser implementation and
`localStorage` doesn't exist in this context.

Fixes #402 

This was making the banks' service crash.

### DIFF
--- a/packages/cozy-flags/src/index.js
+++ b/packages/cozy-flags/src/index.js
@@ -1,17 +1,20 @@
 /* global __ENABLED_FLAGS__ */
 
-import isNode from 'detect-node'
-export { default as FlagSwitcher } from './browser/FlagSwitcher'
+const isNode = require('detect-node')
 
 const flag = isNode
   ? require('./node/flag').default
   : require('./browser/flag').default
 
+if (!isNode) {
+  flag.FlagSwitcher = require('./browser/FlagSwitcher').default
+}
+
 /**
  * Enables a list of flags
  * @param {string[]} flagsToEnable
  */
-export function enableFlags(flagsToEnable) {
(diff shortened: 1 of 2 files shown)


packages/cozy-flags/src/index.js
```diff
@@ -1,17 +1,20 @@
...
+if (!isNode) {
+  flag.FlagSwitcher = require('./browser/FlagSwitcher').default
+}
+
 /**
  * Enables a list of flags
  * @param {string[]} flagsToEnable
  */
-export function enableFlags(flagsToEnable) {
+function enableFlags(flagsToEnable) {
```
ptbrowne: Why is the function not exported anymore ? It makes the build red : https://travis-ci.org/cozy/cozy-libs/builds/524933745#L1215

drazik: Actually this function was not documented and is a duplicate of `flag.enable`. I searched for `enableFlags` in our codebases and it's used nowhere (which makes sense). So I updated test to use `flag.enable` and leaved this function unexported. Is it OK for you ?

ptbrowne: OK

ptbrowne:
LGTM ! Thanks !

PR #405, opened 2019-04-26
Provide account state to reset session

CPatchane:
Set a `RESET_SESSION` state into account document if password/passphrase is changed on edition


ptbrowne:
I don't understand the goal of setting the state in the account document 😇? What is the konnector session ?

CPatchane:
@ptbrowne We need to tell the konnector to reset the stored sessions when the user changes the password. We use the fact that user is currently doing a manual connexion to reset the session and ask  for a new one (see also https://github.com/cozy/cozy-doctypes/pull/89)

PR #407, opened 2019-04-29
[Harvest] Minor fixes

CPatchane:
* Fix 2FA modal margin on mobile, we want the extra top margin only when this is not a full screen modal
* Improve 2FA errors messages


gregorylegarec:
Hiding the login error is not a minor fix, it should be done in a dedicated PR once all previous ones would have been merged

CPatchane:
I put the showError commit into another PR #412

PR #408, opened 2019-04-29
[Harvest] Reset the account state before running the konnector

CPatchane:
Since the konnector and Harvest use the account state to communicate about the connector running process, it's better to be sure to reset it each time before running a konnector, so let's Harvest doing it.

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/accounts.js
+++ b/packages/cozy-harvest-lib/src/helpers/accounts.js
@@ -68,7 +68,8 @@ export const build = (konnector, authData) => {
   return {
     auth: authData,
     account_type: konnector.slug,
-    identifier: manifest.getIdentifier(konnector.fields)
+    identifier: manifest.getIdentifier(konnector.fields),
+    state: null
   }
 }
 
@@ -90,11 +91,21 @@ export const mergeAuth = (account, authData) => ({
  * @return {object}           io.cozy.accounts attributes
  */
 export const updateTwoFaCode = (account, code) => ({
-  ...account,
-  // reset the state,
-  state: null,
+  // reset the state since the konnector is listening it
+  ...resetState(account),
   twofa_code: code
 })
+
+/**
+ * Reset the account state
+ * @param  {Object} account Account document
+ * @return {object}         Changed account document
+ */
+export const resetState = account => ({
+  ...account,
+  state: null
+})
+
 /**
  * Set a state to reset the konnector session into io.cozy.accounts document
  * only if necessary, if password/passphrase have changed
@@ -119,6 +130,7 @@ export default {
   isTwoFANeeded,
   isTwoFARetry,
   mergeAuth,
+  resetState,
   setSessionResetIfNecessary,
   updateTwoFaCode
 }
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -90,11 +91,21 @@ export const mergeAuth = (account, authData) => ({
...
+
+/**
+ * Reset the account state
+ * @param {[type]} account Account document
+ * @return {object}        Changed account document
+ */
+export const resetState = account => ({
+  ...account,
+  state: null
+})
```
ptbrowne: This code could be moved to cozy-libs/doctypes.

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -90,11 +91,21 @@ export const mergeAuth = (account, authData) => ({
...
-  // reset the state,
-  state: null,
+  // reset the state since the konnector is listening it
+  ...resetState(account),
   twofa_code: code
 })
+
+/**
+ * Reset the account state
+ * @param {[type]} account Account document
```
ptbrowne: [type] :)

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -90,11 +91,21 @@ export const mergeAuth = (account, authData) => ({
  * @return {object}           io.cozy.accounts attributes
  */
 export const updateTwoFaCode = (account, code) => ({
-  ...account,
-  // reset the state,
-  state: null,
+  // reset the state since the konnector is listening it
+  ...resetState(account),
```
ptbrowne: It's a bit surprising from the exterior that the `state` property (with no connection to 2fa, at least on the name) changes the state. I am sure it is surprising for me because I am not completely familiar with the business rules behind, but I think it would be more understandable like this :

```js
// This properties are special and reset the state
const shouldPropertyUpdateResetState = {
  twofa_code: true
}

export const setProperty = (account, name, property) => {
  const res = {...account, [property]: name }
  if (shouldPropertyUpdateResetState[property]) {
    res[state] = null
  }
  return res
}
```

It is still a bit surprising to have the update of a property change another but at least it is very explicit and you can have all the properties that do that at the same place. What do you think ?

ptbrowne:
Looking good, can you add tests please ?

packages/cozy-harvest-lib/src/helpers/accounts.js
gregorylegarec: We agree, it's what we aim to do as soon as possible, as we speak a few weeks ago.
For now we need to ship quickly, I suggest to keep thoses models/helpers in Harvest and extract them when everything will work.

packages/cozy-harvest-lib/src/helpers/accounts.js
CPatchane: Harvest will listen to a state `TWO_NEEDED` to ask for the 2FA code and put it in the account, it's why setting the code should reset the state here.
For now this is the only one case and your suggestion is for me not more understandable, maybe when we will have more special properties with state reset but for now it's not the case.

PR #409, opened 2019-04-30
feat(cozy-doctypes): Add getDisplayName for contacts

cedricmessiant:
Use fullname if there is any, else use primary email

### DIFF
--- a/packages/cozy-doctypes/src/contacts/Contact.spec.js
+++ b/packages/cozy-doctypes/src/contacts/Contact.spec.js
@@ -102,4 +102,124 @@ describe('Contact model', () => {
       expect(result).toEqual('https://arya.mycozy.cloud')
     })
   })
+
+  describe('getFullname function', () => {
+    it("should return contact's fullname", () => {
+      const contact = {
+        fullname: 'Doran Martell',
+        name: {
+          givenName: 'Do',
+          familyName: 'Martell'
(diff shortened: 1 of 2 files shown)


packages/cozy-doctypes/src/contacts/Contact.spec.js
```diff
@@ -102,4 +102,42 @@ describe('Contact model', () => {
...
+            address: 'doran.martell@dorne.westeros',
+            primary: true
+          }
+        ]
+      }
+      const result = Contact.getDisplayName(contact)
+      expect(result).toEqual('Doran Martell')
+    })
+
+    it("should return the contact's primary email if no fullname", () => {
```
edas: Are we sure of that? 

I would use fullname > givenName + familyName > email

Crash--: This is the current specified behavior. We asked this morning to have a written specification about that. We'll update this function later if needed ;)

cedricmessiant: I think we always have `fullname` if `givenName` and `familyName` are present but it costs almost nothing and it's obviously better. 
(We asked Thomas to clarify the order for the display name this morning)

ptbrowne:
Great

PR #414, opened 2019-04-30
Feat/universal link

Crash--:
Add support of Universal link for cozy-auth.

### DIFF
--- a/packages/cozy-authentication/src/MobileRouter.jsx
+++ b/packages/cozy-authentication/src/MobileRouter.jsx
@@ -37,6 +37,7 @@ export class MobileRouter extends Component {
   componentDidMount() {
     this.startListeningToClient()
     this.startHandlingDeeplinks()
+    this.startListeningToUniversalLinks()
     this.tryToReconnect()
   }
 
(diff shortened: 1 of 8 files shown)


packages/cozy-authentication/src/utils/deeplink.js
```diff
@@ -12,8 +12,29 @@ const getDeeplink = () => {
   return window.deeplinkURL
 }
 
+const generateRoute = (url, appInfos) => {
+  const { protocol, universalLinkDomain, appSlug } = appInfos
+  const urlObj = new URL(url)
+
+  //remove fallback params not needed anymore at this stage
```
ptbrowne:
```suggestion
  // Remove fallback params not needed anymore at this stage
```

packages/cozy-authentication/src/MobileRouter.jsx
```diff
@@ -69,15 +72,32 @@ export class MobileRouter extends Component {
...
   }
 
   update() {
     this.forceUpdate()
   }
+  startHandlingUniversalLinks(eventData) {
+    /* 
+   openUniversalLink seems to be called only on iOS.
+   android uses handleOpenURL by default ?!
+   */
```
ptbrowne: I find this comment not very clear. Is it to explain something ? Is it a TODO ? Please clarify

packages/cozy-authentication/src/utils/deeplink.js
```diff
@@ -12,8 +12,29 @@ const getDeeplink = () => {
...
+    urlObj.searchParams.delete('fallback')
+  }
+  let newUrl = urlObj.toString()
+  let path = ''
+  if (newUrl.startsWith('http')) {
+    path = newUrl.replace(universalLinkDomain + '/' + appSlug, '')
+  } else {
+    path = newUrl.replace(protocol, '')
+  }
+  if (path.startsWith('/')) path = path.substring(1)
```
ptbrowne: I think you can work with the urlObj a bit more without resorting to string replace that are a bit brittle. You can for example modify the protocol directly on the urlObj object `urlObj.protocol = ''` .

packages/cozy-authentication/src/MobileRouter.jsx
Crash--: it's a todo. I need to double check the behavior on Android.

PR #415, opened 2019-05-01
Harvest: Facilitate trigger launching and two factor authentification with TriggerLauncher

gregorylegarec:
This PR prepares the code for and add the TriggerLauncher component.

Using render props, it allows to easily launch a trigger and handle all steps of a konnector job lifecycle.

To launch a konnector job with a button:

```jsx
render(){
  // trigger has already been fetched
  const { trigger } = this.props
  return (
    <TriggerLauncher trigger={trigger}>
      {({launch, running}) => (
        <Button
          onClick={launch}
          disabled={running}
        />
      )}
    </TriggerLauncher>
```

TriggerLauncher relies internally on a `KonnectorJob` object. `KonnectorJob` is an object which handles a konnector job lifecycle from launch to success or error. It use MicroEE mixins to emit events.

```js

// trigger has already been fetched
const konnectorJob = new KonnectorJob(trigger)
konnectorJob
  .on('success', handleSuccess)
  .on('twoFARequest', handleTwoFARequest)
konnectorJob.launch()

```

KonnectorJob for now relies internally on existing classes KonnectorJobWatcher and KonnectorAccountWatcher, they should both be merged directly into KonnectorJob. But as they are also used in TriggerManager to handle 2FA, we prefer to not introduce regressions in this PR and keep this refactoring for a next PR. In this future PR TriggerManager will lose all logic related to 2FA, which it should not handle.

Tests for TriggerManager and KonnectorJob are coming in a future PR as well.


packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -0,0 +1,140 @@
...
+  trigger: PropTypes.object.isRequired
+}
+
+TriggerLauncher.contextTypes = {
+  client: PropTypes.object.isRequired
+}
+
+export default withMutations(accountsMutations, triggersMutations)(
+  TriggerLauncher
+)
```
CPatchane: Could we have a test for this new component? At least a snapshot will be a good beginning than nothing

packages/cozy-harvest-lib/src/components/TwoFA/Modal.jsx
```diff
@@ -0,0 +1,121 @@
...
+      </Modal>
+    )
+  }
+}
+
+TwoFAModal.contextTypes = {
+  client: PropTypes.object
+}
+
+export default translate()(withMutations(accountsMutations)(TwoFAModal))
```
CPatchane: At least a snapshot would be great to avoid regressions when merging the two TwoFA components in the future

packages/cozy-harvest-lib/src/models/KonnectorJob.js
```diff
@@ -0,0 +1,96 @@
...
+    // KonnectorAccountWatcher into KonnectorJob
+    konnectorJob.on(ERROR_EVENT, this.reEmit(ERROR_EVENT))
+    konnectorJob.on(LOGIN_SUCCESS_EVENT, this.reEmit(LOGIN_SUCCESS_EVENT))
+    konnectorJob.on(SUCCESS_EVENT, this.reEmit(SUCCESS_EVENT))
+  }
+}
+
+MicroEE.mixin(KonnectorJob)
+
+export default KonnectorJob
```
CPatchane: Could we have a test for this model, since it will be the main used model, we have to have to start some tests here

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -0,0 +1,150 @@
...
+    this.setState({ status: PENDING })
+
+    this.konnectorJob = new KonnectorJob(client, trigger)
+    this.konnectorJob
+      .on('error', this.handleError)
+      .on('loginSuccess', this.handleLoginSuccess)
+      .on('success', this.handleSuccess)
+      .on('twoFARequest', this.handleTwoFARequest)
+      .on('twoFAMismatch', this.handleTwoFAMismatch)
+    this.konnectorJob.launch()
```
ptbrowne: Observations:

- In launch we are assigning this.konnectorJob, this means that TriggerLauncher has an internal state and it is not cleaned up on success or error. Do we need to save an internal variable ? Could we simplify and only use a closure ?
 - The component is responsible to remember the status of the konnector in its internal state. I wonder if it would no be simpler to have this status in the konnectorJob and rely on it for the render.

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  render() {
+    const { showTwoFAModal, status } = this.state
+    const { account, children, konnector, submitting } = this.props
```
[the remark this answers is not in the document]
gregorylegarec: I agree. As `submitting` is also used in TriggerManager, I suggest to keep it now for consistence and rename all props at the same time. I made an issue: https://github.com/cozy/cozy-libs/issues/431.

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -0,0 +1,150 @@
...
+
+    return (
+      <div>
+        {children({
+          launch: this.launch,
+          running: ![ERRORED, IDLE, SUCCESS].includes(status) || submitting
```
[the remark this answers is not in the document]
gregorylegarec: How I see it: If we introduce another status we should know it and add it in our model, which strengthen your suggestion to rely totally on konnectorJob to know get information about job status.

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -0,0 +1,150 @@
...
+        })}
+        {showTwoFAModal && (
+          <TwoFAForm
+            account={account}
+            konnector={konnector}
+            dismissAction={this.handleTwoFAModalDismiss}
+            handleSubmitTwoFACode={this.handle2FACode}
+            submitting={status === RUNNING_TWOFA}
+            retryAsked={status === TWO_FA_MISMATCH}
```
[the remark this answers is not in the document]
gregorylegarec: I was asking myself the same kind of question during development. What made me choose this solutions was to avoid passing a complex object like a KonnectorJob as prop. But I am still not sure this was a good solution.

How I see it if we pass the konnectorJob to TwoFAModal:
* TriggerLauncher must still listen to `twoFARequest` event to show 2FA modal
* TwoFAModal only listen to `twoFAMismatch` event

Do you agree ?

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
gregorylegarec: Seems clever. The konnectorJob is saved as an interanl variable to have access to `konnectorJob.sendTwoFACode` method to pass it to TwoFAModal.

Keeping the status in konnectorJob could also spare us all the `handle{KonnectorJobEvent}` methods here.

gregorylegarec:
@ptbrowne Thanks for your feedback 👍

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -17,6 +17,7 @@ const TWO_FA_REQUEST = 'TWO_FA_REQUEST'
 const TWO_FA_MISMATCH = 'TWO_FA_MISMATCH'
 const PENDING = 'PENDING'
 const SUCCESS = 'SUCCESS'
+const RUNNING_TWOFA = 'RUNNING_TWOFA'
```
gregorylegarec: `RUNNING_TWOFA` can be deduced with `TWO_FA_REQUEST`and `TWO_FA_MISMATCH` and should not exist, at least not at this level.

Letting konnectorJob manage statuses and relying on it may spare us listing those statuses here, as @ptbrowne suggests.

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -77,6 +79,11 @@ export class TriggerLauncher extends Component {
     this.setState({ status: TWO_FA_MISMATCH, showTwoFAModal: true })
   }
 
+  handle2FACode(code) {
```
gregorylegarec: Should definitely not go here. That's why I duplicated the `TwoFAForm` on the first place.

CPatchane: But the duplicate didn't work as well. What's the difference of passing a props and passing a method here?

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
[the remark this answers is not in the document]
gregorylegarec: @CPatchane `account` and `konnector` are data which can be retrieved with trigger infos. As konnectorJob is already fetching the related `account`, I suggest to also fetch the related konnector in it.

Then we pass the whole konnectorJob to TwoFAModal or we get account and konnector from it.

We should try to limit the number of props which can be deduced with other ones now, otherwise we will soon have a new Cozy-Home into Harvest.

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
CPatchane: Agree about the status but we still need the internal variable for the `sendTwoFACode` prop passed to the form, I don't see another simpler way to handle it :/

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
CPatchane: Not relevant anymore

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
CPatchane: Done in https://github.com/cozy/cozy-libs/pull/415/commits/fee5d4745535bc71fb6badaf248b2b3fcfb847d3

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
CPatchane: See https://github.com/cozy/cozy-libs/pull/415/commits/fee5d4745535bc71fb6badaf248b2b3fcfb847d3

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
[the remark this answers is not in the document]
CPatchane: Now the TriggerLauncher just use KonnectorJob from props with https://github.com/cozy/cozy-libs/pull/415/commits/fee5d4745535bc71fb6badaf248b2b3fcfb847d3

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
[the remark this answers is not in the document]
CPatchane: Handled directly in KonnectorJob now

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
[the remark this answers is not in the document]
CPatchane: See https://github.com/cozy/cozy-libs/pull/415/commits/fee5d4745535bc71fb6badaf248b2b3fcfb847d3

gregorylegarec:
Thanks @CPatchane 👍

PR #418, opened 2019-05-02
feat(cozy-doctypes): Add isContact method to Contact model

cedricmessiant:
We can now use the model to determine if an object is a contact or not

### DIFF
--- a/packages/cozy-doctypes/src/contacts/Contact.js
+++ b/packages/cozy-doctypes/src/contacts/Contact.js
@@ -9,12 +9,17 @@ const getPrimaryOrFirst = property => obj => {
   return obj[property].find(property => property.primary) || obj[property][0]
 }
 
+/**
+ * Class representing the contact model.
+ * @extends Document
+ */
 class Contact extends Document {
   /**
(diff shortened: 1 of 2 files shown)


ptbrowne:
I am curious about the usecase here ?

cedricmessiant:
> I am curious about the usecase here ?

In drive, we use a custom `Contact` model that also handles non-contact objects (for the sharing's recipients). I added `isContact` to this custom model but we think it may be useful to have it in the Contact model. See https://github.com/cozy/cozy-drive/pull/1640

edas:
> In drive, we use a custom `Contact` model that also handles non-contact objects (for the sharing's recipients). I added `isContact`

isn't `instanceof Contact` enough and more generic for this?

cedricmessiant:
> isn't `instanceof Contact` enough and more generic for this?

`Contact` is a model of contacts but our contacts objects are not instances of this model (the model is just an object that provides helpers as static methods). Am I right @ptbrowne ?

edas:
Oh, you're right, sorry for that

PR #430, opened 2019-05-06
feat: Show LogIn view while logging in

ptbrowne:
When the MobileRouter mounts, it tries to reconnect, based on information it has saved in localStorage.

`client.login` being async (since it instantiates the links, and some of them can be asynced, for example PouchLink), it can take a while to `login()` (hopefully not so much).

In the past, while logging in, we would show the `AuthenticationView`. This was a problem since the user had already logged in, and it is not expected to see a login form when you have already logged in.

Now, we show a component showing a spinner. It can be customized via the LogInComponent prop.

Another problem was a flash of AuthenticationView since the `tryToReconnect` is done after mouting. To mitigate the problem, the component shows null while it has not tried to reconnect yet. It will be during a very short time (a frame).

### DIFF
--- a/packages/cozy-authentication/src/MobileRouter.jsx
+++ b/packages/cozy-authentication/src/MobileRouter.jsx
@@ -52,7 +52,10 @@ export class MobileRouter extends Component {
     this.handleBeforeLogout = this.handleBeforeLogout.bind(this)
     this.handleRequestLogout = this.handleRequestLogout.bind(this)
 
-    this.state = { isLoggingInViaOnboarding: false }
+    this.state = {
+      isLoggingInViaOnboarding: false,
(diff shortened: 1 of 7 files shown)


packages/cozy-authentication/src/MobileRouter.jsx
```diff
@@ -269,7 +318,9 @@ MobileRouter.defaultProps = {
   components: {
     Authentication,
     Revoked
-  }
+  },
+
+  LoginInComponent
 }
 
 MobileRouter.propTypes = {
```
Crash--: can we add `initialTriedToReconnect` & LoginInComponent  to propTypes ?

PR #434, opened 2019-05-07
Harvest: Handle login success state to disable login success timeout

CPatchane:
When the konnector handle the new `LOGIN_SUCCESS` feature, it will set the state `HANDLE_LOGIN_SUCCESS_STATE ` to disable the login timeout in Harvest


gregorylegarec:
Ok for me as long as we do not wait too long to switch TriggerManager to a version relying entirely on KonnectorJob 👍

PR #436, opened 2019-05-09
Harvest: Use <Infos/> for error messages

CPatchane:
Display errors using the Infos component from CozyUI
Screenshot 2019-05-09 at 15 07 23

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm/Error.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/Error.jsx
@@ -2,17 +2,17 @@ import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 
 import { translate } from 'cozy-ui/react/I18n'
+import Infos from 'cozy-ui/react/Infos'
 
 import { KonnectorJobError } from '../../helpers/konnectors'
 import Markdown from '../Markdown'
 
 /**
- * Return the locale key corresponding to error description for a given error
- * code.
+ * Return the locale key corresponding to error for a given error code
  * @param  {string} code Error code as returned by a konnector job
  * @return {string}      locale key
  */
-const getKonnectorJobKey = code => `error.job.${code}.description`
+const getKonnectorJobKey = code => `error.job.${code}`
 
 /**
  * Component used to show an error related to an account or to a konnector job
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/AccountForm/Error.jsx
```diff
@@ -36,9 +36,9 @@ export class AccountFormError extends PureComponent {
    * @param  {Error} error      The error
    * @param  {Object} konnector konnector related to this error
    * @param  {Func} t           Translation function, expected to be Polyglot.t()
-   * @return {String}           The error description
+   * @return {Object}           The error locale object (title + description)
    */
-  getDescription() {
+  getErrorLocale() {
```
gregorylegarec: I am not very comfortable with the closure decltared in the middle of the method.

I think I'll do something like:

```jsx

createSuffixApppender = suffixKey => key => `${key}.${suffixKey}`

[...]

getErrorLocale(suffixKey) {
  const { error, konnector, t } = this.props

  const args = {
    name: konnector.name || '',
    link: konnector.vendor_link || ''
  }

  const withSuffix = createSuffixApppender(suffixKey)

  if(error instanceof KonnectorJobError) {
    return t(withSuffix(error.code), {
      ...args,
      _: t(withSuffix(error.type), {
        ...args,
        _: t(withSuffix(this.defaultKey), args)
      })
    })
  }
}
```

And in render method:

```jsx
    // [...]
    return (
      <Infos
         title={this.getErrorLocale('title')}
         text={<Markdown source={this.getErrlrLocale('description')} />}
         isImportant
      />
   )
```

I find this much simpler and readable. What do you think ?

CPatchane: It's like something I was thinking about but what do you do with this part? https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/components/AccountForm/Error.jsx#L59-L60

gregorylegarec: I guess something like:
```jsx
    // [...]
	const description = this.getErrlrLocale('description')
    return (
      <Infos
         title={this.getErrorLocale('title')}
         text={<Markdown source={error.message ? `${description} (${error.message})` : description
} />}
         isImportant
      />
   )
```

CPatchane: Fixed

PR #438, opened 2019-05-10
Harvest: Prevent refresh on 2FA submit with enter

CPatchane:
~~Returning false in a form submit function allow to disable refreshing the page on submitting with enter key~~
Edit: Finally, after testing, a `e.preventDefault` is doing the job

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
@@ -21,7 +21,10 @@ export class TwoFAForm extends PureComponent {
     this.setState({ twoFACode: e.currentTarget.value })
   }
 
-  handleSubmit() {
+  handleSubmit(e) {
+    // prevent refreshing the page when submitting
+    // (happens not systematically)
+    e.preventDefault()
     this.props.konnectorJob.sendTwoFACode(this.state.twoFACode)
   }
 


packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
```diff
@@ -23,6 +23,9 @@ export class TwoFAForm extends PureComponent {
 
   handleSubmit() {
     this.props.konnectorJob.sendTwoFACode(this.state.twoFACode)
+    // return false prevent refreshing the page when submitting the
+    // form with the button enter
+    return false
```
edas: Tu ne cherches pas plutôt un preventDefault() ?

CPatchane: Ça ne semble pas fonctionner sur le onSubmit du formulaire, cette solution est plus solicitée, comme c'est un cas que je reproduis rarement finalement j'y reviendrai si je le recroise

packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
```diff
@@ -21,8 +21,11 @@ export class TwoFAForm extends PureComponent {
     this.setState({ twoFACode: e.currentTarget.value })
   }
 
-  handleSubmit() {
+  handleSubmit(e) {
     this.props.konnectorJob.sendTwoFACode(this.state.twoFACode)
+    // return false prevent refreshing the page when submitting
```
edas: remove the comment ?

packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
```diff
@@ -21,8 +21,11 @@ export class TwoFAForm extends PureComponent {
     this.setState({ twoFACode: e.currentTarget.value })
   }
 
-  handleSubmit() {
+  handleSubmit(e) {
     this.props.konnectorJob.sendTwoFACode(this.state.twoFACode)
+    // prevent refreshing the page when submitting
+    // (happens not systematically)
+    e.preventDefault()
```
gregorylegarec: I think calling `e.preventDefault()` as first instruction is a good convention (At least a personal convention), could we do this here ?

ptbrowne: Also I find comments like `     // (happens not systematically) ` not very useful, it is fear and doubt inducing.

CPatchane: Keeping just `prevent refreshing the page when submitting` means that the issue happens systematically which is not the case here and we could remove this part thinking that we don't need it anymore due to a false positive

PR #439, opened 2019-05-13
Harvest: Make 2FA modals not closeable

CPatchane:
To quickly avoid users being blocked by closing the 2FA modal (we don't have the resume feature yet)


CPatchane:
[Ne pas fermer la modale duu code 2FA](https://trello.com/c/6gSvsF0h/602-ne-pas-fermer-la-modale-duu-code-2fa-autrement-quen-cliquant-sur-la-croix)

PR #443, opened 2019-05-14
WIP [realtime] feat: Subscribe does not throw if cannot connect, only log

ptbrowne:
Not sure if this is really wanted. Should we throw if realtime.subscribe does not work ?

### DIFF
--- a/packages/cozy-realtime/src/Socket.js
+++ b/packages/cozy-realtime/src/Socket.js
@@ -112,6 +112,18 @@ class Socket {
   }
 
   /**
+   * If Socket is not connected, connects.
+   * Does nothing if Socket is already connected
+   *
+   * @return {Promise} - Resolves when connected
+   */
+  async ensureConnected() {
+    if (!this.isOpen()) {
+      await this.connect()
+    }
+  }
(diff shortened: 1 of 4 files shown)


packages/cozy-realtime/src/Socket.js
```diff
@@ -111,6 +112,18 @@ class Socket {
...
   }
 
+  /**
+   * If Socket is not connected, connects.
+   * Does nothing if Socket is already connected
+   *
+   * @return {Promise} - Resolves when connected
+   */
+  async ensureConnected() {
+    if (!this.isOpen()) {
```
edas: If the socket is establishing the connection, this will start a new connection again. Maybe we should have a special case for `isConnecting()`

nono:
Ping, what is the status of this PR?

edas:
I am closing this as the PR itself is deprecated since #967 , but  the behaviour  has not changed. An error in subscription does not throw and only logs. I am not sure we could do better anyways, as the response from the server is async and  not linked the subscription request (at best, we could wait for some time to see if an error raises)

PR #446, opened 2019-05-16
feat: Add playgrounds

ptbrowne:
We begin to have a lot of small libraries and sometimes it is difficult to see the interactions
between them and how one change to one library might break another one. It is often that
we see that something break when the libraries have already been published. For the app
developer, it is hard to debug. I think it would be best if the bug was catched sooner, before
the library is published.

To achieve this goal, I think it would be useful to have minimal live examples (in a web page) tying together our different
libraries to check for imcompatibility problems without going to each and every
app (= seeing imcompatiblity problems before an app upgrades a library).
My end goal would be to have a kind of styleguidist for our libraries.

The next step would be to have e2d test cases on this.

It can also help when trying to reproduce bugs in a minimal test case to
understand clearly where does the bug come from. With playgrounds, it is easy to
copy paste and example, launch it, modify it to achieve a minimal reproducing
case.

For example here, I tried to reproduce a [reported bug](https://github.com/cozy/cozy-client/issues/439) case of a CozyProvider passed with an
existing store that does not updates the children `<Query />`s. I could not do
that so I can be fairly certain that the bug lies somewhere else (either in the
library or the app, but not only for the particular reason of passing a store
the CozyProvider).

Inspiration : [examples in the doc of react virtualized](http://bvaughn.github.io/react-virtualized/#/components/List)


ptbrowne:
> I'm just wondering if we should not create a dedicated repo for this "playground".

This folder is meant to showcase examples also. It is meant to be like a styleguidist for our examples of how to use our libraries.

This is not meant solely to quickly share reproducible bugs but it can be used that way.

ptbrowne:
I think it would be easier to have playgrounds in the libs repo not to have to manage links and do that via lerna.

edas:
Aren't we able to do a simple automated test for these codes instead of manual playground ?

After two days, I not sure about why we have to exit from the automated jest run for that.

Crash--:
En fait le playground est vachement utile quand tu as des lib qui interagissent entre elles ou quand tu veux tester des choses un peu plus haut niveau sans avoir tout le poids d'une application cozy réelle. 

Pour le debug ça peut te permettre d'isoler vraiment le soucis. 

Sur ce playground, on pourrait facilement ajouter la partie Cordova pour tester de manière plus simple un composant à l'intérieur d'une app cordova par exemple. 

J'étais pas convaincu au début, mais plus j'y pense, plus ça peut être utile. Ça coute pas grand chose d'essayer ?

PR #451, opened 2019-05-20
Account deconnection

CPatchane:
Export a button to handle account removal (seen as log out for the user)

Waiting to be correctly integrated into cozy-home

### DIFF
--- a/packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
+++ b/packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
@@ -1,0 +1,79 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import omit from 'lodash/omit'
+
+import { Button } from 'cozy-ui/react/Button'
+import { translate } from 'cozy-ui/react/I18n'
+import { withMutations } from 'cozy-client'
+
+import accountsMutations from '../connections/accounts'
+
+export class DeleteAccountButton extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      deleting: false
+    }
+    this.handleDelete = this.handleDelete.bind(this)
+  }
+
+  async handleDelete() {
+    this.setState({ deleting: true })
+    const { account, deleteAccount, onError, onSuccess } = this.props
+    try {
+      await deleteAccount(account)
+      if (typeof onSuccess === 'function') onSuccess(account)
+    } catch (e) {
+      if (typeof onError === 'function') {
+        onError(e)
+      } else {
+        // eslint-disable-next-line no-console
+        console.error(e)
+      }
+    } finally {
+      this.setState({ deleting: false })
+    }
+  }
+
+  render() {
+    const { t } = this.props
+    const { deleting } = this.state
+    return (
+      <Button
+        theme="danger-outline"
+        disabled={deleting}
+        busy={deleting}
+        onClick={this.handleDelete}
+        label={t('accountForm.disconnect.button')}
+        {...omit(this.props, Object.keys(DeleteAccountButton.propTypes))}
+      />
+    )
+  }
+}
+
+DeleteAccountButton.propTypes = {
+  /**
+   * io.cozy.accounts document
+   */
+  account: PropTypes.object.isRequired,
+  /**
+   * Provided by accountsMutations
+   */
+  deleteAccount: PropTypes.func.isRequired,
+  /**
+   * callback to handle account deleting error
+   */
+  onError: PropTypes.func,
+  /**
+   * callback to handle account deleting success
+   */
+  onSuccess: PropTypes.func,
+  /**
+   * Provided by translate HOC
+   */
+  t: PropTypes.func.isRequired
+}
+
+export default translate()(
+  withMutations(accountsMutations)(DeleteAccountButton)
+)
--- a/packages/cozy-harvest-lib/src/connections/accounts.js
+++ b/packages/cozy-harvest-lib/src/connections/accounts.js
@@ -145,6 +145,24 @@ const watchKonnectorAccount = (client, account, options) => {
 }
 
 /**
+ * Deletes an account.
+ * @param  {Object}  client  CozyClient
+ * @param  {Object}  account io.cozy.accounts document
+ */
+const deleteAccount = async (client, account) => {
+  try {
+    await client.destroy(account)
+  } catch (error) {
+    if (error.status === 409) {
+      const syncedAccount = await findAccount(client, account._id)
+      await client.destroy(syncedAccount)
+    } else {
+      throw error
+    }
+  }
+}
+
+/**
  * Get accounts mutations
  * @param  {Object} client CozyClient
  * @return {Object}        Object containing accounts mutations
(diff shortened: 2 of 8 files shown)


packages/cozy-harvest-lib/src/connections/accounts.js
```diff
@@ -144,6 +144,24 @@ const watchKonnectorAccount = (client, account, options) => {
...
+ * @param  {Object}  client  CozyClient
+ * @param  {Object}  account io.cozy.accounts document
+ */
+const deleteAccount = async (client, account) => {
+  try {
+    await client.destroy(account)
+  } catch (error) {
+    if (error.status === 409) {
+      const syncedAccount = await findAccount(client, account._id)
+      await client.destroy(syncedAccount)
```
Crash--: We should add a try / catch block here?

packages/cozy-harvest-lib/src/components/AccountLogoutButton.jsx
```diff
@@ -0,0 +1,78 @@
...
+import PropTypes from 'prop-types'
+import omit from 'lodash/omit'
+
+import { Button } from 'cozy-ui/react/Button'
+import { translate } from 'cozy-ui/react/I18n'
+import { withMutations } from 'cozy-client'
+
+import accountsMutations from '../connections/accounts'
+
+export class AccountLogoutButton extends Component {
```
gregorylegarec: As we talk before: I realy don't like the component name here.

I suggest `DeleteAccountButton` instead.

packages/cozy-harvest-lib/src/components/AccountLogoutButton.jsx
```diff
@@ -0,0 +1,78 @@
...
+  constructor(props) {
+    super(props)
+    this.state = {
+      loggingOut: false
```
gregorylegarec: Instead of a flag I suggest to use a status between: `IDLE`, `PENDING` (or `DELETING`), `ERRORED`.

packages/cozy-harvest-lib/src/components/AccountLogoutButton.jsx
```diff
@@ -0,0 +1,78 @@
...
+    }
+    this.logout = this.logout.bind(this)
+  }
+
+  async logout() {
```
gregorylegarec: s/logout/delete/

packages/cozy-harvest-lib/src/components/AccountLogoutButton.jsx
```diff
@@ -0,0 +1,78 @@
...
+    const { t } = this.props
+    const { loggingOut } = this.state
+    return (
+      <Button
+        theme="danger-outline"
+        disabled={loggingOut}
+        busy={loggingOut}
+        onClick={this.logout}
+        label={t('accountForm.disconnect.button')}
+        {...omit(this.props, Object.keys(AccountLogoutButton.propTypes))}
```
gregorylegarec: 👍

packages/cozy-harvest-lib/test/components/AccountLogoutButton.spec.js
```diff
@@ -0,0 +1,83 @@
...
+  it('should run deleteAccount on click', () => {
+    const component = shallow(
+      <AccountLogoutButton account={fixtures.account} {...props} />
+    )
+    component.simulate('click')
+    expect(props.deleteAccount).toHaveBeenCalledWith(fixtures.account)
+    // the rest is asserted in the onSuccess
+  })
+
+  it('should change state on click and handle onSuccess', done => {
```
gregorylegarec: I don't think we should test state here. It's internal implementation and we should not care about that.

Instead, I suggest to test the that the button is rendered with a disabled and busy props.

packages/cozy-harvest-lib/src/connections/accounts.js
CPatchane: It will be [caught by the component itself](https://github.com/cozy/cozy-libs/pull/451/files#diff-44a2d2347aa83b5d53cc4ea6e2981d5eR26) to run `onError` props, it should be enough or I missed something?

Crash--: Should be enough. Thanks

packages/cozy-harvest-lib/src/components/AccountLogoutButton.jsx
CPatchane: I disagree, we don't have different state to handle here, I am not comfortable to use many strings just for this simple case

packages/cozy-harvest-lib/src/connections/accounts.js
ptbrowne: This behavior could be in cozy-client IMHO.

```
client.destroy(account, { force: true })
```

packages/cozy-harvest-lib/src/connections/accounts.js
```diff
@@ -144,6 +144,24 @@ const watchKonnectorAccount = (client, account, options) => {
   return accountWatcher
 }
 
+/**
+ * Delete an account
```
ptbrowne:
```suggestion
 * Deletes an account.
 * 
 * Forces deletion even if the local account does not have the latest revision.
```

packages/cozy-harvest-lib/test/components/DeleteAccountButton.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    expect(component).toMatchSnapshot()
+  })
+
+  it('should run deleteAccount on click', () => {
+    const component = shallow(
+      <DeleteAccountButton account={fixtures.account} {...props} />
+    )
+    component.simulate('click')
+    expect(props.deleteAccount).toHaveBeenCalledWith(fixtures.account)
+    // the rest is asserted in the onSuccess
```
ptbrowne: I think you can remove this comment. It's difficult to know what "rest" is here.

packages/cozy-harvest-lib/test/components/DeleteAccountButton.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    const component = shallow(
+      <DeleteAccountButton
+        account={fixtures.account}
+        {...props}
+        deleteAccount={deleteAccount}
+        onError={onError}
+      />
+    )
+    component.simulate('click')
+    // the rest is asserted in the onError
```
ptbrowne: `expect(component.state().loggingOut).toBe(false)` will test the finally clause

packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
```diff
@@ -0,0 +1,79 @@
...
+import { translate } from 'cozy-ui/react/I18n'
+import { withMutations } from 'cozy-client'
+
+import accountsMutations from '../connections/accounts'
+
+export class DeleteAccountButton extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      loggingOut: false
```
gregorylegarec: We should at least change name. I suggest `deleting`

packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
```diff
@@ -0,0 +1,79 @@
...
+      />
+    )
+  }
+}
+
+DeleteAccountButton.propTypes = {
+  /**
+   * io.cozy.accounts document
+   */
+  account: PropTypes.object.isRequired,
```
ptbrowne: Tip: You can use the shorter

```
  /** @type {[type]} [description] */
```

I get these when I type `/**` `[space]` in my editor.

CPatchane: Thanks for the tip, for now I'll keep like this to be consistent with other components

PR #452, opened 2019-05-22
Fix unsubmittable 2FA Form on react apps

gregorylegarec:
It appears that we are not able to submit the 2FA Form in React apps.

Adding an onClick handler on submit button fix both submit by clik and with keyboard.

An issue will be made to investigate this behaviour further.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TwoFAForm/index.jsx
+++ b/packages/cozy-harvest-lib/src/components/TwoFAForm/index.jsx
@@ -90,6 +90,7 @@ export class TwoFAForm extends PureComponent {
               busy={submitting}
               disabled={submitting || !twoFACode}
               extension="full"
+              onClick={this.handleSubmit}
             />
           </form>
         </ModalDescription>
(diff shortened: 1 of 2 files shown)


CPatchane:
I noticed that when the form was not working, submitting with the Enter key was not working either, it that PR fix this issue?

packages/cozy-harvest-lib/src/components/TwoFAForm/index.jsx
```diff
@@ -89,6 +90,7 @@ export class TwoFAForm extends PureComponent {
               busy={submitting}
               disabled={submitting || !twoFACode}
               extension="full"
+              onClick={this.handleSubmit}
```
Crash--: Did you tried to add a `type=submit`?

CPatchane: The CozyUI button is `type=submit` by default

ptbrowne: This is a bit surprising !

gregorylegarec:
@CPatchane Yes, as explained in the PR description.

packages/cozy-harvest-lib/src/components/TwoFAForm/index.jsx
drazik: Actually `submit` is the default type for raw `<button>`s. See https://codesandbox.io/s/frosty-leaf-dkfb9

PR #453, opened 2019-05-23
fix: One implementation of flags

ptbrowne:
#### Background

Connect was not attached to flags in Jest because it was part
of the browser API of flags. This caused failure of tests that
were importing components that were wrapped in connect.

The separation between browser and node APIs was done because of the
localStorage API not being present in node.

#### Solution

To simplify, this PRs merges back the two implementations, keeping
the browser implementation, meaning the flags in node will use the
same implementation as the browser (meaning that flags are observable
now in Node) which simplifies the maintenance.

The original problem with localStorage is solved by
introducing in FlagStore the notion of `longtermStore`. This
`longtermStore` is used when setting/removing
a key. This long-term store is optional.

The localStorage functionality is a form of long term store and is attached to FlagStore if localStorage is available.


packages/cozy-flags/src/connect.jsx
```diff
@@ -0,0 +1,30 @@
...
+ * Connects a component to the flags. The wrapped component
+ * will be refreshed when a flag changes.
+ */
+const connect = Component => {
+  class Wrapped extends React.Component {
+    constructor(props) {
+      super(props)
+      this.handleChange = this.handleChange.bind(this)
+    }
+    componentWillMount() {
```
Crash--: Can we use `didMount` instead? Since `willMount` is a legacy API

ptbrowne: Thanks, done in https://github.com/cozy/cozy-libs/pull/453/commits/30c2e7845ca49aac59cb0d7dcbc76831a06dd64a

packages/cozy-flags/package.json
```diff
@@ -29,7 +29,6 @@
     ]
   },
   "dependencies": {
-    "detect-node": "2.0.4",
```
Crash--: weird that there is no change in a yarn.lock file?

ptbrowne: Eagle eye !

Packages in the lerna repository do not have `yarn.lock`s. `detect-node` is still used by other packages (cozy-stack-client and another dep), this is why it has not disappeared.

Extract from the yarn.lock.

```

3788-cozy-stack-client@^6.25.0:
3789-  version "6.25.0"
3790-  resolved "https://registry.yarnpkg.com/cozy-stack-client/-/cozy-stack-client-6.25.0.tgz#4fc50087d900be4b0976ceca3606c77703cf5f52"
3791-  integrity sha512-SY7450Yg9n9HDcPsMaNgfTcId5rXOR7hpzyCo1OUZoLINOGlG9Crd10p01aiOLSbvw8z1piKJnGyxYSSrmBOXA==
3792-  dependencies:
3793:    detect-node "2.0.4"

14023-spdy-transport@^2.0.18:
14024-  version "2.1.1"
14025-  resolved "https://registry.yarnpkg.com/spdy-transport/-/spdy-transport-2.1.1.tgz#c54815d73858aadd06ce63001e7d25fa6441623b"
14026-  integrity sha512-q7D8c148escoB3Z7ySCASadkegMmUZW8Wb/Q1u0/XBgDKMO880rLQDj8Twiew/tYi7ghemKUi/whSYOwE17f5Q==
14027-  dependencies:
14028-    debug "^2.6.8"
14029:    detect-node "^2.0.3"
```

Crash--: Thank !

PR #455, opened 2019-05-23
feat(cozy-doctypes): Allow Document to use a CozyClient

cedricmessiant:
This is a first step to allow users to choose between `cozy-client` and `cozy-client-js` when they register a client for their models.

~~This is still at draft state since I would like to do the following before merge:~~
- ~~rename `  newCozyClient` :arrow_right: `cozyClient` and `cozyClient` :arrow_right: `cozyClientJS` in `testUtils.js`~~
- ~~add tests for methods that use "the `stackClient` trick" (`getAll`, `fetchChanges`, `fetchAll`, `updateAll`). To do this, I need to be able to mock the stack client that `CozyClient` uses, it may not be easy.~~
- ~~we could also be more explicit in the methods names (e.g `_ccjs_getIndex` instead of `_getIndex` ?)~~

### DIFF
--- a/packages/cozy-doctypes/src/Document.js
+++ b/packages/cozy-doctypes/src/Document.js
@@ -10,6 +10,7 @@ const groupBy = require('lodash/groupBy')
 const sortBy = require('lodash/sortBy')
 const get = require('lodash/get')
 const { parallelMap } = require('./utils')
+const CozyClient = require('cozy-client').default
 const log = require('cozy-logger').namespace('Document')
 const querystring = require('querystring')
 
@@ -63,6 +64,11 @@ const withoutUndefined = x => omitBy(x, isUndefined)
 const flagForDeletion = x => Object.assign({}, x, { _deleted: true })
 
 class Document {
+  /**
+   * Registers a client
+   *
+   * @param {Client} client - Cozy client from either cozy-client or cozy-client-js
+   */
   static registerClient(client) {
     if (!this.cozyClient) {
       this.cozyClient = client
@@ -75,7 +81,33 @@ class Document {
     }
   }
 
+  /**
+   * Unregisters a client
+   *
+   * This method is dangerous and is meant to be used only by tests.
+   */
+  static dangerousUnregisterClient() {
+    this.cozyClient = null
+  }
+
+  /**
+   * Returns true if Document uses a CozyClient (from cozy-client package)
+   *
+   * @returns {boolean} true if Document uses a CozyClient
+   **/
+  static usesCozyClient() {
+    return this.cozyClient instanceof CozyClient
+  }
+
   static getIndex(doctype, fields) {
+    if (this.usesCozyClient()) {
+      throw new Error('This method is not implemented yet with CozyClient')
+    }
+
+    return this.getIndexViaOldClient(doctype, fields)
+  }
+
+  static getIndexViaOldClient(doctype, fields) {
     const key = `${doctype}:${fields.slice().join(',')}`
     const index = indexes[key]
     if (!index) {
@@ -108,6 +140,14 @@ class Document {
   }
 
   static async createOrUpdate(attributes) {
+    if (this.usesCozyClient()) {
+      throw new Error('This method is not implemented yet with CozyClient')
+    }
+
+    return this.createOrUpdateViaOldClient(attributes)
+  }
+
+  static async createOrUpdateViaOldClient(attributes) {
     const selector = fromPairs(
       this.idAttributes.map(idAttribute => [
         idAttribute,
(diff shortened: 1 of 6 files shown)


packages/cozy-doctypes/src/Document.js
```diff
@@ -63,6 +64,11 @@ const withoutUndefined = x => omitBy(x, isUndefined)
 const flagForDeletion = x => Object.assign({}, x, { _deleted: true })
 
 class Document {
+  /**
+   * Registers a client
+   *
+   * @param {object} client - Cozy client from either cozy-client or cozy-client-js
```
ptbrowne:
```suggestion
   * @param {Client} client - Cozy client from either cozy-client or cozy-client-js
```

packages/cozy-doctypes/src/Document.js
```diff
@@ -75,7 +81,29 @@ class Document {
     }
   }
 
+  static unregisterClient() {
+    this.cozyClient = null
+  }
```
ptbrowne: Why is there an unregister ?

Since the client is registered on all the class, it can be dangerous since if you launch parallel queries on different cozies (like ACH batch), you can think that it is safe, to register a client multiple times, when it is not.

packages/cozy-doctypes/src/Document.js
```diff
@@ -75,7 +81,29 @@ class Document {
...
+
   static getIndex(doctype, fields) {
+    if (this.usesCozyClient()) {
+      throw new Error('This method is not implemented yet with CozyClient')
+    }
+
+    return this._getIndex(doctype, fields)
+  }
+
+  // getIndex method that uses cozy-client-js
```
ptbrowne: getIndexViaOldClient lets you remove the comment.

packages/cozy-doctypes/src/Document.js
```diff
@@ -335,6 +402,15 @@ class Document {
...
   static async queryAll(selector, index) {
+    if (this.usesCozyClient()) {
+      throw new Error('This method is not implemented yet with CozyClient')
+    }
+
+    return this._queryAll(selector, index)
+  }
+
+  // queryAll method that uses cozy-client-js
+  static async _queryAll(selector, index) {
```
ptbrowne: Renaming into queryAllViaClientJS  lets you remove the comment ;)

packages/cozy-doctypes/src/Document.js
```diff
@@ -108,6 +136,15 @@ class Document {
...
   static async createOrUpdate(attributes) {
+    if (this.usesCozyClient()) {
+      throw new Error('This method is not implemented yet with CozyClient')
+    }
+
+    return this._createOrUpdate(attributes)
+  }
+
+  // createOrUpdate method that uses cozy-client-js
+  static async _createOrUpdate(attributes) {
```
ptbrowne: createOrUpdateViaOldClient lets you remove the comment

packages/cozy-doctypes/src/Document.js
cedricmessiant: yeah, it's for the tests (because doing Document.`registerClient(null)` was not working (anymore?) Will change it to `dangerousUnregisterClient` and add a comment

packages/cozy-doctypes/src/Document.js
cedricmessiant: ok for the renaming. As I said in the description, I wanted to prefix those methods with `_ccjs_` but I did not like it. `*ViaOldClient` is more clear

packages/cozy-doctypes/src/Document.js
edas: I am not fond of having a such method, even because of the tests. AFAIR, Jest has ways to reset modules that maintain a state, so you would be able to start over without this method

edas: (Speaking about that, I see more and more statefull objects and modules, like this class with everything static but still an internal state. This is generally an anti-pattern)

edas:
On utilise le même modèle mais on considère que les méthodes ne doivent être accessibles que via un client et pas l'autre. C'est une vraie volonté de les abandonner et de forcer cet abandon ou c'est juste par manque de temps pour les implémenter avec cozy-client ?

cedricmessiant:
@edas c'est une première étape qui permet de permettre l'utilisation avec `cozy-client` sans pour autant passer trop de temps aujourd'hui à tout implémenter. L'idée c'est d'implémenter au fur et à mesure les méthodes lorsqu'on en a besoin et d'abandonner le support `cozy-client-js` à moyen/long terme

packages/cozy-doctypes/src/Document.js
cedricmessiant: Sorry @edas I saw your comment after the merge. I did a PR to remove the method: https://github.com/cozy/cozy-libs/pull/459

PR #467, opened 2019-05-29
Harvest: quick refactor for a better translation mechanism

gregorylegarec:
This PR uses `withLocales` from Cozy-UI.

Default exports from components are now directly translated.

It will allow us to export HOC wrapping component with translated component.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/hoc/withLocales.js
+++ b/packages/cozy-harvest-lib/src/components/hoc/withLocales.js
@@ -1,0 +1,5 @@
+import withLocales from 'cozy-ui/transpiled/react/I18n/withLocales'
+
+const dictRequire = lang => require(`../../locales/${lang}.json`)
+
+export default withLocales(dictRequire)
--- a/packages/cozy-harvest-lib/src/index.js
+++ b/packages/cozy-harvest-lib/src/index.js
@@ -1,39 +1,7 @@
-import React, { Component } from 'react'
-import PropTypes from 'prop-types'
-import I18n from 'cozy-ui/react/I18n'
-
-import TranslatedDeleteAccountButton from './components/DeleteAccountButton'
-import TranslatedTriggerManager from './components/TriggerManager'
-import TranslatedTriggerLauncher from './components/TriggerLauncher'
+export {
+  default as DeleteAccountButton
+} from './components/DeleteAccountButton'
+export { default as TriggerManager } from './components/TriggerManager'
+export { default as TriggerLauncher } from './components/TriggerLauncher'
 
 export { KonnectorJobError } from './helpers/konnectors'
-
-const dictRequire = lang => require(`./locales/${lang}.json`)
-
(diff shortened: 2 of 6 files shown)


packages/cozy-harvest-lib/src/index.js
```diff
@@ -1,39 +1,10 @@
...
-
-import TranslatedDeleteAccountButton from './components/DeleteAccountButton'
-import TranslatedTriggerManager from './components/TriggerManager'
-import TranslatedTriggerLauncher from './components/TriggerLauncher'
+export {
+  default as DeleteAccountButton
+} from './components/DeleteAccountButton'
+export {
+  default as withKonnectorModal
+} from './components/hoc/withKonnectorModal'
```
CPatchane: `export withKonnectorModal from './components/hoc/withKonnectorModal'` doesn't work here?

gregorylegarec:
@ptbrowne @Crash-- @CPatchane The original implementation was causing the fr locales to not be loaded in :
* Dev environment without having pull them from Transifex
* Travis tests

As I don't wanted to log a warning every time a component would not have been able to load the fr language, and I did not want neither to catch an error silently, I came with a new implementation in the last commit, a little more complex. As it retrieve the lang from context, it require only the locales we need. What do yout think ? Do you see any simpler implementation ?

packages/cozy-harvest-lib/src/components/hoc/withLocales.js
```diff
@@ -0,0 +1,8 @@
+import withLocales from 'cozy-ui/transpiled/react/I18n/withLocales'
+
+const locales = {
+  en: require(`../../locales/en.json`),
+  fr: require(`../../locales/fr.json`)
+}
+
+export default withLocales(locales)
```
Crash--: do you still need this file ? When you do `import withLocales from './hoc/withLocales'` is it the jsx or the js ?

gregorylegarec: No we don't need it anymore, it was supposed to be removed in the last commit, please ignore it :)

ptbrowne:
I think you should use a webpack context.

```
const localeContextAsObject = ctx => {
  const res = {}
  for (const filename of ctx.keys()) {
    const key = filename.slice(2, -5)
    if (key.length > 0) {
        res[key] = ctx(filename)
    }
  }
  return res
}

const locales = localeContextAsObject(require.context('./locales', false, '\.json$'))

export withLocales(locales)

gregorylegarec:
Now waiting for https://github.com/cozy/cozy-ui/pull/1019

packages/cozy-harvest-lib/src/index.js
gregorylegarec: When using this syntax, I got `Support for the experimental syntax 'exportDefaultFrom' isn't currently enabled` in Cozy-Bar :sob:

PR #461, opened 2019-05-28
Harvest: Use harvest modal from harvest-lib

CPatchane:
This PR allows having install and edition when configuring konnectors

Wait for #440 to be merged before this one

PR #440, opened 2019-05-13
Harvest: Make the TriggerManager working with konnectorJob

CPatchane:
Since the TriggerLauncher we added a new konnectorJob object to handle all the process related to the trigger execution.
Now the konnectorJob can also handle account and trigger creation and the TriggerManager can now use it to submit the account form and run the trigger just after.

⚠️ Need to update and add more tests before merging it

PR #468, opened 2019-05-29
Harvest: Add a konnector modal to edit existing accounts

gregorylegarec:
For now this modal just fetch the first account related to given konnector and then rely on TriggerManager, like it's done in Cozy-Home.

~~Blocked by https://github.com/cozy/cozy-libs/pull/467, both `KonnectorModal` and `withKonnectorModal` should be exported in index.js.~~

Replace both https://github.com/cozy/cozy-libs/pull/440 and https://github.com/cozy/cozy-libs/pull/461

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
@@ -91,14 +91,27 @@ export class KonnectorModal extends PureComponent {
             <SubTitle className="u-mb-1 u-ta-center">
               {t('modal.konnector.title', { name: konnector.name })}
             </SubTitle>
-            <TriggerManager
-              account={account}
-              konnector={konnector}
-              trigger={trigger}
-              running={running}
-              onLoginSuccess={onLoginSuccess}
-              onSuccess={onSuccess}
-            />
+            {error ? (
+              <Infos
+                actionButton={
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -0,0 +1,110 @@
...
+  state = {
+    account: null,
+    fetching: true,
+    trigger: null
+  }
+
+  constructor(props) {
+    super(props)
+    this.fetchIcon = this.fetchIcon.bind(this)
+    this.fetchAccount()
```
Crash--: We should call `fetchAccount()` in the `componentDidMount`.

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  constructor(props) {
+    super(props)
+    this.fetchIcon = this.fetchIcon.bind(this)
+  }
+
+  componentDidMount() {
+    this.fetchAccount()
+  }
```
Crash--: We should try / catch ?

gregorylegarec: Done.

PR #473, opened 2019-05-31
Prevent undefined success callback

CPatchane:
Throw error during the process if not provided, and this prop is optional


Crash--:
if this param is optional we should say it in the jsdoc

ptbrowne:
@Crash--  onLoginSuccess and onSuccess are not required in the proptypes. I am already unconfortable that propTypes are repeated in the JSDoc, do you think it is realllynecessary to add that in the JSdoc ?

PR #477, opened 2019-06-03
[harvest]: Correctly set fetched trigger in state in KonnectorModal

CPatchane:
The trigger was not stored into the state to be used correctly by the component

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
@@ -61,7 +61,7 @@ export class KonnectorModal extends PureComponent {
 
     try {
       const account = await findAccount(triggers.getAccountId(trigger))
-      this.setState({ account })
+      this.setState({ account, trigger })
     } catch (error) {
       this.setState({
         error
--- a/packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
+++ b/packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
@@ -1,0 +1,35 @@
+/* eslint-env jest */
+import React from 'react'
+import { shallow } from 'enzyme'
+
+import { KonnectorModal } from 'components/KonnectorModal'
+import TriggerManager from 'components/TriggerManager'
+
+const t = key => key
+
+const findAccountMock = jest.fn().mockImplementation(() => ({
+  id: 123,
+  doctype: 'io.cozy.accounts'
+}))
+
+describe('KonnectorModal', () => {
+  it('should pass trigger if konnector has triggers', async () => {
+    const mockKonnector = {
+      slug: 'mock',
+      name: 'Mock',
+      triggers: {
+        data: [{ id: 784, doctype: 'io.cozy.triggers' }]
+      }
+    }
+    const props = {
+      findAccount: findAccountMock,
+      dismissAction: jest.fn(),
+      onSuccess: jest.fn(),
+      konnector: mockKonnector,
+      t
+    }
+    const component = await shallow(<KonnectorModal {...props} />)
+    const manager = component.find(TriggerManager)
+    expect(manager.props().trigger).toBeDefined()
+  })
+})


packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-env jest */
+import React from 'react'
+import { shallow } from 'enzyme'
+
+import { KonnectorModal } from 'components/KonnectorModal'
+import TriggerManager from 'components/TriggerManager'
+
+const t = jest.fn().mockImplementation(key => key)
```
ptbrowne:
```suggestion
const t = key => key
```

packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
```diff
@@ -0,0 +1,35 @@
...
+
+const findAccountMock = jest.fn().mockImplementation(() => ({
+  id: 123,
+  doctype: 'io.cozy.accounts'
+}))
```
ptbrowne: No need for a mock function here either.

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -61,7 +61,7 @@ export class KonnectorModal extends PureComponent {
 
     try {
       const account = await findAccount(triggers.getAccountId(trigger))
-      this.setState({ account })
+      this.setState({ account, trigger })
```
gregorylegarec: I think we should set the trigger in state whatever happen, to avoid being in the situation where no trigger is provided to TriggerManager and it behaves like for a trigger creation. This behaviour is not yet handled by this modal.

packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
```diff
@@ -0,0 +1,35 @@
...
+      }
+    }
+    const props = {
+      findAccount: findAccountMock,
+      dismissAction: jest.fn(),
+      onSuccess: jest.fn(),
+      konnector: mockKonnector,
+      t
+    }
+    const component = await shallow(<KonnectorModal {...props} />)
```
gregorylegarec: I never used `await` for `shallow`, what does it do ?

ptbrowne: I dont think it is correct.

CPatchane: It will wait for async `componentDidMount`, here we fetch the account after mounting. If I removed it, I'll get the loading status here

packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
CPatchane: We need to correctly mock the fetching account from the trigger provided

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
CPatchane: If the accounts fail, it will be in the errored status, which is [correctly handled here](https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx#L120)

packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
gregorylegarec: As findAccount is asynchronous, I suggest: 
```
const findAcocuntMock = jest.fn().mockResolvedValue({
  id: 123,
  doctype: 'io.cozy.accounts'
})
```

packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
ptbrowne: oh, I did not know it was possible to do that. Nice.

packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
ptbrowne: I dont undertand why we use jest.fn here ? 

```
const findAccountMock = () => Promise.resolve({ id: 123, doctype: 'io.cozy.accounts' })
```

It is not really a problem, but I see no point in using jest.fn when we do not need it.
```

File: packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js (hunk 4 above)
Comment (gregorylegarec) [reply packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js]: `componentDidMount` is not asynchronous here (only `fetchAccount` is).

First solution I see is:
```
componentDidMount(){
 this.fetchAccountPromise = this.fetchAccount()
}
```

And in test:
```
  const wrapper = shallow(<KonnectorModal {...props} />)
  await wrapper.instance().fetchAccountPromise
```

packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
CPatchane: Since we don't await fetchAccount in the componentDidMount, it doesn't work

packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
ptbrowne: Seems good enough

PR #483, opened 2019-06-06
Don't close 2FA modal on clickoutside event

CPatchane:
By disabling it the user can't close the modal by inadvertence and being blocked outside of the 2FA process


edas:
It seems like a fragile hack that may break at any moment. Nothing in cozy-ui does document this behaviour.

It would probably be better if we could attach an additional props which will be forwarded to Modal stating that you do not want to dismiss the modal when clicking outside, then update cozy-UI accordingly.

CPatchane:
I agree, @edas. This add is not needed anymore, I think it was the cause of an issue about modal closing which is not, so let's close this PR. If we need this behaviour we will come back with something better handled by Cozy-ui itself

PR #495, opened 2019-06-14
fix(harvest): Rely on messageEvent.source for OAuth message 📝

gregorylegarec:
When receiving OAuth message from popup, we need to ensure that the popup is the source of the message. Previously it was done by checking the origin of the popup location. But it cannot work in https, as the opener window cannot access to popup's `window.location`.

This PR relies instead on `messageEvent.source` to ensure that the message is coming from the OAuth popup.


gregorylegarec:
TIL too, very useful actually, especially for our Popup !

PR #498, opened 2019-06-14
fix(harvest): Call Delete button callbacks after setState

CPatchane:
We get some warnings if we use onSuccess/onError to unmount the component or a parent of containing this component

### DIFF
--- a/packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
+++ b/packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
@@ -17,6 +17,14 @@ export class DeleteAccountButton extends Component {
     this.handleDelete = this.handleDelete.bind(this)
   }
 
+  componentDidMount() {
+    this._mounted = true
+  }
+
+  componentWillUnmount() {
+    this._mounted = false
+  }
+
   async handleDelete() {
     this.setState({ deleting: true })
     const { account, deleteAccount, onError, onSuccess } = this.props
@@ -31,7 +39,7 @@ export class DeleteAccountButton extends Component {
         console.error(e)
       }
     } finally {
-      this.setState({ deleting: false })
+      if (this._mounted) this.setState({ deleting: false })
     }
   }
 
--- a/packages/cozy-harvest-lib/test/components/DeleteAccountButton.spec.js
+++ b/packages/cozy-harvest-lib/test/components/DeleteAccountButton.spec.js
@@ -44,18 +44,15 @@ describe('DeleteAccountButton', () => {
 
   it('should change state on click and handle onSuccess', done => {
     const component = shallow(
-      <DeleteAccountButton account={fixtures.account} {...props} />
+      <DeleteAccountButton
+        account={fixtures.account}
+        onSuccess={() => done()}
+        {...props}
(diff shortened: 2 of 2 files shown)


CPatchane:
[(𝐋) 🔵 🍊 [HOME] Migration de  la Home sur React](https://trello.com/c/CWnlkCYC/1442-%F0%9D%90%8B-%F0%9F%94%B5-%F0%9F%8D%8A-home-migration-de-la-home-sur-react)

packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
```diff
@@ -22,16 +22,18 @@ export class DeleteAccountButton extends Component {
...
-      }
-    } finally {
-      this.setState({ deleting: false })
+      this.setState({ deleting: false }, () => {
+        if (typeof onError === 'function') {
+          onError(e)
+        } else {
+          // eslint-disable-next-line no-console
+          console.error(e)
+        }
```
Crash--: In this case, why not putting `this.setState({deleting: false})` before the `try` block ?

packages/cozy-harvest-lib/test/components/DeleteAccountButton.spec.js
```diff
@@ -42,22 +42,6 @@ describe('DeleteAccountButton', () => {
...
-
-    // async assertions
-    const onSuccessAssert = () => {
-      expect(component.state().deleting).toBe(true)
-      done()
-    }
-    component.setProps({ onSuccess: onSuccessAssert })
-
-    component.simulate('click')
-  })
```
Crash--: Why are you removing this test?

packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
CPatchane: Because we have to wait for the `deleteAccount` be finished before changing this state

Crash--: I would prefer to keep the finally block and having a cancelable promise during the unmount to not set the state. Or having a `this._isMounted` in the finally block. WDYT ?

packages/cozy-harvest-lib/test/components/DeleteAccountButton.spec.js
CPatchane: It's back 👍

packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
CPatchane: FMC, what is the issue to rely on something provided by the `setState` API here?

Crash--: Nothing  really wrong, but as we want to setState({deleting: false}) in each case, we should deal with it only once. 

Also, do we really considered that we have managed the `deletion` before calling the success or the `onError` callback? Not sure since they are part of the worklow. 

I would avoid to trigger a render before calling the callback. 

This is why I prefer to put the setState in the finally block and test if the component is still mounted.

PR #503, opened 2019-06-17
[cozy-doctypes] feat: Basic admin procedure doctype

y-lohse:
Adds a basic doctype for administrative procedures. Keeping it as a draft until we add some logic to it.


y-lohse:
Taken care of elsewhere.

PR #506, opened 2019-06-18
fix(procedures): Translate form title and button, add subtitle

cedricmessiant:


y-lohse:
Hmm, I did similar work on the `Overview` page, but I chose to use the `Subtitle` and `Title` components instead of the utility classes. I think it's best if we write as little css as possible, and that includes utility classes. WDYT?

packages/cozy-procedures/src/components/PersonalDataForm.jsx
```diff
@@ -25,10 +26,11 @@ const widgets = {
 
 class PersonalDataForm extends React.Component {
   render() {
-    const { formData, updateFormData, router } = this.props
+    const { formData, updateFormData, router, t } = this.props
     return (
       <div className="u-ml-2">
-        <h2 className="u-title-h2">Personal data</h2>
+        <h2 className="u-title-h2">{t('personalData.title')}</h2>
+        <h3 className="u-title-h3">{t('personalData.subtitle')}</h3>
```
ptbrowne: Why do we need to do that ? Aren't there typographic components in cozy-ui ?

cedricmessiant:
> Hmm, I did similar work on the `Overview` page, but I chose to use the `Subtitle` and `Title` components instead of the utility classes. I think it's best if we write as little css as possible, and that includes utility classes. WDYT?

yeah, of course !

packages/cozy-procedures/src/components/PersonalDataForm.jsx
cedricmessiant: fixed it after @y-lohse 's comment, thanks

PR #513, opened 2019-06-19
Populate personal data with myself

cedricmessiant:
Fetch "myself" contact to populate the `personalData` slice in the redux store. It is then visible in the form.


packages/cozy-doctypes/src/contacts/Contact.js
```diff
@@ -116,6 +125,44 @@ class Contact extends Document {
   static getDisplayName(contact) {
     return Contact.getFullname(contact) || Contact.getPrimaryEmail(contact)
   }
+
+  /**
+   * Returns personal data for the contact
+   *
+   * @param {Object} contact - A contact
```
Crash--: io.cozy.contacts ?

packages/cozy-doctypes/src/contacts/Contact.js
```diff
@@ -80,6 +80,15 @@ class Contact extends Document {
       ? getPrimaryOrFirst('cozy')(contact).url
       : contact.url
   }
+  /**
+   * Returns the contact's main phone number
+   *
+   * @param {Object} contact - A contact
```
Crash--: io.cozy.contacts?

packages/cozy-procedures/src/Procedure.jsx
```diff
@@ -5,24 +5,30 @@ import get from 'lodash/get'
...
 
   render() {
     return this.props.children
   }
 }
 
-const mapDispatchToProps = dispatch => ({
-  initPersonalData: template => dispatch(initPersonalData(template))
-})
+const mapDispatchToProps = {
```
Crash--: can we have a separate file for the container ?

packages/cozy-procedures/src/redux/personalDataSlice.js
```diff
@@ -22,6 +52,37 @@ const selectors = {
...
+
+export const {
+  init,
+  update,
+  fetchMyselfLoading,
+  fetchMyselfSuccess,
+  fetchMyselfError
+} = actions
+
+export function fetchMyself(client) {
```
Crash--: I think we can add the client to the configuration. 

In drive with have something like :   const middlewares = [thunkMiddleware.withExtraArgument({ client, t })] to inject client in the thunk

packages/cozy-procedures/src/components/Overview.jsx
```diff
@@ -27,14 +27,20 @@ const Overview = ({ personalData, location }) => {
 }
 
 Overview.propTypes = {
-  personalData: PropTypes.object,
+  personalData: PropTypes.shape({
+    data: PropTypes.object,
```
Crash--: 👍

packages/cozy-doctypes/src/contacts/Contact.js
```diff
@@ -116,6 +125,44 @@ class Contact extends Document {
...
   }
+
+  /**
+   * Returns personal data for the contact
+   *
+   * @param {Object} contact - A contact
+   * @param {Array} fields - The list of fields to retrieve
+   * @return {Object} - the personal data
+   **/
+  static getPersonalData(contact, fields) {
```
y-lohse: I thought this would have been in a Procedure model rather than contact, since the personal data names are really coupled to procedures?

packages/cozy-procedures/src/Procedure.jsx
cedricmessiant: It's done but I also had to put `ProcedureWithStore` in another file (named it `App`, not sure it's the best name)

packages/cozy-procedures/src/containers/PersonalDataForm.jsx
```diff
@@ -4,7 +4,7 @@ import PersonalDataForm from '../components/PersonalDataForm'
 import { getSlice, update as updateFormData } from '../redux/personalDataSlice'
 
 const mapStateToProps = state => ({
-  formData: getSlice(state)
+  formData: getSlice(state).data
```
y-lohse: I think it would be better to have a selector that specifically gets the `data` of the slice.

packages/cozy-procedures/src/redux/personalDataSlice.js
```diff
@@ -1,18 +1,48 @@
...
+  },
   slice: 'personalData',
   reducers: {
     init: (state, action) => {
-      return Object.keys(action.payload).reduce((acc, fieldId) => {
-        acc[fieldId] = undefined
-        return acc
-      }, {})
+      return {
+        data: Object.keys(action.payload).reduce((acc, fieldId) => {
```
y-lohse: We could just change the value of `state.data` here since there is now an initial state :)

cedricmessiant: hum, imagine the component is unmounted but `error` is not empty. It's probably safer to ensure that if we mount the component again, `error` is empty and `loading` is false, right?

packages/cozy-doctypes/src/contacts/Contact.js
cedricmessiant: Good point! And I think we actually said that when we talked about it a few days ago. @Crash-- do you agree?

Crash--: ok 👍

packages/cozy-procedures/src/redux/personalDataSlice.js
y-lohse: Yeah actually that's a good point 👍

cedricmessiant:
I think it's ready except missing tests for reducers

PR #516, opened 2019-06-20
(Procedures) Document UI

y-lohse:
The documents view is now using the template to generate the fields. Nothing is connected to the redux store yet.

### DIFF
--- a/packages/cozy-procedures/src/components/Documents.jsx
+++ b/packages/cozy-procedures/src/components/Documents.jsx
@@ -19,18 +19,18 @@ class Documents extends React.Component {
       <div>
         <Topbar title={t('documents.title')} />
         <Title>{t('documents.subtitle')}</Title>
+        <div>
+          <Label>Demo, to remove</Label>
+          <EmptyDocumentHolder />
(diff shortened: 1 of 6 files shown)


packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -6,20 +6,29 @@ import { Title, translate, Label, Button } from 'cozy-ui/transpiled/react/'
 import EmptyDocumentHolder from './documents/EmptyDocumentHolder'
 import LoadingDocumentHolder from './documents/LoadingDocumentHolder'
 import DocumentHolder from './documents/DocumentHolder'
+import creditApplicationTemplate from '../templates/creditApplicationTemplate'
 
 class Documents extends React.Component {
   render() {
     const { t, router } = this.props
+    const { documents } = creditApplicationTemplate
```
Crash--: Should we rename it to documentsTemplate ? 

I would prefer to have `documents` to refer to the "selected" documents / files ?

y-lohse: Yep.

PR #521, opened 2019-06-20
fix(cozy-app-publish): Handle snake case cli options

doubleface:
After 0.16.0 version, options like --build-dir did not work anymore


packages/cozy-app-publish/cozy-app-publish.js
```diff
@@ -96,6 +102,8 @@ function _getPublishMode() {
 }
 
 async function publishApp(cliOptions) {
+    console.log(cliOptions)
+        process.exit()
```
ptbrowne: Please remove this ^ :)

doubleface: oups

PR #525, opened 2019-06-21
Add status to form

cedricmessiant:
- add x/y status to the form `FormFillingStatus` component and container
- refactor selectors
- add tests for selectors
- use these selectors in `Overview`

`FormFillingStatus` is probably not a good name. I wanted to rename it to `FormStatus`, `FormFieldsStatus` or `FormCompletedStatus` but let's discuss it here before to do the changes.

### DIFF
--- a/packages/cozy-procedures/src/components/FormFillingStatus.jsx
+++ b/packages/cozy-procedures/src/components/FormFillingStatus.jsx
@@ -9,9 +9,12 @@ export const FormFillingStatus = ({ completed, total, t }) => {
   return (
     <div>
       <span className="u-weirdGreen">
-        {t('personalDataForm.completedStatus.main', { completed, total })}
+        {t('personalDataForm.completedStatus.main', {
+          smart_count: completed,
+          total
+        })}
       </span>{' '}
-      {t('personalDataForm.completedStatus.rest')}
+      {t('personalDataForm.completedStatus.rest', { smart_count: completed })}
     </div>
   )
 }
--- a/packages/cozy-procedures/src/components/FormFillingStatus.jsx
+++ b/packages/cozy-procedures/src/components/FormFillingStatus.jsx
@@ -1,0 +1,19 @@
+import React from 'react'
+import { translate } from 'cozy-ui/transpiled/react/I18n'
+
+export const FormFillingStatus = ({ completed, total, t }) => {
+  if (completed === '0') {
+    return null
+  }
+
+  return (
+    <div>
+      <span className="u-weirdGreen">
+        {t('personalDataForm.completedStatus.main', { completed, total })}
+      </span>{' '}
+      {t('personalDataForm.completedStatus.rest')}
+    </div>
+  )
+}
+
+export default translate()(FormFillingStatus)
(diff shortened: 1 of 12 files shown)


packages/cozy-procedures/src/locales/fr.json
```diff
@@ -21,6 +21,10 @@
   "personalDataForm": {
     "title": "Informations personnelles",
     "subtitle": "Complétez les informations suivantes pour finaliser votre dossier",
+    "completedStatus": {
+      "main": "%{completed} sur %{total} informations",
+      "rest": "ont déjà été complétées."
```
y-lohse: We should probably pluralize this, but not blocking;

packages/cozy-procedures/src/redux/personalDataSlice.js
```diff
@@ -93,6 +88,12 @@ export function fetchMyself(client) {
   }
 }
 
-export const { getData, getSlice } = selectors
+const getSlice = state => get(state, personalDataSlice.slice)
+const getData = state => get(state, [personalDataSlice.slice, 'data'], {})
+const getCompletedFields = state =>
+  Object.values(getData(state)).filter(Boolean).length
+const getTotalFields = state => Object.keys(getData(state)).length
```
Crash--: Y a une raison à avoir supprimé l'objet `selectors` ?

cedricmessiant: il ne permettait de réutiliser les sélecteurs pour écrire les autres sélecteurs. Ou alors il fallait faire quelque chose comme : 

```
const selectors = {}
selectors.getData = ...
selectors.getCompletedFields =   Object.values(selectors.getData(state)).filter(Boolean).length
```

packages/cozy-procedures/src/locales/fr.json
cedricmessiant: did it, thanks!

packages/cozy-procedures/src/redux/personalDataSlice.js
Crash--: 👍

packages/cozy-procedures/src/components/FormFillingStatus.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { translate } from 'cozy-ui/transpiled/react/I18n'
+
+export const FormFillingStatus = ({ completed, total, t }) => {
+  if (completed === '0') {
```
y-lohse: Why is `0` a string here? Wouldn't it make more sense to store it as a number?

packages/cozy-procedures/src/components/PersonalDataForm.jsx
```diff
@@ -32,6 +33,7 @@ class PersonalDataForm extends React.Component {
       <div>
         <Topbar title={t('personalDataForm.title')} />
         <SubTitle>{t('personalDataForm.subtitle')}</SubTitle>
+        <FormFillingStatus />
```
y-lohse: We're missing the props here.

cedricmessiant: the props are given by the container

y-lohse: Oh right. In that case the problem is somewhere else, because the counter appears even when no fields are completed at the moment :/

PR #530, opened 2019-06-25
(procedures) Amount and duration views

y-lohse:
Implements both Amount and Duration views. i'have had some weird behaviour from yarn while working on this one, so if someone else could check out the PR and makes sure everything runs fine for them, that would be great!


packages/cozy-procedures/package.json
```diff
@@ -10,8 +10,12 @@
     "lint": "cd ..; yarn lint packages/cozy-procedures"
   },
   "dependencies": {
+    "@material-ui/core": "^4.1.2",
+    "@material-ui/lab": "^4.0.0-alpha.17",
     "babel-preset-cozy-app": "1.5.1",
     "lodash": "4.17.11",
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0",
```
y-lohse: Yarn keeps adding those in the dependencies list automatically…

cedricmessiant:
@y-lohse I have this issue when I try to run it (it worked fine before I tried to update things)

```
🚨  /home/cedric/dev/cozy-libs/node_modules/unist-util-visit/node_modules/unist-util-visit-parents/index.js: ENOENT: no such file or directory, open '/home/cedric/dev/cozy-libs/node_modules/unist-util-visit/node_modules/unist-util-visit-parents/index.js'
Error: ENOENT: no such file or directory, open '/home/cedric/dev/cozy-libs/node_modules/unist-util-visit/node_modules/unist-util-visit-parents/index.js'
```

packages/cozy-procedures/src/components/Amount.jsx
```diff
@@ -1,22 +1,74 @@
...
+            extension="full"
+            onClick={this.confirm}
+          />
+        </div>
       </div>
     )
   }
+
+  confirm() {
+    this.props.updateAmount(parseInt(this.inputRef.current.value))
```
ptbrowne: Since you already use have the value via onChange, can't you just remove this line ?

y-lohse: It's trickier than that: on the first visit, a default value is set, but we don't want to save that default value — unless the explicitly clicks on the confirm button. It's the only reason why we need this other call to `updateAmount ` :(

packages/cozy-procedures/package.json
```diff
@@ -21,12 +25,14 @@
...
     "@babel/cli": "7.4.4",
-    "cozy-ui": "^20.18.0",
+    "@material-ui/core": "3.9.3",
+    "cozy-ui": "^20.24.0",
     "enzyme-to-json": "3.3.5",
     "jest": "24.8.0"
   },
   "peerDependencies": {
-    "cozy-ui": "^20.18.0"
+    "@material-ui/core": "3.9.3",
```
cedricmessiant: Isn't it strange that `@material-ui/core` is in deps, peerDeps and devDeps ?

packages/cozy-procedures/src/components/Amount.jsx
```diff
@@ -1,22 +1,74 @@
...
+
+        <div>
+          <Label htmlFor="amount-input">{t('amount.label')}</Label>
+          <Input
+            type="number"
+            id="amount-input"
+            min={min}
+            max={max}
+            value={amount || defaultValue}
+            onChange={e => updateAmount(parseInt(e.target.value))}
```
cedricmessiant: maybe we should use `debounce` for all the inputs to avoid writing the redux state (and thus cause re-render) too often https://lodash.com/docs/4.17.11#debounce
(it is also the case for the personal data form)

packages/cozy-procedures/src/components/Amount.jsx
```diff
@@ -1,22 +1,74 @@
...
 import Topbar from './Topbar'
-import { translate } from 'cozy-ui/transpiled/react/I18n'
+import {
+  Title,
+  translate,
+  Label,
+  Input,
+  Button
+} from 'cozy-ui/transpiled/react/'
+import creditApplicationTemplate from '../templates/creditApplicationTemplate'
```
Crash--: Just though about that, but we have to use a webpack alias to require the template

packages/cozy-procedures/src/components/Duration.jsx
```diff
@@ -1,22 +1,102 @@
...
+    const defaultDuration = get(
+      creditApplicationTemplate,
+      'procedureData.duration.default'
+    )
+    const defaultAmount = get(
+      creditApplicationTemplate,
+      'procedureData.amount.default'
+    )
+
+    const simulationDuration = duration || defaultDuration
```
cedricmessiant: why can't we set the default duration in the initial state?

y-lohse:
Dependency problems are solved!

packages/cozy-procedures/package.json
y-lohse: Solved now!

packages/cozy-procedures/src/components/Duration.jsx
y-lohse: Because then the Overview screen would never show the "ghost" duration button.

packages/cozy-procedures/src/components/Amount.jsx
y-lohse: I just gave it a quick try, but it's not as easy as I thought. We still want the UI to be responsive, so we would need to store the user value in a state, use that in the UI, and only dispatch later… I don't think we need this extra complexity at the moment.

PR #531, opened 2019-06-26
Feat/match doc

Crash--:
We currently need to init all the stuff in the `Procedure.didMount`. 

- Added an initiated props to tell the app to render components only when true 
- Added all the logics to fetch the docs based on the template (rules + count) 
- Added logics to display fetched documents and create EmptyDocumentHolder if needed

TODO : 
- ~~We need a few redux action since fetching documents can't be done automatically by cozy client yet. Since we'll only create an index on a few fields, we'll need to add filters after the end of the query to get the wanted doc. We can have error / loading etc.~~ It seems that we can do all we want ATM. Let's merge this that way, and refactor it later. 

- In order to not duplicate the file object, we can only store the id and access to the file from cozyClient.getDocumentFromSlice for instance.


packages/cozy-procedures/src/index.js
```diff
@@ -6,15 +6,15 @@ import Overview from './containers/Overview'
 import Amount from './components/Amount'
 import Duration from './components/Duration'
 import PersonalDataForm from './containers/PersonalDataForm'
-import Documents from './components/Documents'
+import DocumentsDataForm from './containers/DocumentsDataForm'
 
 const injectProcedureRoutes = ({ root }) => (
```
ptbrowne: Nit: `inject` is IMHO no the right word. When `inject`ing, I expect a "receiving" object being modified. Here, routes are merely rendered.

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -0,0 +1,121 @@
...
+      fetchDocumentLoading({
+        loading: false,
+        idDoctemplate: documentTemplate
+      })
+    )
+  }
+}
+
+export const { getSlice } = selectors
+export default reducer
```
ptbrowne: I don't understand why we need a reducer/slice for this ? cozy-client already has a redux store managing documents.

ptbrowne: Discussed with @Crash-- via zoom. Since some of the query operations are not supported directly via cozy-client, they need to be implemented downstream and thus need downstream error management. It would be interesting to see if those downstream filtering of documents could be a functionality of cozy-client so that queries directly support client-side filtering for example (and having errors automatically managed for example if a document is not found).

It seems to me though that the loading could be implemented as a selector on cozy-client's state. I am not sure though if we want this since this would mean sharing cozy-client internal structure with the app state.

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -0,0 +1,121 @@
...
+        ui: Object.keys(action.payload).reduce((acc, fieldId) => {
+          acc[fieldId] = {
+            loading: false,
+            error: false
+          }
+          return acc
+        }, {})
+      }
+    },
+    update: (state, action) => {
```
y-lohse: We can probably remove this, since we now have `fetchDocumentSuccess ` etc

Crash--: I think we'll use it when importing manually a doc ?

y-lohse: Yeah I'm not sure. I think we can remove it as part of this PR so we don't feel forced to use it later on if it's not the right solution.

packages/cozy-doctypes/src/administrativeProcedures/AdministrativeProcedure.js
```diff
@@ -46,6 +46,63 @@ class AdministrativeProcedure extends Document {
...
+   * @param {int} docRules.count
+   */
+  static async getFilesByRules(docRules) {
+    const { rules, count } = docRules
+    const cozyRules = {
+      trashed: false,
+      type: 'file',
+      ...rules
+    }
+    //Create an index to be order to query and sort
```
y-lohse:
```suggestion
    //Create an index in order to query and sort
```

packages/cozy-doctypes/src/administrativeProcedures/AdministrativeProcedure.js
```diff
@@ -46,6 +46,63 @@ class AdministrativeProcedure extends Document {
...
+
+    return files
+  }
+  /**
+   * This function is used to populate an array based on the order
+   * from the documentsTemplate and set the linked document from the store
+   * @param {Object} docsFromStore
+   * @param {Object} documentsTemplate
+   */
+  static mergeDocsFromStoreAndTemplate(docsFromStore, documentsTemplate) {
```
y-lohse: I'm not sure it's the right place in doctypes, as this seems very specific to the procedure app (it takes the store as param).

packages/cozy-doctypes/src/administrativeProcedures/AdministrativeProcedure.js
```diff
@@ -46,6 +46,63 @@ class AdministrativeProcedure extends Document {
...
+    let sorted = {}
+    Object.keys(documentsTemplate)
+      .sort(function(a, b) {
+        return documentsTemplate[a].order - documentsTemplate[b].order
```
y-lohse: Why do we have to sort the documents here? Since we're putting them in an object afterwards, the order isn't preserved anyway, right?

packages/cozy-procedures/src/Procedure.jsx
```diff
@@ -1,20 +1,59 @@
...
+      Document.registerClient(client)
+    }
+    initPersonalData(
       get(creditApplicationTemplate, 'personalData.schema.properties', {})
     )
-    this.props.fetchMyself(this.props.client)
+
+    initDocuments(get(creditApplicationTemplate, 'documents'))
+    //Since init is done, we tell the app we can start to render thing
+    setStatusDemarche({ initiated: true })
```
y-lohse: Can we translate `statusDemarche` into english? `procedureStatus` or something like that?

packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -1,20 +1,25 @@
...
 import LoadingDocumentHolder from './documents/LoadingDocumentHolder'
+import DocumentsGroups from './documents/DocumentsGroups'
 import DocumentHolder from './documents/DocumentHolder'
 import creditApplicationTemplate from '../templates/creditApplicationTemplate'
-
+import DocumentsContainer from '../containers/DocumentsDataForm'
 class Documents extends React.Component {
   render() {
-    const { t, router } = this.props
+    const { t, router, data: docsFromStore } = this.props
```
y-lohse: Should we rename the prop?

y-lohse: Have you considered renaming them prop itself?

packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -25,12 +30,20 @@ class Documents extends React.Component {
...
-            <EmptyDocumentHolder />
-          </section>
-        ))}
+        {Object.values(populatedTemplateDocsWithStore).map(
+          (documentTemplate, index) => {
+            const { documents: documentsFromStore } = documentTemplate
+            return (
+              <section key={index}>
+                <Label>{t(`documents.labels.${documentTemplate.label}`)}</Label>
+                <DocumentsGroups
```
y-lohse: Nit: probably `DocumentsGroup`, singular group.

packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -25,12 +30,20 @@ class Documents extends React.Component {
...
+                  documents={documentsFromStore}
+                  templateDoc={documentTemplate}
```
y-lohse: I know we talked about this structure together, but seeing it now, I feel like it was maybe misguided. We call `mergeDocsFromStoreAndTemplate` to merge the 2 data structures, in then we split them again a few lines later :/

packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -47,7 +60,8 @@ Documents.propTypes = {
   t: PropTypes.func.isRequired,
   router: PropTypes.shape({
     goBack: PropTypes.func.isRequired
-  }).isRequired
+  }).isRequired,
+  data: PropTypes.object
 }
 
-export default withRouter(translate()(Documents))
+export default withRouter(translate()(DocumentsContainer(Documents)))
```
y-lohse: i'm in favor of this change, but I'm against having multiple ways to do it at the same time in the codebase. Can you please change this for the other components as well?

packages/cozy-procedures/src/components/documents/DocumentHolder.jsx
```diff
@@ -1,15 +1,21 @@
...
-    <Icon icon="file-type-pdf" size={24} className="u-mr-1" />
-    <span className="u-flex-grow-1">
-      Impots gouv 2019
-      <span className="u-coolGrey">.pdf</span>
-    </span>
-    <Icon icon="cross" size={16} className="u-pr-1" />
-  </Chip>
-)
+const DocumentHolder = ({ document }) => {
+  //!TODO to be removed. No need to check because the component should be mounted only when we have info
```
y-lohse: That's the case now, isn't it?

packages/cozy-procedures/src/containers/DocumentsDataForm.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+const mapStateToProps = state => ({
+  data: getDocuments(state)
+})
+
+const mapDispatchToProps = {
+  fetchDocument
+}
+
+const DocumentsDataFormContainer = Component =>
```
y-lohse: Can we keep the same naming convention as the other containers?

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -0,0 +1,118 @@
+/*eslint no-unused-vars: off*/
+/*eslint no-console: off*/
```
y-lohse: I think we can remove those now.

packages/cozy-procedures/src/components/Documents.jsx
Crash--: In fact, now I just need templateDoc for the `count` attribute. I'll edit that :)

packages/cozy-procedures/src/containers/DocumentsDataForm.jsx
Crash--: Since I'll have to edit other containers... Can I rename all the Container* with a suffix ? :D

packages/cozy-procedures/src/components/documents/DocumentHolder.jsx
Crash--: Done on the next PR (since here we still display an empty component https://github.com/cozy/cozy-libs/pull/542/files#diff-f8d3ccbe6490779f446ce1d9e0ebc62aL31)

packages/cozy-procedures/src/containers/DocumentsDataForm.jsx
y-lohse: Yeah sure, as long as there's only one naming convention I'm happy :)

packages/cozy-doctypes/src/administrativeProcedures/AdministrativeProcedure.js
```diff
@@ -46,6 +46,63 @@ class AdministrativeProcedure extends Document {
...
+        ],
+        limit: count ? count : 1
+      })
+
+    return files
+  }
+  /**
+   * This function is used to populate an array based on the order
+   * from the documentsTemplate and set the linked document from the store
+   * @param {Object} docsFromStore
```
cedricmessiant: I don't like the idea of mentioning the redux store here, the model should not care/know about from where the docs come from

packages/cozy-doctypes/src/administrativeProcedures/AdministrativeProcedure.js
cedricmessiant: same here

packages/cozy-doctypes/src/administrativeProcedures/AdministrativeProcedure.js
```diff
@@ -46,6 +46,63 @@ class AdministrativeProcedure extends Document {
...
+  static mergeDocsFromStoreAndTemplate(docsFromStore, documentsTemplate) {
+    let sorted = {}
+    Object.keys(documentsTemplate)
+      .sort(function(a, b) {
+        return documentsTemplate[a].order - documentsTemplate[b].order
+      })
+      .forEach(key => {
+        sorted[key] = documentsTemplate[key]
+      })
+    Object.keys(sorted).map(key => {
```
cedricmessiant: Are we sure the order is ok here?

packages/cozy-doctypes/src/administrativeProcedures/AdministrativeProcedure.js
```diff
@@ -46,6 +46,63 @@ class AdministrativeProcedure extends Document {
...
+      if (docsFromStore[key] && docsFromStore[key].documents) {
+        sorted[key].documents = docsFromStore[key].documents
+      }
```
cedricmessiant: Can't you move this `if () {}` in the `forEach` above to avoid unnecessary loops?

packages/cozy-doctypes/src/administrativeProcedures/AdministrativeProcedure.spec.js
```diff
@@ -64,4 +64,69 @@ describe('AdministrativeProcedure model', () => {
...
+          label: 'payslip',
+          order: 1,
+          count: 3,
+          rules: {}
+        }
+      }
+
+      const documentsFromStore = {
+        tax_notice: {
+          documents: 'file1'
```
cedricmessiant: isn't `documents` an array in the "real" conditions?

packages/cozy-procedures/src/Procedure.jsx
cedricmessiant: avoid French here?

packages/cozy-procedures/src/Procedure.jsx
```diff
@@ -1,20 +1,59 @@
...
+    fetchMyself(client)
+
+    const { documents: documentsTemplate } = creditApplicationTemplate
+    Object.keys(documentsTemplate).map(document => {
```
cedricmessiant: nothing async here? or maybe we don't need to await?

packages/cozy-procedures/src/Procedure.jsx
```diff
@@ -1,20 +1,59 @@
...
   }
 
   render() {
-    return <div className="u-p-1 u-maw-6">{this.props.children}</div>
+    return (
+      <div className="u-p-1 u-maw-6">
+        {this.props.initiated === true ? (
+          this.props.children
+        ) : (
+          <Spinner size="xxlarge" />
```
cedricmessiant: I understand it avoids unnecessary renders but are we want to block the whole UI until we haven't fetch everything? (the user could see what are the sections)

packages/cozy-procedures/src/Procedure.jsx
Crash--: We don't need to await. It populates `au fur et à mesure`

packages/cozy-procedures/src/Procedure.jsx
Crash--: We don't block until we have fetched, we just block until we have initiated our app.

packages/cozy-procedures/src/redux/documentsDataSlice.js
cedricmessiant: I think that `cozy-client`'s state and this state does not represent the same thing. In `cozy-client`, we store the fetched documents whereas here we need to store the documents that are part of the credit application. Maybe that technically you could mark the documents that are part of the procedure in `cozy-client`'s state but I don't like the idea of using the same state for both.

packages/cozy-procedures/src/containers/DocumentsDataForm.jsx
Crash--: We'll work on that in a future PR

packages/cozy-doctypes/src/File.js
```diff
@@ -72,6 +72,21 @@ class CozyFile extends Document {
       }
     }
   }
+  /**
+   * Method to split both the filename and the extension
+   *
+   * @param {Object} file An io.cozy.files
```
cedricmessiant: use `File`?

packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -25,12 +59,20 @@ class Documents extends React.Component {
...
           <DocumentHolder />
         </div>
-        {fields.map((document, index) => (
-          <section key={index}>
-            <Label>{t(`documents.labels.${document.label}`)}</Label>
-            <EmptyDocumentHolder />
-          </section>
-        ))}
+        {Object.values(populatedTemplateDocsWithFiles).map(
+          (document, index) => {
```
cedricmessiant: nit: `document` is a reserved word

packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -25,12 +59,20 @@ class Documents extends React.Component {
...
+            const { files, count: templateDocumentsCount } = document
+            return (
+              <section key={index}>
+                <Label>{t(`documents.labels.${document.label}`)}</Label>
+                <DocumentsGroup
+                  files={files}
+                  templateDocumentsCount={templateDocumentsCount}
```
cedricmessiant: This prop name does not seem very clear to me. In the end, I think I guessed that it is the number of docs for this "category". I would at least remove `template`. Or maybe something like `size` (size of the group)

packages/cozy-procedures/src/components/Documents.spec.jsx
```diff
@@ -0,0 +1,96 @@
...
+        order: 2,
+        count: 3,
+        rules: {}
+      }
+    }
+    const result = mergeDocsFromStoreAndTemplate({}, documentsTemplate)
+    expect(Object.keys(result)).toEqual(['payslip', 'payslip1', 'tax_notice'])
+  })
+
+  it('sould populated a result based on the template and the store', () => {
```
cedricmessiant: nit: "should populate"

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -0,0 +1,116 @@
...
+      state.ui[idDoctemplate].loading = loading
+    },
+    fetchDocumentSuccess: (state, action) => {
+      const { idDoctemplate, loading, files } = action.payload
+      state.ui[idDoctemplate].loading = loading
+      state.data[idDoctemplate].files = files
+    },
+    fetchDocumentError: (state, action) => {
+      const { idDoctemplate, error } = action.payload
+      state.ui[idDoctemplate].error = error
```
cedricmessiant: shouldn't we set `loading` to false in case of error?

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -0,0 +1,116 @@
...
+      }
+    },
+
+    fetchDocumentLoading: (state, action) => {
+      const { idDoctemplate, loading } = action.payload
+      state.ui[idDoctemplate].loading = loading
+    },
+    fetchDocumentSuccess: (state, action) => {
+      const { idDoctemplate, loading, files } = action.payload
+      state.ui[idDoctemplate].loading = loading
```
cedricmessiant: nit: I prefer to set `loading` to `false` directly here. I can't find an example where we would want to dispatch success and `loading` to `true`

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -0,0 +1,116 @@
...
+    setProcedureStatus: (state, action) => {
+      state.ui.initiated = action.payload.initiated
+    }
+  }
+})
+const selectors = {
+  getFiles: state => get(state, [documentsSlice.slice, 'data'], {}),
+  getInitiated: state =>
+    get(state, [documentsSlice.slice, 'ui', ['initiated']], {})
+}
```
cedricmessiant: No selector for `loading`? We never display a spinner while documents are being fetched (like for id document here: https://app.zeplin.io/project/584a6f4e838090371f59518f/screen/5cfac0c72d682519cab68ee7 )

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -0,0 +1,116 @@
...
+        loading: true,
+        idDoctemplate: documentTemplate
+      })
+    )
+    try {
+      const docWithRules = creditApplicationTemplate.documents[documentTemplate]
+
+      const files = await AdministrativeProcedure.getFilesByRules(docWithRules)
+
+      //const files = {}
```
cedricmessiant: nit: remove this comment?

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -0,0 +1,116 @@
...
+      if (files.data) {
+        dispatch(
+          fetchDocumentSuccess({
+            loading: false,
+            files: files.data,
+            idDoctemplate: documentTemplate
```
cedricmessiant: I'm not sure that `idDoctemplate` and `documentTemplate` are very clear names. It is the key in the documents template, right?

PR #537, opened 2019-06-27
Harvest: Fix style prop to not make React crash

gregorylegarec:
This simple proptype error developed under a preact Cozy-Home was making the whole cozy-bar to crash on React apps. This little error cost me a half day.

Ping @GoOz for information.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
+++ b/packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
@@ -68,7 +68,9 @@ export class LaunchTriggerCard extends PureComponent {
                     disabled={running}
                     onClick={launch}
                     subtle
-                    style="line-height:1.4"
+                    // TODO: Extract this directly in Cozy-UI
+                    // (either with an utility class or a Button prop)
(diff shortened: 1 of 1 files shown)


packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
```diff
@@ -68,7 +68,7 @@ export class LaunchTriggerCard extends PureComponent {
                     disabled={running}
                     onClick={launch}
                     subtle
-                    style="line-height:1.4"
+                    style={{ lineHeight: '1.4' }}
```
ptbrowne: Nit: It's weird to have this here. Styling the button should be done via its props, or utility classes, or if the style is really warranted, it should be extracted out of the render so that the same style object is used and the component not re-rendered.

I am OK for the fix but in those cases, I think adding a TODO here to say that the style should be extracted would be nice.

gregorylegarec: Agree. We discuss this with @GoOz yesterday. For now there is no utility class allowing us to manage this case. However, I should have added added a TODO.

PR #540, opened 2019-06-27
feat: Almost a full template

Crash--:
Missing bank_identity ATM, but this is already a good start :)

### DIFF
--- a/packages/cozy-procedures/src/templates/creditApplicationTemplate.js
+++ b/packages/cozy-procedures/src/templates/creditApplicationTemplate.js
@@ -131,31 +131,110 @@ const documents = {
     label: 'identity_document',
     order: 1,
     count: 1,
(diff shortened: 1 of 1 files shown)


packages/cozy-procedures/src/templates/creditApplicationTemplate.js
```diff
@@ -131,31 +131,114 @@ const documents = {
...
+      $or: [
+        {
+          mime: 'text/plain'
+        },
+        {
+          mime: 'application/pdf'
+        }
+      ],
+      metadata: {
+        $and: [
```
y-lohse: Why do we need the $and operator here? There's only one item in the array, right?

Crash--: You right

PR #542, opened 2019-06-27
feat: Unlink a file

Crash--:
We add to pass a `keyDoc` props in order to get the `key` in the global object.


packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -28,22 +27,28 @@ class Documents extends React.Component {
...
-                <Label>{t(`documents.labels.${documentTemplate.label}`)}</Label>
-                <DocumentsGroups
-                  documents={documentsFromStore}
-                  templateDoc={documentTemplate}
-                />
-              </section>
-            )
-          }
-        )}
+        {Object.keys(populatedTemplateDocsWithStore).map((key, index) => {
```
y-lohse: I think renaming `key` to something like `documentId` would be helpful.

packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -28,22 +27,28 @@ class Documents extends React.Component {
...
+                {t(
+                  `documents.labels.${
+                    populatedTemplateDocsWithStore[key].label
+                  }`
+                )}
+              </Label>
+              <DocumentsGroups
+                documents={documentsFromStore}
+                templateDoc={populatedTemplateDocsWithStore[key]}
+                keyDoc={key}
```
y-lohse: I would also rename this to `documentId` or something like that.

packages/cozy-procedures/src/containers/DocumentsDataForm.jsx
```diff
@@ -1,13 +1,18 @@
...
+} from '../redux/documentsDataSlice'
 
 const mapStateToProps = state => ({
   data: getDocuments(state)
 })
 
 const mapDispatchToProps = {
-  fetchDocument
+  fetchDocument,
+  unlinkDocument
```
y-lohse: That does mean that `DocumentHolder` also receives `fetchDocument` and the list of documents, right?

Crash--: Yep. Is it problematic for you?

y-lohse: On principle yes, the component will get props that it doesn't use or want. But in practice I don't mind, it can stay as it is for me :)

PR #543, opened 2019-06-27
Harvest: Disable button when editing if an error exists

gregorylegarec:
AccountForm was designed to enable the submit button after an error occured, to allow user to re-submit the form.

When editing an account, we do not need the form to be submittable when an error already exists.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
@@ -49,7 +49,7 @@ export class AccountForm extends PureComponent {
    */
   isSubmittable({ dirty, error, initialValues, valid }) {
     const untouched = initialValues && !dirty
-    return error || (valid && !untouched)
+    return !untouched && (error || valid)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
```diff
@@ -49,7 +49,7 @@ export class AccountForm extends PureComponent {
    */
   isSubmittable({ dirty, error, initialValues, valid }) {
     const untouched = initialValues && !dirty
-    return error || (valid && !untouched)
+    return !untouched && (error || valid)
```
edas: Est-ce qu'il existe un état "touché mais ni valide ni erreur" ?

packages/cozy-harvest-lib/test/components/AccountForm.spec.js
```diff
@@ -198,6 +198,33 @@ describe('AccountForm', () => {
       )
     })
 
+    /**
+     * This case corresponds to an account edition of a trigger already having
+     * an error. We are not supposed to submit it without modificiations.
```
ptbrowne:
```suggestion
     * an error. We are not supposed to submit it without modifications.
```

packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
```diff
@@ -47,9 +47,11 @@ export class AccountForm extends PureComponent {
    * @param  {Object} formState See https://github.com/final-form/final-form#formstate
    * @return {Boolean}
    */
+  // Here error is not a validation error but an instance of a
+  // KonnectorJobError, so submitting again is possible
```
ptbrowne: Nit: I would put this comment above the return line, I am not sure putting the comment here will play well with the JSDoc comment just below.

PR #544, opened 2019-06-27
Procedures submit

cedricmessiant:
Submit the procedure on click on the button in Overview. 
What it does:
- create a json containing all data about the procedure (amount, duration, personal information, ...)
- put it somewhere in the drive
- create a zip with this json and the documents
- once the zip is ready (job), display a success message then remove the json file

Remarks:
- I think that `submitProcedure` should be sliced.
- Waiting for the documents management PR to be merged to have documents in state and to be able to add those to the zip.
- Some unit tests would be more than welcome and this process.

### DIFF
--- a/packages/cozy-procedures/src/components/Overview.jsx
+++ b/packages/cozy-procedures/src/components/Overview.jsx
@@ -1,9 +1,13 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { withRouter } from 'react-router'
-import Topbar from './Topbar'
+
 import { creditApplicationTemplate } from 'cozy-procedures'
+import { withClient } from 'cozy-client'
+import { AdministrativeProcedure } from 'cozy-doctypes'
+import CozyRealtime from 'cozy-realtime'
 import {
+  Alerter,
   Title,
   SubTitle,
   Caption,
@@ -12,7 +16,17 @@ import {
 } from 'cozy-ui/transpiled/react'
 import InlineCard from 'cozy-ui/transpiled/react/InlineCard'
 
+import Topbar from './Topbar'
+
 class Overview extends React.Component {
+  realtime = null
+
+  componentWillUnmount() {
+    if (this.realtime) {
+      this.realtime.unsubscribeAll()
+    }
+  }
+
   navigateTo = view => {
     const { location, router } = this.props
     const rootPath = location.pathname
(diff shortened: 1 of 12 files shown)


packages/cozy-procedures/src/components/Overview.jsx
```diff
@@ -1,24 +1,103 @@
...
+  handleRealtime = jsonFile => job => {
+    console.log('realtime: the zip is ready', new Date(), job)
+    Alerter.success('Your procedure is available in your Drive.')
+    this.props.client.collection('io.cozy.files').destroy(jsonFile)
+  }
+
+  submitProcedure = async () => {
+    const { client } = this.props
+    // TODO: id of the Administratif dir
+    const dirId = 'c56777a2ab386f28f6e12c034e0c0162'
```
Crash--: Just added to the template a pathToSave. You can use it with `ensureDirectoryExists` to get the id 

https://github.com/cozy/cozy-libs/pull/540/files#diff-986178431654e5c01cfcb9dabd2331feR254

packages/cozy-procedures/src/components/Overview.jsx
```diff
@@ -12,32 +16,115 @@ import {
...
-      duration,
+      data,
       t
     } = this.props
+    const { amount, duration } = data.procedureData
     return (
       <div>
         <Topbar title={creditApplicationTemplate.name} />
         <Title className="u-mb-2">{t('overview.subtitle')}</Title>
+        <Alerter />
```
y-lohse: That might be a problem if there's already an Alerter instance in the app's tree :/

packages/cozy-procedures/src/components/Overview.jsx
```diff
@@ -12,32 +16,115 @@ import {
 } from 'cozy-ui/transpiled/react'
 import InlineCard from 'cozy-ui/transpiled/react/InlineCard'
 
+import Topbar from './Topbar'
+
 class Overview extends React.Component {
+  realtime = null
```
Crash--: do we need it?

packages/cozy-procedures/src/components/Overview.jsx
```diff
@@ -12,32 +16,115 @@ import {
...
+      const jsonFilename = `${baseFilename}.json`
+      const administrativeProcedure = AdministrativeProcedure.create(
+        this.props.data,
+        creditApplicationTemplate
+      )
+      const response = await client.create(
+        'io.cozy.procedures.administratives',
+        administrativeProcedure
+      )
+      const jsonData = AdministrativeProcedure.createJson(response.data)
```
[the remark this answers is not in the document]
cedricmessiant: To be able to create the file.
Also, the json may not exactly match the objet's attributes in the end, that's why I put its creation in the model

packages/cozy-procedures/src/components/Overview.jsx
cedricmessiant: How would you "fix" this? 
We could either :
- say it's the app's responsibility to display the `Alerter` component
- add a prop like `withAlerter` to know if we want to display it (and override this prop if we don't want to)

y-lohse: I think 1 is ok, especially since we might move away from alert's in the near future.

packages/cozy-procedures/src/components/Overview.jsx
cedricmessiant: I guess that not really since if it is `undefined`, `if (!this.realtime) {` will be `true` too but I prefer that it is explicit, don't you?

PR #552, opened 2019-06-28
(procedures) Upload files

Crash--:
A simple implementation of the upload.

### DIFF
--- a/packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
+++ b/packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
@@ -10,7 +10,8 @@ import {
   FileInput
 } from 'cozy-ui/transpiled/react/'
 
-import creditApplicationTemplate from '../../templates/creditApplicationTemplate'
+import { creditApplicationTemplate } from 'cozy-procedures'
+
 import DocumentsDataFormContainer from '../../containers/DocumentsDataForm'
 const EmptyDocumentHolder = ({
   t,
--- a/packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
+++ b/packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
@@ -1,16 +1,62 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { translate, Button, Icon } from 'cozy-ui/transpiled/react/'
+import get from 'lodash/get'
 
-const EmptyDocumentHolder = ({ t }) => (
-  <Button theme="ghost" extension="full" align="left">
-    <Icon icon="plus" size={16} className="u-mr-1 u-pa-half" />
-    <span>{t('documents.import')}</span>
-  </Button>
+import { withClient } from 'cozy-client'
+import {
+  translate,
+  ButtonLink,
+  Icon,
+  FileInput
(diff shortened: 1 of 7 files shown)


packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
```diff
@@ -1,16 +1,66 @@
...
-  </Button>
+import { withClient } from 'cozy-client'
+import {
+  translate,
+  ButtonLink,
+  Icon,
+  FileInput
+} from 'cozy-ui/transpiled/react/'
+
+import creditApplicationTemplate from '../../templates/creditApplicationTemplate'
```
y-lohse: Nope, please use the new import path.

packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
```diff
@@ -1,16 +1,66 @@
...
+import DocumentsDataFormContainer from '../../containers/DocumentsDataForm'
+const EmptyDocumentHolder = ({
+  t,
+  documentId,
+  client,
+  linkDocumentSuccess
+}) => (
+  <FileInput
+    onChange={async file => {
+      const dirPath = creditApplicationTemplate.pathToSave
```
y-lohse: Would be cool to move this in a method.

packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
Crash--: Woa ! cherrypick. Good catch thanks !

packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
```diff
@@ -1,16 +1,66 @@
...
+        }
+        const dirId = await filesCollection.ensureDirectoryExists(dirPath)
+
+        const createdFile = await client
+          .collection('io.cozy.files')
+          .createFile(file, { dirId, metadata })
+
+        linkDocumentSuccess({ document: createdFile.data, documentId })
+      } catch (uploadError) {
+        console.log('errr', uploadError)
```
cedricmessiant: `console.error`?

packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
cedricmessiant: are we sure we want to upload the docs and create the zip at the same path? Maybe it's ok for now but we could use a `pathToSaveZip` or something

packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
Crash--: It'll be removed in the next PR with conflicts ;)

packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
Crash--: ATM we're not sure where we want to save the files and the content. So let's keep it that way and see what happen next, after a real test :)

Crash--:
thanks for the suggestion @y-lohse. I've a more work to do for the refacto :D

PR #554, opened 2019-06-28
feat(procedures): Add files to zip when we submit the procedure

cedricmessiant:


packages/cozy-procedures/src/components/Overview.jsx
```diff
@@ -17,6 +17,7 @@ import {
 import InlineCard from 'cozy-ui/transpiled/react/InlineCard'
 
 import Topbar from './Topbar'
+import documentsDataSlice from '../../dist/redux/documentsDataSlice'
```
Crash--: to be removed ?

packages/cozy-procedures/src/components/Overview.jsx
```diff
@@ -55,16 +56,31 @@ class Overview extends React.Component {
...
+        files[identifier] = f.id
+      })
+    })
+
+    return files
+  }
+
   submitProcedure = async () => {
-    const { client } = this.props
+    const { client, data } = this.props
```
Crash--: data ? or documentsData ?

packages/cozy-procedures/src/components/Overview.jsx
cedricmessiant: :+1:

PR #555, opened 2019-06-28
Harvest: Fix TriggerErrorInfo bottom margin

gregorylegarec:
When extracting TriggerErrorInfo, margin bottom has been lost.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/infos/TriggerErrorInfo.jsx
+++ b/packages/cozy-harvest-lib/src/components/infos/TriggerErrorInfo.jsx
@@ -67,11 +67,12 @@ export class TriggerErrorInfo extends PureComponent {
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/infos/TriggerErrorInfo.jsx
```diff
@@ -67,14 +67,16 @@ export class TriggerErrorInfo extends PureComponent {
   }
 
   render() {
-    const { error } = this.props
+    const { className, error, ...rest } = this.props
```
Crash--: What can be in `...rest` ? If spreading props can be avoided we should

gregorylegarec: You're right, we can get rid of it here, at least for now.

PR #559, opened 2019-07-01
(procedures) chore: Added a watch task

y-lohse:
We can't do `yarn build --watch` anymore because of the copy-files task that has been appended, but we need to keep the copy file. So this PR creates a dedicated watch command. In watch mode, copy files still needs to be done manually but I think that's ok.


ptbrowne:
Another solution would be to put the copy files part inside a postbuild task. Ths way `yarn build --watch` would work.

PR #561, opened 2019-07-01
(procedures) fix: Set the slider step value

y-lohse:
Avoids being able to select 3.45 months.


cedricmessiant:
it seems that we need a test for this component ;-)

y-lohse:
I know :(

PR #567, opened 2019-07-01
Harvest: Better synchronisation between components in KonnectorModal

gregorylegarec:
This PR make all the components in KonnectorModal up to date when a job is launched, succeed or throw error.

* AccountErrorInfo si displayed if TriggerManager get errors
* Both TriggerManager and LaunchTriggerCard are rendered as running when a job is launched
* DeleteAccountCard is disabled when a job is launched

As this PR is almost an UX quickfix, it will be followed by another one unifying prop names like running/submitting/isJobRunning and error/konnectorJobError

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
@@ -175,6 +175,7 @@ export class KonnectorModal extends PureComponent {
                 />
                 <DeleteAccountCard
                   account={account}
+                  disabled={isJobRunning}
                   onSuccess={dismissAction}
                 />
               </div>
--- a/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
@@ -161,6 +161,7 @@ export class KonnectorModal extends PureComponent {
                   onError={this.handleKonnectorJobError}
                   onLaunch={this.handleTriggerLaunch}
                   onSuccess={this.handleKonnectorJobSuccess}
+                  submitting={isJobRunning}
                 />
                 <TriggerManager
                   account={account}
--- a/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
@@ -13,6 +13,7 @@ import Spinner from 'cozy-ui/transpiled/react/Spinner'
 import { SubTitle } from 'cozy-ui/transpiled/react/Text'
 
 import accountMutations from '../connections/accounts'
+import triggersMutations from '../connections/triggers'
 import * as triggers from '../helpers/triggers'
 import TriggerManager from './TriggerManager'
 import DeleteAccountCard from './cards/DeleteAccountCard'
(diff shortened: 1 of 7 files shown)


packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -90,8 +91,13 @@ export class KonnectorModal extends PureComponent {
...
   }
 
-  handleKonnectorJobError(trigger) {
-    this.setState({ isJobRunning: false, trigger })
+  async handleKonnectorJobError() {
+    const { fetchTrigger } = this.props
+    const { trigger } = this.state
+    this.setState({
+      isJobRunning: false,
+      trigger: await fetchTrigger(trigger._id)
```
ptbrowne: If the fetchTrigger does not succeed, the `isJobRunning` state attribute is not correctly set. I think the update of `trigger` and setting `isJobRunning` should be decoupled.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -235,18 +239,27 @@ export class TriggerManager extends Component {
...
     this.jobWatcher = watchKonnectorJob(await launchTrigger(trigger))
     this.jobWatcher.on('error', this.handleError)
-    this.jobWatcher.on('loginSuccess', () =>
-      this.handleSuccess(onLoginSuccess, [trigger])
-    )
-    this.jobWatcher.on('success', () =>
-      this.handleSuccess(onSuccess, [trigger])
-    )
+    this.jobWatcher.on('loginSuccess', () => this.handleSuccess(onLoginSuccess))
+    this.jobWatcher.on('success', () => this.handleSuccess(onSuccess))
```
drazik: A watcher seems to be created on every `launch` call, and the attached events don't seem to be detached. Shouldn't we detach it after error or success?

packages/cozy-harvest-lib/src/components/cards/DeleteAccountCard.jsx
```diff
@@ -26,6 +27,11 @@ export class DeleteAccountCard extends PureComponent {
 
 DeleteAccountCard.propTypes = {
   ...DeleteAccountButton.propTypes,
+  /**
+   * Indicates if the card should be disabled
+   * @type {boolean}
+   */
+  disabled: PropTypes.func,
```
drazik: It looks like it should be a `PropTypes.bool`

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
gregorylegarec: We could. Actually each event is assumed to occur once : there will be no `error` after `success` and no `success` after `error`.

By the way, we still want to merge this part with the KonnectorJob model. Are you ok to do this in a next PR ?

packages/cozy-harvest-lib/src/components/cards/DeleteAccountCard.jsx
gregorylegarec: Well spotted ! thanks.

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
gregorylegarec: Lazy me :)

gregorylegarec: BTW I am not happy with this pattern. I guess something is wrong in the conception.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
drazik: I was also thinking about potential memory leak. Maybe add a TODO comment if you can't fix this now.

gregorylegarec: https://github.com/cozy/cozy-libs/issues/570 🙏

PR #569, opened 2019-07-01
Display an error and delete files if zip creation fails

cedricmessiant:


packages/cozy-procedures/src/components/Overview.jsx
```diff
@@ -45,6 +45,26 @@ class Overview extends React.Component {
...
+        .find(
+          {
+            dir_id: job.message.dir_id,
+            name: job.message.filename
+          },
+          {
+            indexedFields: ['dir_id', 'name']
+          }
+        )
+        .then(resp => {
```
Crash--: why not an await ?

packages/cozy-procedures/src/components/Overview.jsx
```diff
@@ -45,6 +45,26 @@ class Overview extends React.Component {
...
+          resp.data.forEach(zipFile => {
```
Crash--: can we have multiple zips? Or do you want to remove also the files within the zip? If the second, I don't think this is the wanted behavior. We just want to remove the json file?

cedricmessiant: it's in case we have multiple zip (which should not happen)

packages/cozy-procedures/src/components/Overview.jsx
cedricmessiant: as it is passed as a callback to `realtime.subscribe` I can not `await` it. When I first implemented it I had issues with it but in fact we don't care, I can still give an async callback. I changed it.

PR #571, opened 2019-07-01
feat: Menu mobile and web for upload

Crash--:


packages/cozy-procedures/src/components/documents/menuUpload/MenuUploadWeb.jsx
```diff
@@ -0,0 +1,61 @@
...
+        component={
+          <Button theme="ghost" extension="full" align="left">
+            <Icon icon="plus" size={16} className="u-mr-1 u-pa-half" />
+            <span>{t('documents.import')}</span>
+          </Button>
+        }
+        className="u-db"
+        position={'right'}
+        /**
+         * We need to stop the propagation since when we click on an Menu, since auto-close
```
cedricmessiant: remove this comment (or maybe fix it?)

Crash--: In fact, we stop the propagation by doing a onSelect=false. Will update the commit to reflect it better 👍

PR #575, opened 2019-07-02
fix(cozy-procedure): Update template to reflect latest spec

Crash--:
- Subject in now an array => Subjects 
- Missed on `r` :'(

### DIFF
--- a/packages/cozy-procedures/src/templates/creditApplicationTemplate.js
+++ b/packages/cozy-procedures/src/templates/creditApplicationTemplate.js
@@ -144,7 +144,9 @@ const documents = {
     rules: {
       metadata: {
         classification: 'tax_notice',
-        subject: 'income'
+        subjects: {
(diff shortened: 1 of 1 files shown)


packages/cozy-procedures/src/templates/creditApplicationTemplate.js
```diff
@@ -144,7 +144,9 @@ const documents = {
     rules: {
       metadata: {
         classification: 'tax_notice',
-        subject: 'income'
+        subjects: {
+          $in: ['income']
```
y-lohse: I would have used `$elemMatch` here, but cool if `$in` works as well.

Crash--: I can confirm, it works has expected :) 
Capture d’écran 2019-07-02 à 10 37 43

PR #596, opened 2019-07-03
Handle missing password for encrypted placeholder

CPatchane:
Some connectors don't keep the password, with this fix, if an account is already configured but encrypted fields are missing, we force the placeholder

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm/AccountField.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/AccountField.jsx
@@ -47,6 +47,7 @@ export class AccountField extends PureComponent {
     const {
       disabled,
       hasError,
+      forceEncryptedPlaceholder,
       initialValue,
       label,
       name,
@@ -78,7 +79,8 @@ export class AccountField extends PureComponent {
       }),
       placeholder: getFieldPlaceholder(
         this.props,
-        t(`fields.${name}.placeholder`, { _: '' })
+        t(`fields.${name}.placeholder`, { _: '' }),
+        { forceEncryptedPlaceholder }
       ),
       side: required ? null : t('accountForm.fields.optional'),
       size: 'medium'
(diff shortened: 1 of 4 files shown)


CPatchane:
[[APR] Je n'ai plus les étoiles dans l'input du mot de passe dans la modale de modification](https://trello.com/c/8g9JatNU/1582-apr-je-nai-plus-les-%C3%A9toiles-dans-linput-du-mot-de-passe-dans-la-modale-de-modification)

packages/cozy-harvest-lib/src/helpers/fields.js
```diff
@@ -16,17 +16,22 @@ export const getEncryptedFieldName = name => {
...
  *                        field does not need encrypted placeholder
  */
-export const getFieldPlaceholder = (props, fallback) => {
+export const getFieldPlaceholder = (props, fallback, options = {}) => {
   const { encrypted, initialValue, placeholder } = props
+  const { forcePlaceholder } = options
   return (
-    (encrypted && initialValue && ENCRYPTED_PLACEHOLDER) ||
+    (encrypted &&
+      (initialValue || forcePlaceholder) &&
```
gregorylegarec: Reading this line make me think that maybe `forceEncryptedPlaceholder` is a better option name. Otherwise someone could be suprised to see that a not encrypted field has no placeholder even with this option. What do you think ?

packages/cozy-harvest-lib/src/components/AccountForm/AccountField.jsx
```diff
@@ -47,6 +47,7 @@ export class AccountField extends PureComponent {
     const {
       disabled,
       hasError,
+      hasInitialValues,
```
gregorylegarec: I find this prop name a little bit confusing, especially with initialValues right below. It does not describe enought what it does. But a better name is not easy to find. What to you think about `mode="edition"` or maybe better `forceEncryptedPlaceholder` ? After all, it is exactly what we want to do here.

CPatchane: I also prefer `forceEncryptedPlaceholder` 👍

gregorylegarec:
Thanks @CPatchane 🐻

PR #597, opened 2019-07-03
Add connectWithProcedures HOC

cedricmessiant:
Add a HOC to directly connect a component to the procedures store with the correct context


nono:
Ping @y-lohse @Crash--

PR #598, opened 2019-07-03
Refactor containers/slices

cedricmessiant:
- Replace amount/duration slices by procedureData
- Move `initiated` to `procedureDataSlice`


packages/cozy-procedures/src/redux/procedureDataSlice.js
```diff
@@ -0,0 +1,42 @@
...
+  slice: 'procedureData',
+  reducers: {
+    updateAmount: (state, action) => {
+      state.data.amount = action.payload
+    },
+    updateDuration: (state, action) => {
+      state.data.duration = action.payload
+    },
+    updateStatus: (state, action) => {
+      state.ui.initiated = action.payload.initiated
```
y-lohse: As far as I can see, we only ever changed it from the initial state to `true`. What do you think of changing the action to something like `initiationSuccess` without a payload?

nono:
Ping @Crash-- @y-lohse

PR #608, opened 2019-07-03
[cozy-ci] Add new scripts

kosssi:
On this PR:
- Add small documentation
- Update codeowners
- Add script to allow a script to run more 5 minutes on travis `cozy-travis-wait` from https://github.com/m3t/travis_wait
- Add script to upload apk on downcloud `cozy-push-artifact`


packages/cozy-ci/package.json
```diff
@@ -6,6 +6,7 @@
   "bin": {
     "cozy-ci-github": "./github-comment.sh",
     "cozy-ci-transifex": "./transifex.sh",
-    "cozy_travis_wait": "./travis_wait"
+    "cozy_travis_wait": "./travis_wait",
+    "cozy_push_artifact": "./push_artifact.js"
```
drazik: We should be consistent and use dashes or underscores for every bin

packages/cozy-ci/push_artifact.js
```diff
@@ -0,0 +1,157 @@
...
+    // eslint-disable-next-line no-console
+    console.log(appBuildUrl)
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error(`↳ ❌  Error while uploading: ${error.message}`)
+    process.exit(1)
+  }
+}
+
+run()
```
drazik: Some parts looks a lot like things we have in `cozy-app-publish`. Do you think there is something to do to be more DRY? Would it be worth it?

kosssi: Yes but this script was created by @y-lohse I just copy it and adapt some lines

PR #609, opened 2019-07-04
feat: Display error when conflict during upload

Crash--:
In order to display an error when something went wrong during upload. 

We have to manage the conflict case.

### DIFF
--- a/packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
+++ b/packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
@@ -4,7 +4,7 @@ import get from 'lodash/get'
 import flow from 'lodash/flow'
 
 import { withClient } from 'cozy-client'
-import { withBreakpoints } from 'cozy-ui/transpiled/react/'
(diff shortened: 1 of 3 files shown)


packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
```diff
@@ -41,7 +41,12 @@ class EmptyDocumentHolder extends Component {
 
       linkDocumentSuccess({ document: createdFile.data, documentId })
     } catch (uploadError) {
-      console.log('errr', uploadError)
+      if (uploadError.status === 409) {
+        Alerter.error(t('documents.upload.conflict_error'))
+      } else {
+        Alerter.error(t('documents.upload.error'))
+        console.log('Upload error : ', uploadError)
```
cedricmessiant: remove `console.log` ?

Crash--: I don't know. Since we don't log anything yet in Sentry... With that we can ask the user to tell us what appears in the console.

PR #614, opened 2019-07-04
Better input props

cedricmessiant:
- Some useful props for `Input`/`SelectBox`/etc (`type`, `min`, `max`, `step`, ...) need to be created from the `schema` prop. This code is an adaptation of https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/widgets/BaseInput.js (perhaps you'll have to look at `react-jsonschema-form` code to add/tweak some stuff in our adapters, that's the downside of having our own rendering components
- I had to remove `getDefaulValue` since it was not very useful and I had a bug with `integer` type


packages/cozy-procedures/src/components/form/getInputProps.js
```diff
@@ -0,0 +1,64 @@
...
+    if (schema.type === 'number') {
+      inputProps.type = 'number'
+      // Setting step to 'any' fixes a bug in Safari where decimals are not
+      // allowed in number inputs
+      inputProps.step = 'any'
+    } else if (schema.type === 'integer') {
+      inputProps.type = 'number'
+      // Since this is integer, you always want to step up or down in multiples
+      // of 1
+      inputProps.step = '1'
```
y-lohse: I don't get the difference between `integer` and `number`?

cedricmessiant: I think that `integer` doesn't allow point numbers but I don't see any difference :thinking:

Crash--:
What is the consequence to remove the getDefaultValue() ? Do we lose the `3000` value on the Amount view?

cedricmessiant:
No, it was just a function that was returning `''` for `string` and `number` types in the personal data form. I removed it and directly use `''` as default value for all fields since it is sufficient for now

PR #615, opened 2019-07-04
Various fixes

y-lohse:
See commit titles for details.

### DIFF
--- a/packages/cozy-procedures/package.json
+++ b/packages/cozy-procedures/package.json
@@ -31,7 +31,8 @@
   "peerDependencies": {
     "@material-ui/core": "3.9.3",
     "@material-ui/lab": "^3.0.0-alpha.30",
-    "cozy-ui": "^20.24.0"
+    "cozy-ui": "^20.24.0",
+    "cozy-doctypes": "^1.49.3"
   },
   "jest": {
     "setupFiles": [
--- a/packages/cozy-procedures/src/components/documents/DocumentHolder.jsx
+++ b/packages/cozy-procedures/src/components/documents/DocumentHolder.jsx
@@ -2,7 +2,7 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import {
   translate,
-  Chip,
+  Card,
   Icon,
   Modal,
   SubTitle,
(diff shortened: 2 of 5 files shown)


packages/cozy-procedures/src/components/documents/DocumentHolder.jsx
```diff
@@ -41,7 +41,7 @@ class DocumentHolder extends Component {
             }
           />
         )}
-        <Chip variant="dashed" className="u-w-100">
+        <Card className="u-flex u-flex-row u-flex-items-center">
```
ptbrowne: So simple :)

packages/cozy-procedures/src/components/documents/LoadingDocumentHolder.jsx
```diff
@@ -1,13 +1,13 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { translate, Chip, Icon, Spinner } from 'cozy-ui/transpiled/react/'
+import { translate, Card, Icon, Spinner } from 'cozy-ui/transpiled/react/'
 
 const LoadingDocumentHolder = ({ t }) => (
-  <Chip variant="dashed" className="u-w-100">
+  <Card className="u-flex u-flex-row u-flex-items-center">
```
ptbrowne: Ouh, second time, should TextCard be used in these cases ?

packages/cozy-procedures/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "dependencies": {
     "babel-preset-cozy-app": "1.5.1",
+    "cozy-doctypes": "^1.49.3",
```
Crash--: Why not. I understand the why, but to be coherent we should put it as peerDep ? :d

y-lohse: Damnit yes, I told it --peer.

packages/cozy-procedures/src/components/documents/DocumentHolder.jsx
y-lohse: I guess it's subjective but I really don't find the declaration that complicated. It's not like they were vastly different css rules, they are all flexbox-related and the names are pretty explicit.

PR #622, opened 2019-07-05
Delete files permanently after zip success or error

cedricmessiant:
Will require a cozy-client update


nono:
Ping @ptbrowne @Crash--

PR #624, opened 2019-07-08
feat: Loading state

Crash--:
a few things in this PR : 

- having an "importing" state in order to display a "loading" component. 
- dealing with error during import 
- no need to pass loading: true / false, we already know what we're doing 

In order to have a loading state, we need to init the documents store with the same numbers of document required in the template. Like that, we can use the  `index` to populate at the right place

### DIFF
--- a/packages/cozy-procedures/src/components/Documents.jsx
+++ b/packages/cozy-procedures/src/components/Documents.jsx
@@ -53,15 +53,15 @@ class Documents extends React.Component {
         <Title className="u-ta-center u-mb-2">{t('documents.subtitle')}</Title>
         <CompletedFromDriveStatus />
         {Object.keys(populatedTemplateDocsWithFiles).map(
-          (documentId, index) => {
-            const { files, count } = populatedTemplateDocsWithFiles[documentId]
-            const filesStatusByCategory = filesStatus[documentId]
+          (categoryId, index) => {
+            const { files, count } = populatedTemplateDocsWithFiles[categoryId]
+            const filesStatusByCategory = filesStatus[categoryId]
             return (
               <section key={index}>
                 <Label>
                   {t(
                     `documents.labels.${
-                      populatedTemplateDocsWithFiles[documentId].label
+                      populatedTemplateDocsWithFiles[categoryId].label
                     }`
                   )}
                 </Label>
@@ -69,7 +69,7 @@ class Documents extends React.Component {
                   files={files}
                   filesStatusByCategory={filesStatusByCategory}
                   templateDocumentsCount={count}
-                  documentId={documentId}
+                  categoryId={categoryId}
                 />
               </section>
             )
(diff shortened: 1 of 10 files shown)


packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -55,6 +55,7 @@ class Documents extends React.Component {
         {Object.keys(populatedTemplateDocsWithFiles).map(
           (documentId, index) => {
             const { files, count } = populatedTemplateDocsWithFiles[documentId]
+            const filesStatusByCategory = filesStatus[documentId]
```
y-lohse: Nit: reading this, I now wonder of `documentId` shoudn't be `categoryId`.

packages/cozy-procedures/src/components/documents/EmptyDocumentHolder.jsx
```diff
@@ -18,7 +18,16 @@ class EmptyDocumentHolder extends Component {
...
+    const {
+      documentId,
+      client,
+      linkDocumentSuccess,
+      t,
+      index,
+      fetchDocumentLoading,
+      fetchDocumentError
+    } = this.props
+    fetchDocumentLoading({ idDoctemplate: documentId, index })
```
y-lohse: The `fetch` prefix in this context seem wrong, since there's no real fetching happening here.

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -25,59 +29,76 @@ const documentsSlice = createSlice({
...
       state.ui.initiated = action.payload.initiated
     },
     unlinkDocument: (state, action) => {
-      const { document, documentId } = action.payload
-
-      state.data[documentId].files = state.data[documentId].files.filter(
-        doc => doc._id !== document.id
-      )
+      const { documentId, index } = action.payload
+      state.data[documentId].files[index] = undefined
```
y-lohse: We don't need to change the state.ui here? `error` and `loading` don't need to be reset?

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -86,13 +107,16 @@ const selectors = {
   getCompletedDocumentsCount: getCompletedCount,
   getDocumentsTotal: () => {
     // FIXME: don't use the template directly here
-    return Object.values(creditApplicationTemplate.documents).reduce(
+    const test = Object.values(creditApplicationTemplate.documents).reduce(
       (acc, { count }) => acc + count,
       0
     )
+    console.log('test', test)
+    return test
```
y-lohse: Needs changing :p

packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -104,45 +128,60 @@ export const {
...
+            fetchDocumentSuccess({
+              file: file,
+              idDoctemplate: documentTemplate,
+              index
+            })
+          )
+        })
+      }
+
+      if (files.data.length < docWithRules.count) {
```
y-lohse: It might be good to add a comment for this section, it took me a little bit to understand that we are unsetting the loading status on document holders for which we couldn't find a file.

Edit: Or maybe we can rename the action to something less generic?

ptbrowne:
Can you add a description please ? ;)

PR #625, opened 2019-07-08
Harvest: Manage margin issues

gregorylegarec:
Related to https://github.com/cozy/cozy-ui/pull/1094

For OAuth konnectors, LaunchTriggerCard did not have bottom margin with DeleteButtonCard.

We using a dirty hack here while waiting Cozy-UI to introduce breaking changes for Field label issue.

BEFORE

Capture d’écran 2019-07-08 à 17 28 09

AFTER

Capture d’écran 2019-07-08 à 17 11 53

Capture d’écran 2019-07-08 à 17 11 41

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
@@ -190,6 +190,10 @@ export class AccountForm extends PureComponent {
             ref={element => {
               container = element
             }}
+            // Hacking manually the style here, waiting for Cozy-UI to remove
+            // top margin on Field's label elements.
+            // See https://github.com/cozy/cozy-libs/issues/629
+            style={{ marginTop: '-1rem' }}
           >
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
```diff
@@ -190,6 +190,9 @@ export class AccountForm extends PureComponent {
             ref={element => {
               container = element
             }}
+            // Hacking manually the style here, waiting for Cozy-UI to remove
+            // top margin on Field's label elements.
```
ptbrowne: I think it is very useful to add a link to an issue, this way when we see this text,  we can check the issue to see if it is resolved.

gregorylegarec: Done :thumb

ptbrowne:
:shipit: 👍

PR #626, opened 2019-07-08
Implement Document::create and Document::createOrUpdate with cozy-client

drazik:
These two methods were not usable with cozy-client. Now they are. They will be used in banks' accounts stats service.

### DIFF
--- a/packages/cozy-doctypes/src/Document.js
+++ b/packages/cozy-doctypes/src/Document.js
@@ -220,10 +220,14 @@ class Document {
 
   static create(attributes) {
     if (this.usesCozyClient()) {
-      throw new Error('This method is not implemented yet with CozyClient')
+      return this.createViaNewClient(attributes)
     }
 
     return this.createViaOldClient(attributes)
+  }
+
+  static createViaNewClient(attributes) {
+    return this.cozyClient.create(this.doctype, attributes)
   }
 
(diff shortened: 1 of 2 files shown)


packages/cozy-doctypes/src/Document.js
```diff
@@ -152,12 +152,69 @@ class Document {
...
+    const selector = fromPairs(
+      this.idAttributes.map(idAttribute => [
+        idAttribute,
+        get(attributes, sanitizeKey(idAttribute))
+      ])
+    )
+    let results = []
+    const compactedSelector = withoutUndefined(selector)
+    if (size(compactedSelector) === this.idAttributes.length) {
+      results = await this.queryAll(selector)
```
ptbrowne: I am worried by this method, it means that everytime we create a document, we need to query all the documents ?

packages/cozy-doctypes/src/Document.js
```diff
@@ -152,12 +152,69 @@ class Document {
...
+        const updated = this.addCozyMetadata({
+          ...results[0],
+          ...update
+        })
+
+        return this.cozyClient.save(updated)
+      } else {
+        log(
+          'debug',
+          `[bulkSave] Didn't update ${
```
ptbrowne: bulkSave ?

PR #628, opened 2019-07-08
fix: Remove some proptypes warning

Crash--:
we had a few errors. Let's remove them.

### DIFF
--- a/packages/cozy-procedures/src/components/Documents.jsx
+++ b/packages/cozy-procedures/src/components/Documents.jsx
@@ -58,7 +58,7 @@ class Documents extends React.Component {
             const filesStatusByCategory = filesStatus[categoryId]
             return (
               <section key={index}>
-                <Label>
+                <Label htmlFor="">
                   {t(
                     `documents.labels.${
                       populatedTemplateDocsWithFiles[categoryId].label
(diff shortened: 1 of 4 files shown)


packages/cozy-procedures/src/components/Documents.jsx
```diff
@@ -58,7 +58,7 @@ class Documents extends React.Component {
             const filesStatusByCategory = filesStatus[categoryId]
             return (
               <section key={index}>
-                <Label>
+                <Label htmlFor="">
```
drazik: I think this shows that `htmlFor` should not be a required prop on `Label`. What do you think?

Crash--: Or that we shouldn't use `Label` since `label` should be use with `input`.

Crash--: We should use `<Caption>` or something like that instead.

drazik: OK :+1:

Crash--: After reading the HTML spec, `for` is not mandatory, so let's remove the required

PR #630, opened 2019-07-08
(Procedure) feat: Procedure end screen

y-lohse:
Replaces the Alert's at the end of the process with a modal and a message.

One thing that I don't like too much with this PR is that EndModal ends up with the responsability or redirecting to outside the procedure at the end of the process, and I don't think it's very natural. But the AppLinker has to be there, so I still think it's the best option.


ptbrowne:
> that EndModal ends up with the responsability or redirecting to outside the procedure at the end of the process, and I don't think it's very natural.

Do you mean that you'd prefer if this was an option of the whole Procedure ?

`<Procedure endComponent={<AppLinker />}/>`

If this is what you mean I am not very worried, it seems that the code could be easily replaced with an option on the whole <Procedure /> ? `<Procedure endComponent={<AppLinkerToDrive} />`

packages/cozy-procedures/src/components/EndModal.jsx
```diff
@@ -0,0 +1,98 @@
...
+  Title,
+  translate,
+  Button,
+  ButtonLink
+} from 'cozy-ui/transpiled/react/'
+import AppLinker, {
+  generateUniversalLink
+} from 'cozy-ui/transpiled/react/AppLinker'
+import IllustrationOK from '../assets/ok.svg'
+import IllustrationError from '../assets/error.svg'
```
Crash--: Come on, is it working? What I've done when I tested it? :o

packages/cozy-procedures/src/components/EndModal.jsx
```diff
@@ -0,0 +1,98 @@
...
+      >
+        <ModalContent>
+          <img
+            src={isSuccessful ? IllustrationOK : IllustrationError}
+            className="u-db u-w-4 u-h-4 u-mh-auto u-mb-1"
+          />
+          <Title className="u-mb-2 u-ta-center">
+            {t(`${translationKey}.title`)}
+          </Title>
+          <p className="u-mb-2 u-ta-center">{t(`${translationKey}.text`)}</p>
```
Crash--: Can we remove the translationKey var and put the ternary here? Since in the following line we already know if we are sucessful or not, we can write the right label directly. It'll be easier to find and replace. I'm not a big fan of having a var in the translation key.

packages/cozy-procedures/src/locales/en.json
```diff
@@ -139,5 +137,17 @@
       "main": "%{smart_count} document out of %{total} |||| %{smart_count} documents out of %{total}",
       "rest": "has been gathered by the application. |||| have been gathered by the application."
     }
+  },
+  "success": {
+    "title": "[DÉMO] Your loan request has been generated",
+    "text": "In a live scenario, the loan request would have been transmitted to your bank. At the moment, an archive containing the request informations has been placedin your documents.",
```
Crash--:
```suggestion
    "text": "In a live scenario, the loan request would have been transmitted to your bank. At the moment, an archive containing the request informations has been placed in your documents.",
```

packages/cozy-procedures/src/locales/en.json
```diff
@@ -139,5 +137,17 @@
       "main": "%{smart_count} document out of %{total} |||| %{smart_count} documents out of %{total}",
       "rest": "has been gathered by the application. |||| have been gathered by the application."
     }
+  },
+  "success": {
+    "title": "[DÉMO] Your loan request has been generated",
```
Crash--:
```suggestion
    "title": "[DEMO] Your loan request has been generated",
```

packages/cozy-procedures/src/components/EndModal.jsx
ptbrowne: Does this mean that cozy-procedures rely on a particular loader ? If so I am not sure that it should, since it is a library. A babel plugin could be used to inline the SVGs at transpile time. https://github.com/airbnb/babel-plugin-inline-react-svg

y-lohse: Yes, i'm assuming it relies on an SVG loader. I don't know, it seems fair to be abble to import an SVG or a JSON file and that a loader would be present, but I'll look into the babel inliner plugin.

y-lohse:
> Do you mean that you'd prefer if this was an option of the whole Procedure ?


No, I meant that I think it would make more sense if all the end screen logic was handled by the Overview component — making the final requests, showing/hiding the modal, redirecting to other apps and redirecting to the root of the app.

packages/cozy-procedures/src/components/EndModal.jsx
ptbrowne: >  I don't know, it seems fair to be abble to import an SVG or a JSON file and that a loader would be present,

For example, I am not sure importing the cozy-procedures lib in codesandbox would work in codesandbox.

Researching just a bit what create-react-app used led me to https://github.com/facebook/create-react-app/issues/5293#issuecomment-427085505, it seems create-react-app uses import names to automatically switch behavior, seems very interesting.

PR #631, opened 2019-07-09
feat: Display fully completed component when docs are ok

Crash--:
When we have the right numbers of linked documents, we change the display 

Capture d’écran 2019-07-09 à 09 21 06

### DIFF
--- a/packages/cozy-procedures/src/components/overview/DocumentsFullyCompleted.jsx
+++ b/packages/cozy-procedures/src/components/overview/DocumentsFullyCompleted.jsx
@@ -1,0 +1,35 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { InlineCard, Icon } from 'cozy-ui/transpiled/react'
+const DocumentsFullyCompleted = ({ documents, navigateTo }) => {
+  const flatenedFiles = []
+  Object.values(documents).map(document => {
+    Object.values(document['files']).map(file => {
+      flatenedFiles.push(file)
+    })
+  })
+  return (
+    <InlineCard
+      className="u-c-pointer u-p-1 u-flex u-flex-column"
+      onClick={() => navigateTo('documents')}
+    >
+      {flatenedFiles.map((file, index) => {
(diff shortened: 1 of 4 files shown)


packages/cozy-procedures/src/components/overview/DocumentsFullyCompleted.jsx
```diff
@@ -0,0 +1,35 @@
...
+    })
+  })
+  return (
+    <InlineCard
+      className="u-c-pointer u-p-1 u-flex u-flex-column"
+      onClick={() => navigateTo('documents')}
+    >
+      {flatenedFiles.map((file, index) => {
+        let style = 'u-ellipsis u-flex'
+        if (index !== 0) style += ' u-mt-1'
```
y-lohse: Why do we need a top margin on the first element, since there's a padding on the wrapper?

Crash--: This is because we don't need a top margin on the first element ;)

y-lohse: Oh right, I misread that. Nit, but we use single directional margins whenever possible and usually it's towards the bottom, so using bottom margins would be more appropriate. Or given the layout, a small top and bottom padding might be even better!

Crash--: Will fix it soon thanks for the feedback

packages/cozy-procedures/src/components/overview/DocumentsFullyCompleted.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { InlineCard, Icon } from 'cozy-ui/transpiled/react'
+const DocumentsFullyCompleted = ({ documents, navigateTo }) => {
+  const flatenedFiles = []
```
ptbrowne:
```suggestion
  const flattenedFiles = []
```

PR #633, opened 2019-07-09
feat(harvest): Do not autofocus identifier field on iPhone

ptbrowne:
Autofocus is buggy on iPhone, it focuses the input but the keyboard is not shown.
D

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm/AccountField.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/AccountField.jsx
@@ -10,6 +10,17 @@ import {
   predefinedLabels,
   ROLE_IDENTIFIER
 } from '../../helpers/manifest'
+import { isIOS } from 'cozy-device-helper'
+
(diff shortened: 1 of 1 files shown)


packages/cozy-harvest-lib/src/components/AccountForm/AccountField.jsx
```diff
@@ -11,6 +11,17 @@ import {
...
 } from '../../helpers/manifest'
 
+// On iOS, focusing an input not in the same tick as the user action is buggy.
+// The field is focused but the keyboard does not show up,
+// See https://github.com/apache/cordova-plugin-wkwebview-engine/pull/37/#issuecomment-308321094
+const canAutofocus = () => {
+  const userAgent = typeof navigator !== 'undefined' ? navigator.userAgent : ''
+  if (userAgent.includes('AppleWebKit') && userAgent.includes('Mobile')) {
+    return false
+  }
```
Crash--: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-device-helper/src/platform.js#L20

Issue #637, opened 2019-07-09
Harvest: move default base directory from locale to code


gregorylegarec:
Harvest locales are defining a `default.baseDir` locale which plays both the role of translation and default base directory.

The default base directory should be moved directly in code, and translation be provided by another key than `default`.

[ ] Provide translation with a `folder.administrative` locale key
[ ] Merge changes with both locales to avoid conflicts
[ ] Remove the previous locale key

PR #638, opened 2019-07-09
Harvest: Handle folder baseDir declared in manifest

gregorylegarec:
Related to https://github.com/cozy/cozy-libs/issues/637

When switching to Harvest, we accidentally forgot a recent feature used by Facebook konnector : the manifest attribute to declare a base directory (https://github.com/konnectors/cozy-konnector-facebook/blob/8baa1881f365db8828bfdf8650e6ec40722f8d35/manifest.konnector#L16)

The code in Cozy-Home was recently removed here https://github.com/cozy/cozy-home/commit/7203bb74bbf7b34443ac089a238edc9e4d053a6f#diff-7b43428512328530cec55542e1b2d84cL30

As OAuth account are now able to have the account id before creating the folder, we do not need the `uuid()` call in the old Cozy-Home code, and we can just let the default folder be used.

So this PR is basically just handling the `$photos` value for `baseDir`.

Based on [our specifications](https://paper.dropbox.com/doc/Tech-Konnectors-Dossier-de-destination--AgnU3mjTJJxUiVOe2UQCZLspAg-IqNRKAavd2dImlrhYVHs5), I have also added a way to handle a custom `defaultPath`.

So now, we will be able to declare in a manifest:
```json
folders: [
  {
    "defaultDir": "$photos/My photos from my konnector",
  }
]
```

which will creates a `/Photos/My photos from my konnector` folder.


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -100,9 +99,12 @@ export class TriggerManager extends Component {
...
     if (konnectors.needsFolder(konnector)) {
-      const path = `${t('default.baseDir')}/${konnector.name}/${slugify(
-        accounts.getLabel(account)
-      )}`
+      // TODO: remove previous legacy key `default.baseDir`
+      // https://github.com/cozy/cozy-libs/issues/637
+      const path = konnectors.buildFolderPath(konnector, account, {
+        administrative: t('folder.administrative'),
+        photos: t('folder.photos')
+      })
```
Crash--: Ça me parait un peu trop simple comme gestion du dossier `Photos` par exemple. L'utilisateur peut l'avoir renommé ou déplacé et pourtant, on va lui re-créer à chaque fois à l'endroit où il ne veut pas. J'ai pas l'impression que ce cas est géré, mais je me trompe peut-être ? 

La stack vient mettre un attribut sur le dossier `Photos` par exemple, pourquoi ne pas l'utiliser pour récupérer le bon chemin ? 

Sur Drive on fait comme ça : https://github.com/cozy/cozy-drive/pull/1243/files#diff-8f9d41bc1595335b09bde52219ed7844R7

gregorylegarec: Probablement en effet. Ici on est surtout isofonctionnel avec l'ancien existant dans la Home. Je pense que cela serait une bonne amélioration. Pendant mon absence tu peux voir avec @poupotte ce qu'elle en pense ?

ptbrowne: A-t-on du nouveau la dessus ?

poupotte: Je souhaiterais effectivement dès maintenant avoir une gestion plus propre des dossiers de destination. Je pense que ça vaut le coup de travailler là dessus.
A ma connaissance greg a regardé à ça la semaine dernière, il ne devrait plus rester beaucoup de temps.

ptbrowne:
If I understand well, this is a bug fix. What Quentin is suggesting is a feature improvement, I'd be personally OK to merge this code and file an issue for the improvement.

edas:
> If I understand well, this is a bug fix. What Quentin is suggesting is a feature improvement, I'd be personally OK to merge this code and file an issue for the improvement.

On est de toutes façons sur un nouveau code. Tout ce qu'on y fait est nouveau.
On a eu la même discussion que Quentin dans les daily Backformers et il y a eu une volonté assez claire de Zoe et moi pour dire que ça nous semblait indispensable. On peut effectivement boucler celle ci et en ouvrir une autre mais ça me semble à priori aussi bien de garder celle là, non ?

packages/cozy-harvest-lib/src/helpers/konnectors.js
```diff
@@ -103,6 +109,109 @@ export const needsFolder = konnector => {
...
+
+/**
+ * Parse the given folder path using the given `variables` object.
+ * Variables are injected in `buildFolderPath` function. Available variable are:
+ * $konnector (konnector name) and $account (account label, i.e. id or name)
+ * @param  {String} path      Path to parse
+ * @param  {Object} variables Object mapping variable to actual values
+ * @return {String}           Parsed path
+ */
+const parseFolderPath = (path, variables = {}) => {
```
ptbrowne: IMHO, we are not parsing here, we are `rendering`.

```suggestion
const renderFolderPath = (templatePath, variables = {}) => {
```

packages/cozy-harvest-lib/src/helpers/konnectors.js
```diff
@@ -103,6 +109,109 @@ export const needsFolder = konnector => {
...
+  // Trim `/` and avoid multiple `/` characters with regexp
+  const sanitizedPath = trim(path.replace(/(\/+)/g, '/'), '/')
+
+  // Let's get only full variable name limited by '/'. We want to avoid false
+  // positive like parsing `$variableInString` to `valueInString`
+  const segments = sanitizedPath.split('/')
+  return segments
+    .map(segment => variables[segment.slice(1)] || segment)
+    .join('/')
+}
```
ptbrowne: I

packages/cozy-harvest-lib/src/helpers/konnectors.js
```diff
@@ -103,6 +109,109 @@ export const needsFolder = konnector => {
...
+
+  // Look for variable name into folders but without $ prefix
+  const parsedBaseDir = folders[baseDir.slice(1)] || DEFAULT_LOCALIZED_BAR_DIR
+  // Trim `/` and avoid multiple `/` characters with regexp
+  return trim(parsedBaseDir.replace(/(\/+)/g, '/'), '/')
+}
+
+/**
+ * Parse the given folder path using the given `variables` object.
+ * Variables are injected in `buildFolderPath` function. Available variable are:
```
ptbrowne: "Variables are injected in `buildFolderPath` function.", as this is an implementation detail, I would leave this out of the comment of the functions.

```suggestion
 * Available variables are:
```

Also, this is not a `parse`, it's a `render` (like for example Handlebars.render)

packages/cozy-harvest-lib/src/helpers/konnectors.js
```diff
@@ -103,6 +109,109 @@ export const needsFolder = konnector => {
...
+    .map(segment => variables[segment.slice(1)] || segment)
+    .join('/')
+}
+
+/**
+ * Build folder path for a given konnector and a given account.
+ *
+ * If konnector.folders[0].defaultDir exists, it is used as default directory.
+ * If konnector.folders[0].defaultPath exists, it is used as default folder path
+ * as well.
```
ptbrowne: It seems defaultDir and defaultPath are both paths and this makes it hard to understand. defaultDir and defaultSubDir would be easier to understand..

gregorylegarec: Well spotted and great suggestion, thanks !

gregorylegarec:
I updated this PR with @ptbrowne last feedback. I think that we should merge it now as it fixes a missing behaviour from Cozy-Home.

The Magic Folders are implemented in cozy-doctype https://github.com/cozy/cozy-libs/pull/681, but it's not clear if we should provide a single value in manifest to describe the whole destination folder (for example : `/$administrative/$konnector/$account`), or keep two values. A argument in favor of keeping two values is that it allows contributor to easily change the base directory (for example `$administrative` to `$photos` without knowing anything about how subdirectory are built, i.e. without providing also the `/$konnector/account` part. The same argument may apply when a contributor wants to only change the subdirectory template, without worrying about the base dir. I think technically the two concernes are very close, but functionnally there a tiny difference. It's not a strong opinion but we must have this in mind if we choose to only keep one parameter to handle the whole destination folder path.

Crash--:
Just updated the code with our latest discussion: 

- We only want to provide a single value in the manifest , something like `/$administrative/$konnector/$account`. No need to `defaultDir` and `defaultPath`. 

In order to do the right checks, we kept this notion internally with a `baseDir` and a `subDir`.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
Crash--: Pour le moment, on prend pas en compte ça. On a une autre carte pour ça :)

packages/cozy-harvest-lib/src/helpers/konnectors.js
```diff
@@ -1,5 +1,11 @@
 import get from 'lodash/get'
 import has from 'lodash/has'
+import trim from 'lodash/trim'
+
+import * as accounts from './accounts'
+
+// Default name for base directory
+const DEFAULT_LOCALIZED_BASE_DIR = 'Administrative'
```
edas: You do not try to localize it ?

packages/cozy-harvest-lib/src/helpers/konnectors.js
Crash--: We have a baseDir and a subDir now

packages/cozy-harvest-lib/src/helpers/konnectors.js
```diff
@@ -103,6 +109,130 @@ export const needsFolder = konnector => {
...
+ *
+ * @param  {Object} konnector Konnector document
+ * @param  {Object} account   Account document
+ * @param  {Object} folders   Object containing a mapping from folder
+ * identifiers (ex: $administrative) to their localized values (ex:
+ * Administratif).
+ * @return {String}           The result path
+ */
+
+export const buildFolderPath = (konnector, account, folders = {}) => {
```
edas: I was expecting something simplier, like (not tested) : 

```javascript
const path = get(konnector, 'folders[0].defaultDir', '/$administrative/$konnector/$account')
const fullPath = (path[0]=='$') ? `/${path}` : path
const replace = { ...folders,  konnector: konnector.name, account: accounts.getLabel(account))
const op = function (txt, keys) {
  const [key, ...rest] = keys
  const res = txt.replace('$' + key, replace[key])
  return (rest.length > 0) ? op(res, rest) : res
}
return op(fullPath)
```
It has not all the complex behaviour that you had, but do we need them ? In my understanding, we start from zero or near zero (only two connectors using this kind of property, not live).

Crash--: Ça peut fonctionner mais ça veut dire qu'il faut qu'on soit vachement vigilant sur ce chemin dans le manifest. 

Car là, ça ne va pas vérifier qu'on commence avec les bons répertoires, que si on met juste `$photos` ça ne va pas aller automatiquement dans `Photos/konnecteur/account` mais juste dans `Photos` et donc potentiellement mettre un sacré binz. 

Au final, même si on a la main sur les connecteurs, ça me parait léger de compter sur le fait que notre attention sera toujours au top lors des merges sur ce point et donc j'ai la préférence pour du code défensif quitte à le complexifier (un peu). 

WDYT?

edas: Ok. On a le code complexe déjà fait, tu penses qu'il est utile, laissons le. Je n'ai pas de raison de bloquer.

PR #639, opened 2019-07-10
Interapp: Fix missing _id from intent documents

CPatchane:
Since the interapp lib used to be based on `_id` attribute of documents, we add it if it's missing.
Missing `_id` here was causing `invalid event id` or `unexpected handshake message from intent service` errors
For now we don't know why this attribute is now missing or not sent by the stack anymore.
Edit: We should use `id` only for documents using jsonapi from the cozy-stack. For now we keep the both but the `_id` usage should be removed at term.

### DIFF
--- a/packages/cozy-interapp/src/request.js
+++ b/packages/cozy-interapp/src/request.js
@@ -4,9 +4,11 @@ class Request {
   }
 
   get(id) {
-    return this.stackClient
-      .fetchJSON('GET', `/intents/${id}`)
-      .then(resp => resp.data)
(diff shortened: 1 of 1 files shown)


CPatchane:
[[STO] L'intent d'installation ne fonctionne plus](https://trello.com/c/xTSM2Gff/1593-sto-lintent-dinstallation-ne-fonctionne-plus)

gregorylegarec:
Probably since https://github.com/cozy/cozy-libs/commit/0e52ea86a4309f5c41d29d2305f234dba613bd86#diff-5408683acb5bc4e73d7142f408234ed5R3

PR #640, opened 2019-07-10
feat: Add webpack-expose-package-info-plugin

ptbrowne:
Putting this in WIP, I have the feeling that the plugin could be a bit
more useful if it could supply to the application a bit more than the
version.

```patch
- webpack-expose-package-versions-plugin
+ webpack-expose-package-information-plugin
```

with an usage like this ?

```js
new PackageInformationPlugin({
  packages: ['cozy-bar', 'cozy-ui'],
  fields: ['version']
})
```

### DIFF
--- a/packages/webpack-expose-package-info-plugin/README.md
+++ b/packages/webpack-expose-package-info-plugin/README.md
@@ -1,0 +1,52 @@
+Expose package information
+======================
+
+This plugin is useful if you want to expose chosen package information to
+a bundle created via webpack.
+
+The mechanism uses the DefinePlugin to define a variable that will contain
+package informations.
+
+### Usage
+
+```
+const PackageInfoPlugin = require('webpack-expose-packages-version')
+
+new PackageInfoPlugin({
+  packages: ['cozy-bar', 'cozy-ui']
+})
+```
+
+#### Exposed variable
+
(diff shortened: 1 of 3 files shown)


y-lohse:
MmmhI can't really think of any other interesting field for debuging purposes in the package.json. Do you have something specific in mind?

ptbrowne:
For example if we had the repository we could make a clickable link to the latest release ?

y-lohse:
Ok, good point :) And I guess package.json files can host custom fields too.

ptbrowne:
Ready to review ;)

packages/webpack-expose-package-info-plugin/README.md
```diff
@@ -0,0 +1,52 @@
...
+ })
+```
+
+You can now access package information in your app.
+
+```
+/* global PACKAGES */
+
+console.log(PACKAGES)
+/* {"cozy-bar": {"version": "1.0.0"}, "cozy-ui": {"version": "6.26.2"}} */
```
Crash--: It works on mobile since we build the bar, but it will give us a false information on the web version. In our app, we should take care of that( better to hide a wrong information, right ? :D)

ptbrowne: That's a good point ! We'll put a conditional in our config.

PR #641, opened 2019-07-10
fix(cozy-harvest-lib): Remove cozy-realtime from dependencies

drazik:
cozy-realtime was present in dependencies, devDependencies and
peerDependencies with different version. Keeping only the highest
version in dependencies should by sufficient

This fixes the following warning: https://travis-ci.org/cozy/cozy-libs/builds/556705656#L793

### DIFF
--- a/packages/cozy-harvest-lib/package.json
+++ b/packages/cozy-harvest-lib/package.json
@@ -26,7 +26,6 @@
     "watch:doc:react": "(cd ../.. && TARGET=cozy-harvest-lib yarn watch:doc:react)"
   },
   "dependencies": {
-    "cozy-realtime": "^3.1.1",
     "date-fns": "^1.30.1",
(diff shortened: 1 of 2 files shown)


edas:
I do not understand why we have realtime, client and ui as dev dependencies. It seems an mistake to me

Crash--:
Because all of our internal dep in libs has to be in peerDep. That means that the APP should provide the dep not the lib itself. 

But for UT, snapshot or else, we need to have these dep so devDep.

Conclusion: we need peerDep & devDep and not dep

drazik:
> Conclusion: we need peerDep & devDep and not dep

OK so I do this change. Thanks for the explanation

Crash--:
We use peerDep to be sure to have the same exact version between the app and the lib

PR #647, opened 2019-07-11
fix(cozy-procedures): Few visual issues

Crash--:
- Put bold on  numberOfDependants 
- Better alignment on Overview 

Capture d’écran 2019-07-11 à 09 24 56

### DIFF
--- a/packages/cozy-procedures/src/components/overview/PersonalDataFullyCompleted.jsx
+++ b/packages/cozy-procedures/src/components/overview/PersonalDataFullyCompleted.jsx
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import snarkdown from 'snarkdown'
 import capitalize from 'lodash/capitalize'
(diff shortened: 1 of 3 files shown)


packages/cozy-procedures/src/components/overview/PersonalDataFullyCompleted.jsx
```diff
@@ -23,10 +23,11 @@ const PersonalDataFullyCompleted = ({ navigateTo, personalData, t }) => {
   } = personalData
   return (
     <InlineCard
-      className="u-c-pointer u-p-1 u-flex u-flex-column"
+      className="u-c-pointer u-ph-1 u-pt-half u-pb-1 u-flex u-flex-column"
```
y-lohse: Ideally though, we would do it like the DocumentsFullyCmpleted: horizontal padding of 1 and vertical padding of half on the container, and vertical padding of half on all the children :)

Crash--: thanks !

PR #656, opened 2019-07-15
(Authentication) Execute `onAuthenticated` before redirecting

drazik:
In banks, we want to do some things when the user is successfuly logged in, but before redirecting him to `loginPath` (initialize the cozy-bar, actually). I moved `onAuthenticated` before the redirection. I think this is the right place for doing this: if something needs to be done just before redirecting, this looks like the best place to do it; and if something needs to be done after redirecting, it can be done via the component that will be rendered on the `loginPath`.

I had to use a dirty `setTimeout` to make tests pass. Without it, `history.replace` is reported as not called, but it's because the expectation seems to be executed before the code itself.

Any thoughts?

### DIFF
--- a/packages/cozy-authentication/src/MobileRouter.jsx
+++ b/packages/cozy-authentication/src/MobileRouter.jsx
@@ -71,7 +71,7 @@ export class MobileRouter extends Component {
     try {
       if (saved && saved.oauthOptions) {
         client.stackClient.setOAuthOptions(saved.oauthOptions)
-        client.login({ uri: saved.uri, token: saved.token })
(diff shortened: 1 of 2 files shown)


drazik:
> But in Banks can't you use `client.on('login')` to init the bar?

Actually that's what we do, but we need to change it because we need to ensure the bar is initialized before we redirect to the app's homepage: we use some bar's APIs in it (mostly `setTheme`). But since the `MobileRouter`'s `client.on('login')` callback is registered before banks' one (in which we do `cozy.bar.init()`), we have a race condition that fails sometimes (more details coming soon in the according PR on banks :))

PR #657, opened 2019-07-15
[WIP] refactor(cozy-procedures): Use cozyclient to fetch myself

Crash--:
A naïve refactor to use Cozy-Client in order to fetch MySelf.

Still need a redux action for `success` since we do a few things in it. 

Need to : 
- Manage query error 
- Manage when data ===0 
- Pass fetchMyselfStatus to children to add spinner where needed


Crash--:
Closing this PR since I will not work on it in the upcoming weeks

PR #662, opened 2019-07-17
feat: Add two small utilities

ptbrowne:
I used them at some point, thought it would be useful to have them here.


packages/cozy-codeshifts/src/utils/add-import.js
```diff
@@ -0,0 +1,16 @@
+const addImport = (root, j, options) => {
```
drazik: Can you add a comment with a before / after example? I think this would be easier so figure out what it does than by reading the code.

drazik: Thanks!

PR #669, opened 2019-07-19
feat: Lib option to babel-preset-cozy-app

ptbrowne:
- No transform of imports
- Use of babel helpers (should we have that by default by the way ?)


Crash--:
I think we can add babel helper by default (reading https://brunoscopelliti.com/a-simple-babel-optimization-i-recently-learned/)

PR #678, opened 2019-07-19
WIP Use PageLayout for Wizard

ptbrowne:
The wizard uses custom classes for a lot a functionality.

This would lead to less code inside wizard.styl and more coherence.

- PageLayout could be used
- InputGroup should be used instead of the custom dualfield classes

### DIFF
--- a/packages/cozy-authentication/src/steps/SelectServer.jsx
+++ b/packages/cozy-authentication/src/steps/SelectServer.jsx
@@ -11,7 +11,9 @@ import {
   Input,
   MainTitle,
   Icon,
-  InputGroup
+  InputGroup,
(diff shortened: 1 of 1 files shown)


nono:
Ping @ptbrowne

ptbrowne:
Since I did not work on the wizard, I could not finish the work on this PR, I still think it is interesting, and should more work be done on the wizard, I think this could be used.

Crash--:
This component will be deprecated soon since the auth will be served by the cloudery in order to be able to customize it. The work is already in production, we need to update this component. I'll work on it this week since we've a short deadline for a customer ;)

PR #681, opened 2019-07-22
Doctypes: Add Folder class

gregorylegarec:
The Folder class mutualize our logic related to folder.

The first two methods added in this PR allow to retrieve referenced folders.

It's based and the code from Drive (https://github.com/cozy/cozy-drive/pull/1243/files#diff-8f9d41bc1595335b09bde52219ed7844R7)

We would need to use it in Harvest to deal with baseDir declared in konnector manifests, for example to retrieve the `$photo` folder even if its path has changed. See https://github.com/cozy/cozy-libs/pull/638

UPDATE: After discussion, it appears that we could expose the concept of magic folders directly in the API, and create a method `ensureMagicFolder`

### DIFF
--- a/packages/cozy-doctypes/src/Folder.js
+++ b/packages/cozy-doctypes/src/Folder.js
@@ -1,0 +1,88 @@
+const Application = require('./Application')
+const CozyFile = require('./File')
+
+/**
+ * Class representing the folder model.
+ * @extends CozyFile
+ */
+class CozyFolder extends CozyFile {
+  /**
+   * Create a folder with a reference to the given document
+   * @param  {String}  path     Folder path
+   * @param  {Object}  document Document to make reference to. Any doctype.
+   * @return {Object}  Folder document
+   */
+  static async createFolderWithReference(path, document) {
+    const collection = this.cozyClient.collection(CozyFile.doctype)
+    const dirId = await collection.ensureDirectoryExists(path)
+    await collection.addReferencesTo(document, [
+      {
+        _id: dirId
+      }
+    ])
+
+    const { data: dirInfos } = await collection.get(dirId)
+
+    return dirInfos
+  }
+
+  /**
+   * Returns an array of folder referenced by the given document
+   * @param  {Object}  document  Document to get references from
+   * @return {Array}             Array of folders referenced with the given
+   * document
+   */
+  static async getReferencedFolders(document) {
+    const { included } = await this.cozyClient
+      .collection(CozyFile.doctype)
+      .findReferencedBy(document)
+    return included.filter(folder => !CozyFolder.isTrashed(folder))
+  }
+
+  /**
+   * Returns an unique folder referenced with the given reference. Creates it
+   * if it does not exist.
+   * @param  {String}  path      Path used to create folder if the referenced
+   * folder does not exist.
+   * @param  {Object}  document  Document to create references from
+   * @return {Objet}             Folder referenced with the give reference
+   */
+  static async ensureFolderWithReference(path, document) {
+    const existingFolders = await CozyFolder.getReferencedFolders(document)
+    if (existingFolders.length) return existingFolders[0]
+
+    const collection = this.cozyClient.collection(CozyFile.doctype)
+    const dirId = await collection.ensureDirectoryExists(path)
+    await collection.addReferencesTo(document, [
+      {
+        _id: dirId
+      }
+    ])
+
+    const { data: dirInfos } = await collection.get(dirId)
+
+    return dirInfos
+  }
+
+  /**
+   * Indicates if a folder is in trash
+   * @param  {Object}  folder `io.cozy.files` document
+   * @return {Boolean}        `true` if the folder is in trash, `false`
+   * otherwise.
+   */
+  static isTrashed(folder) {
+    return /^\/\.cozy_trash/.test(folder.attributes.path)
+  }
+}
+
+/**
+ * References used by the Cozy platform and apps for specific folders.
+ */
+CozyFolder.refs = {
+  ADMINISTRATIVE: `${Application.doctype}/administrative`,
+  PHOTOS: `${Application.doctype}/photos`,
+  PHOTOS_BACKUP: `${Application.doctype}/photos/mobile`,
+  PHOTOS_UPLOAD: `${Application.doctype}/photos/upload`
+}
+
+module.exports = CozyFolder
--- a/packages/cozy-doctypes/src/Folder.js
+++ b/packages/cozy-doctypes/src/Folder.js
@@ -27,6 +27,39 @@ class CozyFolder extends CozyFile {
   }
 
   /**
+   * Returns a "Magic Folder", given its id
+   * @param  {String} id Magic Folder id. `CozyFolder.magicFolders` contains the
+   * ids of folders that can be magic folders.
+   * @param {String} path Default path to use if magic folder does not exist
+   * @return {Object} Folder document
+   */
+  static async ensureMagicFolder(id, path) {
+    const magicFolderDocument = {
+      _type: Application.doctype,
+      _id: id
+    }
+    const folders = await CozyFolder.getReferencedFolders(magicFolderDocument)
+    const existingMagicFolder = folders.length ? folders[0] : null
+
+    if (existingMagicFolder) return existingMagicFolder
+
+    const magicFoldersValues = Object.values(CozyFolder.magicFolders)
+    if (!magicFoldersValues.includes(id)) {
+      throw new Error(
+        `Cannot create Magic folder with id ${id}. Allowed values are ${magicFoldersValues.join(
+          ', '
+        )}.`
+      )
+    }
+
+    if (!path) {
+      throw new Error('Magic folder default path must be defined')
+    }
+
+    return CozyFolder.createFolderWithReference(path, magicFolderDocument)
+  }
+
+  /**
    * Returns an array of folder referenced by the given document
    * @param  {Object}  document  Document to get references from
    * @return {Array}             Array of folders referenced with the given
@@ -78,7 +111,7 @@ class CozyFolder extends CozyFile {
 /**
  * References used by the Cozy platform and apps for specific folders.
  */
-CozyFolder.refs = {
+CozyFolder.magicFolders = {
   ADMINISTRATIVE: `${Application.doctype}/administrative`,
   PHOTOS: `${Application.doctype}/photos`,
   PHOTOS_BACKUP: `${Application.doctype}/photos/mobile`,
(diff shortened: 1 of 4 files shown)


packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,54 @@
+const Application = require('./Application')
+const CozyFile = require('./File')
+
+/**
+ * Class representing the folder model.
+ * @extends CozyFile
+ */
+class CozyFolder extends CozyFile {
+  static async getReferencedFolders(client, reference) {
```
Crash--: No need to pass a `client`. We have access to `this.cozyClient`

Crash--: Can you add a JS Doc please ?

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,54 @@
...
+      .findReferencedBy({
+        _type: Application.doctype,
+        _id: reference
+      })
+    return included.filter(
+      folder => !/^\/\.cozy_trash/.test(folder.attributes.path)
+    )
+  }
+
+  static async getOrCreateFolderWithReference(client, path, reference) {
```
Crash--: Can we have a JS Doc plz ?

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,54 @@
...
+    const existingFolders = await CozyFolder.getReferencedFolders(
+      client,
+      reference
+    )
+
+    if (existingFolders.length) return existingFolders[0]
+
+    const collection = client.collection(CozyFile.doctype)
```
Crash--: I know this is in the Drive code, but we should not use `collection()` since it bypasses all the cozyclient `links`. We should use `find` ?

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,54 @@
...
+      ]
+    )
+
+    const { data: dirInfos } = await collection.get(dirId)
+
+    return dirInfos
+  }
+}
+
+CozyFolder.REF_ADMINISTRATIVE = `${Application.doctype}/administrative`
```
ptbrowne: I suppose those are references kept in a special attribute of the folder, enabling an application to find a specific folder even if it has moved ? Can you add a comment ?

Instead of adding the refs directly at the root of the class, they could be namespaced under `CozyFolder.refs`.

`CozyFolder.refs.PHOTOS`

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,54 @@
...
+CozyFolder.REF_PHOTOS = `${Application.doctype}/photos`
+CozyFolder.REF_UPLOAD = `${Application.doctype}/photos/upload`
+CozyFolder.REF_BACKUP = `${Application.doctype}/photos/mobile`
```
ptbrowne: I personally do not understand why REF_BACKUP is not `${Application.doctype}/backup` ? What is this folder supposed to be ?

gregorylegarec: This is a question for @Crash-- or @y-lohse I guess. This code is directly extracted from Drive.

packages/cozy-doctypes/src/Folder.js
gregorylegarec: Love that, thanks !

packages/cozy-doctypes/src/Folder.js
gregorylegarec: I am not sure using find allows us to use then `ensureDirectoryExists`. Could you confirm ?

packages/cozy-doctypes/src/Folder.js
gregorylegarec: To me CozyFile and Document both are utility classes and are only exposing static methods. I did not find any constructor but I could be missing a point.

packages/cozy-doctypes/src/Folder.js
Crash--: because it has to be specific. We only backup the photos on mobile, so photos/mobile is more pertinent & relevant than /backup. `backup` is not precise enough, it can be the folder backuped by drive desktop or else. 

REF_BACKUP should / can be renamed to REF_PHOTO_BACKUP since we're not in a Drive / Photos context anymore.

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,75 @@
...
+    )
+  }
+
+  /**
+   * Returns an unique folder referenced with the given reference. Creates it
+   * if it does not exist.
+   * @param  {Object}  client    CozyClient instance
+   * @param  {String}  path      Path used to create folder if the referenced
+   * folder does not exist.
+   * @param  {String}  reference
```
Crash--: what is a reference ? :D

packages/cozy-doctypes/src/Folder.js
Crash--: We use https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/Document.js#L72 to register the client in the application.

gregorylegarec: Missed that. Thanks !

packages/cozy-doctypes/src/Folder.js
gregorylegarec: I updated the whole code to distinguish reference to documents and our "Magic folders" logic.

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,105 @@
...
+      _type: Application.doctype,
+      _id: id
+    }
+    const folders = await CozyFolder.getReferencedFolders(magicFolderDocument)
+    const existingMagicFolder = folders.length ? folders[0] : null
+
+    if (existingMagicFolder) return existingMagicFolder
+
+    const magicFoldersValues = Object.values(CozyFolder.refs)
+    if (!magicFoldersValues.includes(id)) {
```
ptbrowne:
```suggestion
    if (!CozyFolder.refs.hasOwnProperty(id)) {
```

This way, there is no need for the intermediate list. Inclusion check is better done with hashmaps than with list in general.

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,105 @@
...
+    return dirInfos
+  }
+
+  /**
+   * Returns a "Magic Folder", given its id
+   * @param  {String} id Magic Folder id. Could be:
+   * * `CozyFolder.refs.ADMINISTRATIVE`
+   * * `CozyFolder.refs.PHOTOS`
+   * * `CozyFolder.refs.PHOTOS_BACKUP`
+   * * `CozyFolder.refs.PHOTOS_UPLOAD`
```
ptbrowne: I think having the list of folders here is not good. It will be out of date since someone will update the CozyFolder.refs and not update this list.

```suggestion
   * CozyFolder.refs contains the ids of folders that can be magic folders.
```

Writing this comment, I was thinking if we'd better not rename `CozyFolder.refs` into `CozyFolders.magicFolders`. It will be clearer and we'll have less risk of someone wanting to store a ref to a document there, *without* wanting it to be a magic folder.

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,105 @@
...
+   * @param  {Object}  document  Document to get references from
+   * @return {Array}             Array of folders referenced with the given
+   * document
+   */
+  static async getReferencedFolders(document) {
+    const { included } = await this.cozyClient
+      .collection(CozyFile.doctype)
+      .findReferencedBy(document)
+    return included.filter(
+      folder => !/^\/\.cozy_trash/.test(folder.attributes.path)
```
ptbrowne: This should be extracted as a method.

```
const isTrashFolder = folder => !/^\/\.cozy_trash/.test(folder.attributes.path)
```

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,105 @@
...
+   */
+  static async getOrCreateFolderWithReference(path, document) {
+    const existingFolders = await CozyFolder.getReferencedFolders(document)
+    if (existingFolders.length) return existingFolders[0]
+    return CozyFolder.createFolderWithReference(path, document)
+  }
+}
+
+/**
+ * Reference used by the Cozy platform and apps for specific folders.
```
ptbrowne:
```suggestion
 * References used by the Cozy platform and apps for specific folders.
```

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Returns an unique folder referenced with the given reference. Creates it
+   * if it does not exist.
+   * @param  {String}  path      Path used to create folder if the referenced
+   * folder does not exist.
+   * @param  {String}  reference
+   * @return {Objet}             Folder referenced with the give reference
+   */
+  static async getOrCreateFolderWithReference(path, document) {
```
ptbrowne: Shouldn't we rename this to `ensureFolderWithReference` to echo the method `ensureMagicFolder` ?

packages/cozy-doctypes/src/Folder.js
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Class representing the folder model.
+ * @extends CozyFile
+ */
+class CozyFolder extends CozyFile {
+  /**
+   * Create a folder with a reference to the given document
+   * @param  {String}  path     Folder path
+   * @param  {Object}  document
```
Crash--: document is a io.cozy.files ?

packages/cozy-doctypes/src/Folder.js
gregorylegarec: I was thinking the same thing, let's go.

packages/cozy-doctypes/src/Folder.js
gregorylegarec: No, I will make the doc clearer about it.

packages/cozy-doctypes/src/Folder.js
gregorylegarec: I could be wrong, but here we are testing values and not keys, so I think `hasOwnProperty` is not appropriate are id is not `'ADMINISTRATIVE'` for example but `CozyFolder.refs.ADMINISTRATIVE` which value is `io.cozy.apps/administrative`.

ptbrowne: Oups, my bad !

gregorylegarec:
@ptbrowne @Crash-- Rebased and ready to merge.

PR #685, opened 2019-07-27
feat: Codemod used to output incorrect usage of global classes in cozy-ui

ptbrowne:
Use to detect cases of global classes used without importing. Should help to resolve
https://github.com/cozy/cozy-ui/issues/1119.


ptbrowne:
Not necessary anymore since utility classes are included in the stylesheet shipped with cozy-ui.

PR #689, opened 2019-07-29
fix(cozy-harvest-lib): Fix OAuth login 

Crash--:
We had a few issues with Oauth konnector : 

- Collection.get do not use as wanted for Trigger since the stack doesn't return `data: {} ` for this end point (PR not finished yet here : https://github.com/cozy/cozy-client/compare/fix/get?expand=1) 
- We should avoid `client.collection` and use `client.query` instead 
- We had a typo in the successCallback

### DIFF
--- a/packages/cozy-harvest-lib/src/connections/accounts.js
+++ b/packages/cozy-harvest-lib/src/connections/accounts.js
@@ -25,8 +25,12 @@ const createAccount = async (client, konnector, attributes) => {
  */
 const findAccount = async (client, id) => {
   try {
-    const { data } = await client.collection(ACCOUNTS_DOCTYPE).get(id)
-    return data
+    const { data } = await client.query(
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/connections/accounts.js
```diff
@@ -25,8 +25,13 @@ const createAccount = async (client, konnector, attributes) => {
...
 const findAccount = async (client, id) => {
   try {
-    const { data } = await client.collection(ACCOUNTS_DOCTYPE).get(id)
-    return data
+    const { data } = await client.query(
+      client.find(ACCOUNTS_DOCTYPE).where({
+        _id: id
+      })
+    )
+    if (data && data[0]) return data[0]
```
edas:
```suggestion
    return data && data[0] || null
```

PR #690, opened 2019-07-30
[Harvest] Use referenced folders

y-lohse:
We want connectors to save their files in the special folders (like `Administrative`), even if these folders are moved or renamed. So this PR uses the special references they have instead hardcoding their path.

This PR will probably fail the CI because of a problem in cozy-logger, I'm looking into it.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -108,14 +108,17 @@ export class TriggerManager extends Component {
     let folder
 
     if (konnectors.needsFolder(konnector)) {
-      const adminFolder = await CozyFolder.ensureMagicFolder(
-        CozyFolder.magicFolders.ADMINISTRATIVE,
-        `/${t('folder.administrative')}`
-      )
-      const photosFolder = await CozyFolder.ensureMagicFolder(
-        CozyFolder.magicFolders.PHOTOS,
-        `/${t('folder.photos')}`
-      )
+      const [adminFolder, photosFolder] = await Promise.all([
+        CozyFolder.ensureMagicFolder(
+          CozyFolder.magicFolders.ADMINISTRATIVE,
+          `/${t('folder.administrative')}`
+        ),
+        CozyFolder.ensureMagicFolder(
+          CozyFolder.magicFolders.PHOTOS,
+          `/${t('folder.photos')}`
+        )
+      ])
+
       const path = konnectors.buildFolderPath(konnector, account, {
         administrative: adminFolder.path,
         photos: photosFolder.path
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -1,7 +1,8 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
-import { withMutations } from 'cozy-client'
+import { withMutations, withClient } from 'cozy-client'
+import { CozyFolder } from 'cozy-doctypes'
 
 import AccountForm from './AccountForm'
 import OAuthForm from './OAuthForm'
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/src/helpers/konnectors.js
```diff
@@ -206,12 +206,10 @@ export const buildFolderPath = (konnector, account, folders = {}) => {
     konnector,
     // For now konnectors are only defining one folder in their folders array
     'folders[0].defaultDir',
-    `$administrative/$konnector/$account`
+    '$administrative/$konnector/$account'
   )
   // Trim `/` and avoid multiple `/` characters with regexp
   let sanitizedPath = trim(fullPath.replace(/(\/+)/g, '/'), '/')
-  //We remove the first / if needed
-  if (sanitizedPath.startsWith('/')) sanitizedPath = sanitizedPath.substring(1)
```
y-lohse: lodash.trim removes leading and trailing `/` just above.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -55,6 +56,15 @@ export class TriggerManager extends Component {
     }
   }
 
+  componentDidMount() {
+    try {
+      CozyFolder.registerClient(this.props.client)
```
ptbrowne: It feels weird to have this here. We should find something better than that IMHO.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -55,6 +56,15 @@ export class TriggerManager extends Component {
...
+    } catch (error) {
+      if (error.message !== 'Document cannot be re-registered to a client.')
```
ptbrowne: This assumes than the client used does not change. It is dangerous IMHO to make this assumption.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -98,11 +108,17 @@ export class TriggerManager extends Component {
...
-      // TODO: remove previous legacy key `default.baseDir`
-      // https://github.com/cozy/cozy-libs/issues/637
+      const adminFolder = await CozyFolder.ensureMagicFolder(
+        CozyFolder.magicFolders.ADMINISTRATIVE,
+        `/${t('folder.administrative')}`
+      )
+      const photosFolder = await CozyFolder.ensureMagicFolder(
+        CozyFolder.magicFolders.PHOTOS,
+        `/${t('folder.photos')}`
+      )
```
Crash--: Can we parallelize these two promises plz?

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
```diff
@@ -7,6 +7,19 @@ import cronHelpers from 'helpers/cron'
 import KonnectorJobWatcher from '../../src/models/konnector/KonnectorJobWatcher'
 import { KonnectorJobError } from 'helpers/konnectors'
 
+jest.mock('cozy-doctypes', () => {
+  return {
+    CozyFolder: {
+      registerClient: () => {},
+      ensureMagicFolder: () => ({ path: '/Administrative' }),
+      magicFolders: {
+        ADMINISTRATIVE: '/adminsitrative',
```
Crash--: Can we have something closer to reality? 

```
magicFolders. ADMINISTRATIVE : 'io.cozy.apps/administrative'
```

Should be easier to understand the difference between this magic folder and the path returned by `ensureMagicFolder`

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
Crash--: Should we have something similar to `cozy-authentication`? https://github.com/cozy/cozy-libs/blob/master/packages/cozy-authentication/src/MobileRouter.jsx#L84 

Cozy-client should expose the list of its events, and we should map on this list to update the client used by cozy-doctypes.

I don't think we currently have a use case where a client can be different in `Document`?

PR #693, opened 2019-08-01
Feat/harvest accounts

Crash--:
KonnectorModal now displays also a `Select` with accounts linked to the konnector. 

I added a few functions to `cozy-doctypes/Account` and their tests from `Home`

Several things to do after : 
- Move MenuFixed to UI (we used it here, but also in `Contacts`) 
- Refactor the logic to fetch the account / triggers 

Capture d’écran 2019-08-02 à 15 49 15

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
@@ -29,7 +29,7 @@ import { Text } from 'cozy-ui/transpiled/react/Text'
 
 import AccountSelectControl from './KonnectorModal/AccountSelectControl'
 import CreateAccountButton from './KonnectorModal/CreateAccountButton'
-
+import { Account } from 'cozy-doctypes'
 const MenuWithFixedComponent = props => {
   const { children } = props
   const { createAction, ...selectProps } = props.selectProps
@@ -68,6 +68,8 @@ export class KonnectorModal extends PureComponent {
    * Next tasks: remove these methods and rewrite the override
    */
   componentDidMount() {
+    const { client } = this.context
+    Account.copyWithClient(client)
     this.fetchAccount(this.props.konnector.triggers.data[0])
     this.fetchAccounts()
   }
@@ -198,12 +200,16 @@ export class KonnectorModal extends PureComponent {
                     this.fetchAccount(option.trigger)
                   }}
                   createAction={createAction}
-                  getOptionLabel={option => option.account.auth.login}
+                  getOptionLabel={option =>
+                    Account.getAccountName(option.account)
+                  }
                   getOptionValue={option => option.trigger._id}
                   defaultValue={accounts[0]}
                   components={{
                     Control: reactSelectControl(
-                      <AccountSelectControl name={account.auth.login} />
+                      <AccountSelectControl
+                        name={Account.getAccountName(account)}
+                      />
                     ),
                     Menu: MenuWithFixedComponent
                   }}
--- a/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
@@ -14,12 +14,13 @@ import { SubTitle } from 'cozy-ui/transpiled/react/Text'
 
 import accountMutations from '../connections/accounts'
 import triggersMutations from '../connections/triggers'
-import * as triggers from '../helpers/triggers'
+import * as triggersModel from '../helpers/triggers'
 import TriggerManager from './TriggerManager'
 import DeleteAccountCard from './cards/DeleteAccountCard'
 import LaunchTriggerCard from './cards/LaunchTriggerCard'
 import TriggerErrorInfo from './infos/TriggerErrorInfo'
 import withLocales from './hoc/withLocales'
+import SelectBox from 'cozy-ui/transpiled/react/SelectBox'
 
 /**
  * KonnectorModal open a Modal related to a given konnector. It fetches the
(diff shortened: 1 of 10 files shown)


packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -10,17 +10,53 @@ import Modal, {
   ModalHeader
 } from 'cozy-ui/transpiled/react/Modal'
 import Spinner from 'cozy-ui/transpiled/react/Spinner'
-import { SubTitle } from 'cozy-ui/transpiled/react/Text'
 
 import accountMutations from '../connections/accounts'
 import triggersMutations from '../connections/triggers'
-import * as triggers from '../helpers/triggers'
+import * as triggersModel from '../helpers/triggers'
```
y-lohse: A bit annoying that this model is not in cozy-doctypes, should we move it there?

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -10,17 +10,53 @@ import Modal, {
...
+import SelectBox, {
+  reactSelectControl,
+  components
+} from 'cozy-ui/transpiled/react/SelectBox'
 
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import { Text } from 'cozy-ui/transpiled/react/Text'
+import { translate } from 'cozy-ui/transpiled/react/I18n'
+
+const MyAccountComponent = ({ name }) => {
```
y-lohse: Need to split those into separate files.

I would also rename this one, maybe to `AccountSelectControl` ?

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -10,17 +10,53 @@ import Modal, {
...
+  return (
+    <Text className="u-slateGrey u-flex u-flex-items-center u-c-pointer">
+      {name} <Icon icon={'bottom-select'} size="12" className="u-ml-half" />
+    </Text>
+  )
+}
+
+const CreateAccount = translate()(({ createAction, t }) => {
```
y-lohse: And this one to `CreateAccountButton`

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -10,17 +10,53 @@ import Modal, {
...
+      size={'small'}
+      className={'u-m-half'}
+      onClick={createAction}
+      label={t('modal.konnector.create_account')}
+    />
+  )
+})
+const MenuWithFixedComponent = props => {
+  const { children } = props
+  const { createAction, ...selectProps } = props.selectProps
```
y-lohse: I would like to move this component to cozy-ui, but not sure how to handle the `createAction` prop. What do you think of having a prop called `fixedMenuProps` on the `Select` that is forwarded to the fixed menu element?

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -141,9 +194,52 @@ export class KonnectorModal extends PureComponent {
...
+        <ModalHeader className="u-pr-2">
+          <div className="u-flex u-flex-row u-w-100 u-flex-items-center">
+            <div className="u-w-3 u-h-3 u-mr-half">
+              <AppIcon fetchIcon={this.fetchIcon} />
+            </div>
+            <div className="u-flex-grow-1 u-mr-half">
+              <h3 className="u-title-h3 u-m-0">{konnector.name}</h3>
+
+              {accounts.length === 0 && (
+                <Text className="u-slateGrey u-flex u-flex-items-center">
```
y-lohse: flex and items-center are not useful here.

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -141,9 +194,52 @@ export class KonnectorModal extends PureComponent {
...
+                  components={{
+                    Control: reactSelectControl(
+                      <MyAccountComponent name={account.auth.login} />
+                    ),
+                    Menu: MenuWithFixedComponent
+                  }}
+                />
+              )}
+            </div>
+            <div className="">
```
y-lohse: className can be removed.

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
```diff
@@ -6,8 +6,8 @@ import Button from 'cozy-ui/transpiled/react/Button'
 import Card from 'cozy-ui/transpiled/react/Card'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
 import Icon from 'cozy-ui/transpiled/react/Icon'
-import { Uppercase, Text } from 'cozy-ui/transpiled/react/Text'
-
+import { Text } from 'cozy-ui/transpiled/react/Text'
+import Uppercase from 'cozy-ui/transpiled/react/Text'
```
y-lohse: I think we can remove those changes.

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
Crash--: I agree with you. Let's do that in an other PR since we'll have comments / refacto to do on the copy pasted code ?

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
Crash--: Removed div since not needed

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
Crash--: Let's create an issue on UI to discuss about the API and component ? :D

PR #697, opened 2019-08-05
fix(cozy-doctypes): Don't use arrow functions in Account model

drazik:
This syntax is only available since node 12. Since we mainly use node
10, this is causing some errors. In banks' services build for example:

```
ERROR in ./node_modules/cozy-doctypes/src/Account.js 14:24
Module parse failed: Unexpected token (14:24)
You may need an appropriate loader to handle this file type, currently
no loaders are configured to process this file. See
https://webpack.js.org/concepts#loaders
|
| class Account extends Document {
  >   static getAccountName = account => {
  |     if (!account) return null
    |     if (account.auth) {
     @ ./node_modules/cozy-doctypes/src/index.js 1:16-36
      @ ./src/targets/services/stats.js
    }
  }
}
```

This is somehow the same problem as we had in
https://github.com/cozy/cozy-libs/pull/557

### DIFF
--- a/packages/cozy-doctypes/src/Account.js
+++ b/packages/cozy-doctypes/src/Account.js
@@ -11,7 +11,7 @@ export const probableLoginFieldNames = [
 ]
 
 class Account extends Document {
-  static getAccountName = account => {
+  static getAccountName(account) {
     if (!account) return null
     if (account.auth) {
       return (
@@ -22,7 +22,7 @@ class Account extends Document {
     }
   }
 
-  static getAccountLogin = account => {
+  static getAccountLogin(account) {
     if (account && account.auth) {
       for (const fieldName of probableLoginFieldNames) {
         if (account.auth[fieldName]) return account.auth[fieldName]


Crash--:
Sorry about that just copy pasted the code from an app. 

Can we add a build step on node 10 for this repo?

drazik:
> Can we add a build step on node 10 for this repo?

Actually tests should already be failing since we use [node 8 in travis](https://github.com/cozy/cozy-libs/blob/master/.travis.yml#L2-L3). But it seems that since we have a babel config, jest is automatically running our tests using babel-jest (see https://jestjs.io/docs/en/getting-started#using-babel).

> Note: babel-jest is automatically installed when installing Jest and will automatically transform files if a babel configuration exists in your project. To avoid this behavior, you can explicitly reset the transform configuration option

It seems we should add `transform: {}` to our jest config, so it doesn't automatically transform the code using babel. I try it.

drazik:
Adding `transform: {}` to our jest config make tests break at a lot of other places. I'm opening another PR related only to that so we don't block this one.

edas:
> Actually tests should already be failing since we use node 8 in travis.

Could we change that ASAP ?

drazik:
> Adding `transform: {}` to our jest config make tests break at a lot of other places. I'm opening another PR related only to that so we don't block this one.

See https://github.com/cozy/cozy-libs/pull/699

> > Actually tests should already be failing since we use node 8 in travis.
>
> Could we change that ASAP ?

See https://github.com/cozy/cozy-libs/pull/700

I'm merging the current one.

PR #703, opened 2019-08-06
[harvest] Added tabs

y-lohse:


packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -68,8 +53,6 @@ export class KonnectorModal extends PureComponent {
    * Next tasks: remove these methods and rewrite the override
    */
   componentDidMount() {
-    const { client } = this.context
-    Account.copyWithClient(client)
```
ptbrowne: Where did that go ? Was it not important ? Also, I am a bit surprised that the result of `Account.copyWithClient(client)` is not used, I thought copyWithClient was supposed to return a copy that had the client in it.

ptbrowne:
LGTM. It seems no tests were changed, I think adding a snapshot test would be good.

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
y-lohse: Yes `copyWithClient` returns a new copy with a client in it, and it wasn't used here. The call to copyWithClient was not useful here, none of the functions of `Account` use the cozy-client.

packages/cozy-harvest-lib/test/components/KonnectorModal.spec.js
```diff
@@ -13,22 +13,51 @@ const findAccountMock = jest.fn().mockImplementation(() => ({
...
+  }
+
+  it('should show a spinner while loading', () => {
+    const component = shallow(<KonnectorModal {...props} />, shallowOptions)
+
+    const spinner = component
+      .dive()
+      .find('ModalContent')
+      .childAt(0)
+      .dive()
```
ptbrowne: Can't we do find(Spinner) directly ? Is it because the shallow render prevents us from doing that ?

y-lohse: Yep, that's why!

PR #705, opened 2019-08-07
refactor(procedures): Parallelize settings' fetch

edas:
Fetch settings in parallel to the transaction fetch in order to avoid
waiting for one round trip


packages/cozy-procedures/src/components/overview/helpers.js
```diff
@@ -62,20 +62,22 @@ export const fetchBanksSettings = async client => {
...
+  const settingsPromise = fetchBanksSettings(client)
   const accounts = await fetchCheckingsLikeAccounts(client)
   const accountsIds = accounts.map(account => account._id)
   const transactions = await fetchTransactionsHistory(client, accountsIds)
-  const settings = await fetchBanksSettings(client)
+  const transactionsByAccount = groupBy(
+    transactions,
+    transaction => transaction.account
+  )
+  const settings = await settingsPromise
```
ptbrowne: IMHO it is a bit surprising to see this await here. I think the parallelizing intention would be clearer with a Promise.all.

```js
const fetchCheckingsLikeTransactions = async client => {
  const accounts = await fetchCheckingsLikeAccounts(client)
  const accountsIds = accounts.map(account => account._id)
  return fetchTransactionsHistory(client, accountsIds)
}

const [transactions, settings] = await Promise.all([
  fetchCheckingsTransactions(client),
  fetchBanksSettings(client),
])
```

edas: I wanted to leave the `groupBy` parallelized but you're right, I might be micro-optimizing and readability is more important.

edas: Is it better for you now?

ptbrowne: :+1:

PR #708, opened 2019-08-09
feat(cozy-app-publish): Add space and application type in mattermost

doubleface:
Now you can get the following messages in mattermost on new publication :

Application __drive__ version 1.2.3 has been published.

or

Connector __cic__ version 1.0.3 has been published on space __selfhosted__

### DIFF
--- a/packages/cozy-app-publish/lib/hooks/post/mattermost.js
+++ b/packages/cozy-app-publish/lib/hooks/post/mattermost.js
@@ -7,9 +7,19 @@ const MATTERMOST_ICON =
   'https://travis-ci.com/images/logos/TravisCI-Mascot-1.png'
 const MATTERMOST_USERNAME = 'Travis'
 
-const sendMattermostReleaseMessage = (appSlug, appVersion) => {
-  const message = `__${appSlug}__ version \`${appVersion}\` has been published.`
-
+const sendMattermostReleaseMessage = (
+  appSlug,
+  appVersion,
+  spaceName,
+  appType
+) => {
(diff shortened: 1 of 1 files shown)


packages/cozy-app-publish/lib/hooks/post/mattermost.js
```diff
@@ -7,9 +7,19 @@ const MATTERMOST_ICON =
...
 const MATTERMOST_USERNAME = 'Travis'
 
-const sendMattermostReleaseMessage = (appSlug, appVersion) => {
-  const message = `__${appSlug}__ version \`${appVersion}\` has been published.`
-
+const sendMattermostReleaseMessage = (
+  appSlug,
+  appVersion,
+  spaceName,
+  appType
```
Crash--: can we default to appType = 'webapp' ? If we don't do it, we'll have an error on `appTypeLabelMap[appType]` if we don't update our current script?

doubleface: If appType is undefined, the label will be ''

```javascript
appTypeLabelMap[appType] || ''
```

And appType is already defaulted to 'webapp' : https://github.com/cozy/cozy-libs/blob/c6b443fd3fa013d209df47ba3ea028d2c7fe7d24/packages/cozy-app-publish/lib/publisher.js#L56

I don't thing we need a default for appType here

Crash--: Woa, didn't see the ` || ''` thanks

PR #709, opened 2019-08-09
feat(cozy-app-publish): Make mattermost channel optional

doubleface:
A default channel can be specified in the mattermost hook.

This is especially useful for connectors when we want to change the channel (which is the same for all connectors) and don't want to change the travis configuration file for all of them.


Crash--:
I think this is a breaking change? Are we sure that all of our Mattermost hooks have a defined channel to post on?

y-lohse:
> Are we sure that all of our Mattermost hooks have a defined channel to post on?

Yes, it's required when setting up a hook on mattermost. i think this change is safe.

PR #710, opened 2019-08-12
[Harvest] Account picker

y-lohse:
Refactors KonnectorModal so that it can switch between 3 states:

- adding an account
- picking one of several existing accounts
- configuring an account

The switch can be handled by the modal itself, or controller by a parent component. This is necessary at the moment because the KonnectorModal is used in different contexts.


packages/cozy-harvest-lib/src/components/AccountsList/AccountsListItem.jsx
```diff
@@ -0,0 +1,66 @@
...
+      return (
+        <div className="u-pomegranate u-flex u-flex-justify-center u-flex-items-center">
+          <span className="u-mr-half u-caption u-pomegranate">
+            {errorTitle}
+          </span>
+          <Icon icon="warning" />
+        </div>
+      )
+    } else {
+      return <Icon icon="check-circleless" color="#2BBA40" />
```
Crash--: I know this is not part of this PR but do we really want to keep an hex color?

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -78,18 +73,73 @@ export class KonnectorModal extends PureComponent {
...
+    this.setState({ fetchingAccounts: true })
+    try {
+      const accounts = (await Promise.all(
+        triggers.map(async trigger => {
+          return {
+            account: await findAccount(triggersModel.getAccountId(trigger)),
+            trigger
+          }
+        })
+      )).filter(({ account }) => !!account)
```
Crash--: why this filter?

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -78,18 +73,73 @@ export class KonnectorModal extends PureComponent {
...
+      this.setState({ accounts, fetchingAccounts: false })
```
Crash--: should we reset the error state?

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -208,117 +224,95 @@ export class KonnectorModal extends PureComponent {
...
-              </Tabs>
-            )}
-          </ModalContent>
-        )}
+        <ModalContent>{this.renderModalContent()}</ModalContent>
       </Modal>
     )
   }
+
+  renderModalContent() {
```
Crash--: I don't like all these `renderXXX` because it means that the component should render multiple things, but as we want to keep the render method as small as possible we create an other render method to do so... 

Maybe we should create a `<ModalContentSwitcher>` component ?

packages/cozy-harvest-lib/src/locales/en.json
```diff
@@ -1,5 +1,4 @@
 {
-  "loading": "Loading",
```
Crash--: should we add `'modal.addAccount.button`?

packages/cozy-harvest-lib/src/components/AccountSelectBox/AccountSelectBox.jsx
```diff
@@ -22,7 +22,13 @@ const AccountSelectBox = ({
       createAction={onCreate}
       getOptionLabel={option => Account.getAccountName(option.account)}
       getOptionValue={option => option.trigger._id}
-      defaultValue={accountList[0]}
+      defaultValue={
+        selectedAccount
+          ? accountList.find(
```
Crash--: We should rename it to `accountsList` :$

packages/cozy-harvest-lib/src/components/AccountsList/AccountsList.jsx
```diff
@@ -0,0 +1,52 @@
...
+              konnector={konnector}
+              trigger={account.trigger}
+              onClick={() => onPick(account)}
+            />
+          </li>
+        ))}
+        <li className="u-mb-half">
+          <Card className="u-p-0">
+            <Button
+              label="Add account"
```
Crash--: translate

packages/cozy-harvest-lib/src/components/AccountsList/AccountsListItem.jsx
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  render() {
+    const { account, onClick } = this.props
+
+    const accountName = Account.getAccountName(account)
+    const accountLogin = Account.getAccountLogin(account)
+    const nameAndLoginDiffer = accountName !== accountLogin
+    const statusContent = this.getStatusContent()
```
Crash--: why do we create a `getStatusContent` method and do not write the method in the render directly?

packages/cozy-harvest-lib/src/locales/en.json
Crash--: and `modal.konnector.create_account` is still used?

packages/cozy-harvest-lib/src/components/AccountsList/AccountsList.jsx
y-lohse: 😅

packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
y-lohse: It is possible that the triggers we receive as props don't have any matching account, in which case we probably want to ignore them.

It happens when we delete an account — the redux store of the Home doesn't know the trigger has been deleted, and will still pass it to the KonnectorModal component. This was a last minute fix, there's probably a better way to handle this.

packages/cozy-harvest-lib/src/components/AccountsList/AccountsListItem.jsx
y-lohse: It was like that before, but there's no good reason for it.

packages/cozy-harvest-lib/src/locales/en.json
y-lohse: `modal.addAccount.button` already exists, and `modal.konnector.create_account` was already removed ;)

PR #713, opened 2019-08-17
fix(cozy-doctypes): Avoid es2015 export syntax

doubleface:
This allows nodejs to use not built version of cozy-doctypes

### DIFF
--- a/packages/cozy-doctypes/.eslintrc
+++ b/packages/cozy-doctypes/.eslintrc
@@ -1,6 +1,9 @@
 {
   "env": { "jest": true },
-  "extends": ["cozy-app"],
+  "extends": ["cozy-app", "plugin:node/recommended"],
+  "rules": {
+    "node/exports-style": ["error", "module.exports"]
+  },
   "globals": {
     "jest": true
   }
--- a/packages/cozy-doctypes/.eslintrc
+++ b/packages/cozy-doctypes/.eslintrc
@@ -2,7 +2,10 @@
   "env": { "jest": true },
   "extends": ["cozy-app", "plugin:node/recommended"],
   "rules": {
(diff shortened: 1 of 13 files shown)


Crash--:
I think I already asked this question, but what is the interest of not using the built version?

ptbrowne:
I think adding https://www.npmjs.com/package/eslint-plugin-node and `"node/exports-style": ["error", "module.exports"],` would be beneficial here as we will forget that we cannot use export const here.

I also wonder why we cannot put the built version as the preferred `main` in package.json.

doubleface:
@Crash-- @ptbrowne I did not do this choice of putting the source version for node, but the only advantage I see is that the async/await stack trace is easier to read.

ptbrowne:
> but the only advantage I see is that the async/await stack trace is easier to read.

It seems a good reason to me, and also we know that we are on nodejs runtime with `main` so there is no need to use transpilation here (apart from es6 modules not being supported). What do you think of th eslint rule ?

ptbrowne:
I've added a bit of doc on environments: https://github.com/cozy/cozy-libs/pull/714

Crash--:
> but the only advantage I see is that the async/await stack trace is easier to read

Good reason to me also! 

> I've added a bit of doc on environments: #714
Thanks! 

> What do you think of th eslint rule ?
Let's go for it! We should also add a rule for the arrow function (https://github.com/mysticatea/eslint-plugin-es/blob/master/docs/rules/no-arrow-functions.md)

ptbrowne:
> We should also add a rule for the arrow function

Why ? Arrow functions are supported in node.

```
$   nvm use 8
$   node
> () => {}
[Function]
```

PR #714, opened 2019-08-19
docs: Add docs on environments

ptbrowne:
Add a bit of doc on how different code is used on browser or node and why.

### DIFF
--- a/packages/cozy-doctypes/README.md
+++ b/packages/cozy-doctypes/README.md
@@ -8,11 +8,9 @@ This repository is meant to bring together methods and functions for doctypes th
 
 ## `Document`
 
-The [Document](./src/Document.js) API is useful in NodeJS contexts where you need to have global models, available across files.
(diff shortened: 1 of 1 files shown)


packages/cozy-doctypes/README.md
```diff
@@ -124,3 +122,19 @@ The created document will have a `cozyMetadata` property like this:
...
+## Environments
+
+The `package.json` has both `main` and `browser` fields. This allows
+
+- a nodeJS runtime to run the non transpiled code (async/await are for example
+supported in nodeJS) which allows for easier to read stacktraces.
+
+- browser runtimes to use transpiled code
+
+The only caveat here is the lack of ES6 module support in nodeJS, we have to use `module.exports` in this lib
```
Crash--: we also have an issue with the arrow function not supported in a node context (see https://github.com/cozy/cozy-libs/commit/25f0e7fed7b260f05f5b5fa3233ea5fca8887f58)

ptbrowne: Ah ok, I understand better, it is not arrow functions that are not supported here, it is the class properties syntax.

Crash--: In my memories it was about arrow function, but after reading the PR you right, it about class properties

PR #716, opened 2019-08-20
chore(eslint-config-cozy-app): Update package.json fields

drazik:
The package has been moved from CPatchane/create-cozy-app to
cozy/cozy-libs but we didn't update the corresponding fields in the
package.json. So renovate PRs links were not correct anymore.


drazik:
Actually there is a redirection, I don't know why I landed on an error page. Anyway, I merge it so there's no confusion.

PR #718, opened 2019-08-20
fix: Status icon sizes

y-lohse:
WIth large content, the icons would shrink because of flexbox rules.


packages/cozy-harvest-lib/src/components/AccountsList/Status.jsx
```diff
@@ -16,11 +16,18 @@ const Status = ({ t, trigger, konnector }) => {
...
-        <Icon icon="warning" />
+        <Icon icon="warning" size={16} className="u-flex-shrink-0" />
       </div>
     )
   } else {
-    return <Icon icon="check" color={palette['malachite']} />
+    return (
+      <Icon
+        icon="check"
+        color={palette['malachite']}
```
gregorylegarec: Didn't knew about this array, it's pretty useful.

PR #723, opened 2019-08-21
feat: Fix split date

ptbrowne:
Transactions older than the split date are ignored as a fail safe
against duplicates when switching from Linxo to BI. Future transactions
should be ignored in the computation of this split date.

### DIFF
--- a/packages/cozy-doctypes/src/banking/BankTransaction.spec.js
+++ b/packages/cozy-doctypes/src/banking/BankTransaction.spec.js
@@ -43,8 +43,23 @@ describe('getIdentifier', () => {
 })
 
 describe('split date', () => {
-  it('should be the oldest date present in the transaction', () => {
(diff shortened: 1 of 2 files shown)


packages/cozy-doctypes/src/banking/BankTransaction.spec.js
```diff
@@ -24,24 +42,14 @@ describe('getIdentifier', () => {
   })
 })
 
+describe('split date', () => {
+  it('should be the oldest date present in the transaction', () => {
```
drazik: When reading "oldest" I thought we would expect the result to be `2018-10-02`, because it's the "lowest" date in the transactions. Should we use another word to make the test name more easy to understand? I think "highest" reflects what we are expecting.

ptbrowne: You're right, I changed it to "most recent"

PR #729, opened 2019-08-22
fix(cozy-harvest-lib): Check if accountLogin is not undefined

Crash--:
Before checking if the accountName differs from the login, we have to check if the login is not undefined ;). It's the case on a few OAuth konnector

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountsList/AccountsListItem.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountsList/AccountsListItem.jsx
@@ -11,10 +11,13 @@ import Status from './Status'
 export class AccountsListItem extends React.PureComponent {
   render() {
     const { account, onClick, konnector, trigger } = this.props
-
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/AccountsList/AccountsListItem.jsx
```diff
@@ -14,7 +14,7 @@ export class AccountsListItem extends React.PureComponent {
 
     const accountName = Account.getAccountName(account)
     const accountLogin = Account.getAccountLogin(account)
-    const nameAndLoginDiffer = accountName !== accountLogin
+    const nameAndLoginDiffer = accountLogin && accountName !== accountLogin
```
y-lohse: I think this was the original code, and I removed the first condition because it didn't seem to make any sense. Can we rename the variable to something like `shouldShowAccountLogin` to show that it's not just about the difference ?

PR #731, opened 2019-08-22
Add konnector new version informatin in Konnector Configuration

gregorylegarec:
This PR get back the information panel to update a konnector.

It's displayed when a konnector has `available_version` attribute to `true`.

I kept the behavior from Home when a `TERMS_VERSION_MISMATCH` error occurs : the Infos panel theme switch to `danger`.

Capture d’écran 2019-08-22 à 12 56 52

Capture d’écran 2019-08-22 à 12 54 37

During tests I encountered issues when dealing with Intents mocking, so I choose to inject them with a `withIntents HOC`. If reviewers agree with its implementation we will then move it to the `cozy-interapp` package, it will just need a little babel configuration. For now it was convenient to have it directly in `cozy-harvest-lib`.

We talked about using AppLinker for redirections via intents, but actually it is `cozy-interapp` responsability so I will update directly the package with this feature.

I look after adding the "available version in account list", but it's missing from the mockup and I don't get the point with indicating in each account that the whole konnector has an available update. By the way, Home do not indicates anything. At least in it current version, I gonna checkout old ones to see if it was previously done.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorConfiguration.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorConfiguration.jsx
@@ -12,12 +12,14 @@ import Icon from 'cozy-ui/transpiled/react/Icon'
 import { SubTitle, Text } from 'cozy-ui/transpiled/react/Text'
 import Button from 'cozy-ui/transpiled/react/Button'
 
+import KonnectorUpdateInfos from '../infos/KonnectorUpdateInfos'
 import TriggerErrorInfo from '../infos/TriggerErrorInfo'
 import TriggerManager from '../TriggerManager'
 import LaunchTriggerCard from '../cards/LaunchTriggerCard'
 import DeleteAccountButton from '../DeleteAccountButton'
 import withLocales from '../hoc/withLocales'
 import * as triggersModel from '../../helpers/triggers'
+import * as konnectorsModel from '../../helpers/konnectors'
 
 class KonnectorConfiguration extends React.Component {
   constructor(props) {
@@ -78,7 +80,10 @@ class KonnectorConfiguration extends React.Component {
     const hasError = !!konnectorJobError
     const shouldDisplayError = !isJobRunning && konnectorJobError
     const hasLoginError = hasError && konnectorJobError.isLoginError()
-    const hasGenericError = hasError && !hasLoginError
+    const hasTermsVersionMismatchError =
+      hasError && konnectorJobError.isTermsVersionMismatchError()
+    const hasGenericError =
+      hasError && !hasLoginError && !hasTermsVersionMismatchError
 
     return (
       <Tabs initialActiveTab={hasLoginError ? 'configuration' : 'data'}>
@@ -100,6 +105,13 @@ class KonnectorConfiguration extends React.Component {
 
         <TabPanels>
           <TabPanel name="data" className="u-pt-1-half u-pb-0">
+            {konnectorsModel.hasNewVersionAvailable(konnector) && (
+              <KonnectorUpdateInfos
+                className="u-mb-2"
+                konnector={konnector}
+                isBlocking={hasTermsVersionMismatchError}
+              />
+            )}
             {shouldDisplayError && hasGenericError && (
               <TriggerErrorInfo
                 className="u-mb-2"
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorConfiguration.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorConfiguration.jsx
@@ -75,16 +75,18 @@ class KonnectorConfiguration extends React.Component {
     } = this.props
     const { isJobRunning, konnectorJobError } = this.state
 
+    const hasError = !!konnectorJobError
     const shouldDisplayError = !isJobRunning && konnectorJobError
-    const hasLoginError = konnectorJobError && konnectorJobError.isLoginError()
-    const hasErrorExceptLogin = konnectorJobError && !hasLoginError
+    const hasLoginError = hasError && konnectorJobError.isLoginError()
+    const hasGenericError = hasError && !hasLoginError
 
     return (
       <Tabs initialActiveTab={hasLoginError ? 'configuration' : 'data'}>
         <TabList>
           <Tab name="data">
             {t('modal.tabs.data')}
-            {hasErrorExceptLogin && (
+            {// Login error should not be mentionned in data tab
+            hasError && !hasLoginError && (
               <Icon icon="warning" size={13} className="u-ml-half" />
             )}
           </Tab>
@@ -98,7 +100,7 @@ class KonnectorConfiguration extends React.Component {
 
         <TabPanels>
           <TabPanel name="data" className="u-pt-1-half u-pb-0">
-            {shouldDisplayError && hasErrorExceptLogin && (
+            {shouldDisplayError && hasGenericError && (
               <TriggerErrorInfo
                 className="u-mb-2"
                 error={konnectorJobError}
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorConfiguration.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorConfiguration.jsx
@@ -82,8 +82,15 @@ class KonnectorConfiguration extends React.Component {
     const hasLoginError = hasError && konnectorJobError.isLoginError()
     const hasTermsVersionMismatchError =
       hasError && konnectorJobError.isTermsVersionMismatchError()
+
+    const isTermsVersionMismatchErrorWithVersionAvailable =
+      hasTermsVersionMismatchError &&
+      konnectorsModel.hasNewVersionAvailable(konnector)
+
     const hasGenericError =
-      hasError && !hasLoginError && !hasTermsVersionMismatchError
+      hasError &&
+      !hasLoginError &&
+      !isTermsVersionMismatchErrorWithVersionAvailable
 
     return (
       <Tabs initialActiveTab={hasLoginError ? 'configuration' : 'data'}>
--- a/packages/cozy-harvest-lib/src/components/KonnectorUpdateLinker.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorUpdateLinker.jsx
@@ -1,0 +1,74 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+
+import { Query } from 'cozy-client'
+import AppLinker from 'cozy-ui/react/AppLinker'
+import Button from 'cozy-ui/react/Button'
+
+import { getStoreUrltoUpdateKonnector } from '../helpers/apps'
+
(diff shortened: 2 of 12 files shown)


packages/cozy-harvest-lib/src/components/hoc/withIntents.jsx
```diff
@@ -0,0 +1,38 @@
...
+ * @return {function} - Component that will receive intents as prop
+ */
+export const withIntents = WrappedComponent => {
+  class ComponentWithIntents extends Component {
+    static intents
+
+    constructor(props, context) {
+      super(props, context)
+      const { client } = props
+      this.intents = this.intents || new Intents({ client })
```
y-lohse: Are you sure this works as intended? Since the class is redefined every time the HOC is used, I would think that `this.intents` is never truthy, but maybe I'm missing something?

packages/cozy-harvest-lib/src/locales/en.json
```diff
@@ -158,6 +158,18 @@
     "administrative": "Administrative",
     "photos": "Photos"
   },
+  "infos": {
```
y-lohse: Nit: I would remove the `infos`, it doesn't really make sense in the context of translations (like other existing translation keys, but might as well not pile them on)

packages/cozy-harvest-lib/test/components/KonnectorUpdateInfos.spec.js
```diff
@@ -0,0 +1,64 @@
...
+      <KonnectorUpdateInfos {...props} isBlocking={true} />
+    ).getElement()
+    expect(component).toMatchSnapshot()
+  })
+
+  it('should render button busy and disabled after click', () => {
+    const wrapper = shallow(<KonnectorUpdateInfos {...props} />)
+    wrapper.instance().handleClick()
+    wrapper.update()
+    expect(wrapper.getElement()).toMatchSnapshot()
```
y-lohse: Not blocking, but I think this sort of test shouldn't be a snapshot. It would be easier to maintain if we tested the busy and disabled props, something like `expect(wrapper.dive().props.busy).toBe(true)`, instead of having to refer to a snapshot in a different file.

packages/cozy-harvest-lib/src/components/hoc/withIntents.jsx
gregorylegarec: Well spotted, I missed the point that the class was redefined every time. This still works here because the only goal of the singleton is to avoid creating a new instance of Intents every time, but as you pointed out, it is eventually what it is done.

What solution do you prefer ?

1) Definining the `ComponentWithIntents` outside the HOC
2) Definining `intents` outside the HOC
3) Another solution I missed.

I prefer 1).

packages/cozy-harvest-lib/src/locales/en.json
gregorylegarec: That's a personal habit and taste, I have been using namespaces in locale keys for years. In the same file there is also `cards` and `errors`. I don't mind to change it, the thing is we really never statued a way to name our locale keys.

packages/cozy-harvest-lib/test/components/KonnectorUpdateInfos.spec.js
gregorylegarec: You're totally right, I plaid guilty.

gregorylegarec:
@y-lohse Yes, to me it's totally the `intents.redirect()` method responsability. I would replace [this line](https://github.com/cozy/cozy-libs/blob/master/packages/cozy-interapp/src/intents.js#L86) with a call to AppLinker

packages/cozy-harvest-lib/src/components/hoc/withIntents.jsx
y-lohse: 1 would be cool but I don't think we can do it since we won't have access to WrappedComponent. But 2 is good as well 👍

packages/cozy-harvest-lib/src/locales/en.json
y-lohse: Yeah fair enough. I just find that when we move the titles etc outside of the `infos` component, suddenly the translation key looks super weird. But it's really a detail, I don't mind keeping things as they are in the PR :)

y-lohse:
> we would now always have our intents.redirect() to be compatible with mobile.

That sounds very compelling, and ok, since intents has a built in concept of redirecting, it would make sense that it's native-app aware. That being said, do you know how you'll avoid duplicating the logic that's already present in AppLinker? Because it's not the easiest stuff and we should really avoid duplicating it…

Crash--:
I'm really wondering if using intent.redirect is the correct answer to our solution. It'll use `window.location.href` in a redirect and we have a few issues on Android with a redirect since Android will check the http response before doing anything else. And it'll not work (cf https://gist.github.com/Crash--/5e080811d24ec3cc6ab81bb2e01758b8#gistcomment-2919298).

I think we should not use intent redirect, but only use `<AppLinker>`.

packages/cozy-harvest-lib/src/locales/en.json
gregorylegarec: Maybe I am a little bit psychorigid here, but in my opinion if the title is move its key must change. In my mind a locale key is also a treasure map about where to find my translation in the app. By the way I could be wrong, I will make a few research about best practices and arguments about locale keys naming conventions :)

packages/cozy-harvest-lib/src/components/infos/TriggerErrorInfo.jsx
```diff
@@ -18,7 +18,9 @@ import Markdown from '../Markdown'
  */
 export class TriggerErrorInfo extends PureComponent {
   render() {
-    const { className, error, konnector, t } = this.props
+    const { className, error, forceUnknown, konnector, t } = this.props
+
+    const displayeError = forceUnknown ? new UnknownKonnectorJobError() || error
```
ptbrowne: Why do we need a new property `forceUnknown` ? Isn't it the responsibility of TriggerInfo to recognize that the error should be unknown ?

ptbrowne: s/displaye/displayed

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorConfiguration.jsx
```diff
@@ -82,8 +82,13 @@ class KonnectorConfiguration extends React.Component {
     const hasLoginError = hasError && konnectorJobError.isLoginError()
     const hasTermsVersionMismatchError =
       hasError && konnectorJobError.isTermsVersionMismatchError()
+
+    const hasConsistentTermsVersionMismatchError =
+      hasTermsVersionMismatchError &&
+      konnectorsModel.hasNewVersionAvailable(konnector)
```
ptbrowne: Why do we need to do that ? Is it related to banking connectors ?

packages/cozy-harvest-lib/src/components/infos/TriggerErrorInfo.jsx
gregorylegarec: The case were we have a `TERMS_MISMATCH_ERROR` for a konnector with no update available is not supposed to happen. BUT as we rely on informations from two sources (error from trigger and available_version from konnector), I think we must care about the case when only the TERMS_MISMATCH_ERROR is present.

I talk with @poupotte about this point, and the expected behaviour should be to get an Unknown error displayed.

Now that you're pointing this out, maybe actually we should deal in another way, for example display a specific message for `TERMS_MISMATCH_ERROR`.

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorConfiguration.jsx
gregorylegarec: It has been used for banking konnector, but it is not specific to them. The case when this happens is explained in [the comment above](https://github.com/cozy/cozy-libs/pull/731#discussion_r318661364)

gregorylegarec:
In the last commits have been added :

Update message in account list
Capture d’écran 2019-08-26 à 19 56 53

Dedicated message for `TERMS_MISMATCH_ERROR`
Capture d’écran 2019-08-29 à 19 17 30

packages/cozy-harvest-lib/src/components/KonnectorUpdateLinker.jsx
```diff
@@ -0,0 +1,83 @@
...
+    onClick={onClick}
+    theme={isBlocking ? 'danger' : 'secondary'}
+  />
+)
+
+export class KonnectorUpdateLinker extends PureComponent {
+  render() {
+    const { client, label, isBlocking, konnector } = this.props
+    return (
+      <Query query={client.find('io.cozy.apps')}>
```
Crash--: Can we add a `where` type=== konnector is this request?

packages/cozy-harvest-lib/src/components/KonnectorUpdateLinker.jsx
```diff
@@ -0,0 +1,83 @@
...
+        {({ data, fetchStatus }) => {
+          const isFetching = fetchStatus !== 'loaded'
```
Crash--: wrong. If fetchStatus === 'errored' we are not fetching and I want to display an error message instead of a loader

packages/cozy-harvest-lib/src/components/KonnectorUpdateLinker.jsx
```diff
@@ -0,0 +1,83 @@
...
+                )}
+              </AppLinker>
+            )
+          } else {
+            return (
+              <KonnectorUpdateButton
+                disabled={true}
+                label={label}
+                isBlocking={isBlocking}
+              />
```
Crash--: I don't like this `KonnectorUpdateButton` abstraction. Adding it and adding this "big" if just to deal with 2 props don't sound good. I think we can do something like : 
```jsx
<KonnectorUpdateButton
{...isReady ? {href, onClick} : {disabled: true}} 
```

packages/cozy-harvest-lib/src/components/KonnectorUpdateLinker.jsx
```diff
@@ -0,0 +1,83 @@
...
+        }}
+      </Query>
+    )
+  }
+}
+
+KonnectorUpdateLinker.propTypes = {
+  isBlocking: PropTypes.bool,
+  konnector: PropTypes.object.isRequired,
+  label: PropTypes.string
```
Crash--:
```suggestion
  label: PropTypes.string,
  client: PropTypes.object.isRequired
```

packages/cozy-harvest-lib/src/components/infos/KonnectorUpdateInfos.jsx
```diff
@@ -0,0 +1,44 @@
...
+        />
+      }
+      className={className}
+      isImportant={isBlocking}
+      text={
+        isBlocking
+          ? t('infos.konnectorUpdate.body.blocking')
+          : t('infos.konnectorUpdate.body.regular')
+      }
+      title={t('infos.konnectorUpdate.title')}
```
Crash--: Should we add an `icon` warning if `isBlocking`?

PR #736, opened 2019-08-26
[Harvest] refactor: Deduplicate trigger launch logic

y-lohse:
The code to launch a trigger was more or less duplicated, once in TriggerManager and once in TriggerLauncher. This made the data flow in Harvest more complicated than it needed to be (a lot of prop drilling and callback chains to keep a single source of  truth).

This PR removes the duplicated code from TriggerManager and uses TriggerLauncher instead. We also added support for an `onLoginSuccess` callback in TriggerLauncher.

We tried to keep the breaking changes to a minimum, but there is one. TriggerLauncher now needs to work without an initial trigger, since it's also used when we are adding a new account. To accomodate this:

- The `trigger` prop has been renamed to `initialTrigger`
- The `launch` prop that TriggerLauncher passes to it's children must be called with a `trigger` parameter

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
@@ -7,10 +7,12 @@ import TwoFAModal from './TwoFAModal'
 import { accountsMutations } from '../connections/accounts'
 import { triggersMutations } from '../connections/triggers'
 import withLocales from './hoc/withLocales'
+import triggers from '../helpers/triggers'
 
 import KonnectorJob, {
   ERROR_EVENT,
   SUCCESS_EVENT,
+  LOGIN_SUCCESS_EVENT,
   TWO_FA_REQUEST_EVENT,
   TWO_FA_MISMATCH_EVENT
 } from '../models/KonnectorJob'
@@ -22,9 +24,9 @@ import KonnectorJob, {
  *
  * ### Example
  * ```js
- * <TriggerLauncher trigger={trigger}>
- *   {(launch, running) => (
- *     <Button onClick={launch} disabled={running}
+ * <TriggerLauncher initialTrigger={trigger}>
+ *   {(launch, running, trigger) => (
+ *     <Button onClick={() => launch(trigger)} disabled={running} />
  *    )}
  * </TriggerLauncher>
  * ```
@@ -36,29 +38,36 @@ import KonnectorJob, {
 export class TriggerLauncher extends Component {
   constructor(props, context) {
     super(props, context)
-    const { trigger } = this.props
-    this.state = { showTwoFAModal: false, trigger }
+    const { initialTrigger } = this.props
+    this.state = {
+      showTwoFAModal: false,
+      trigger: initialTrigger,
+      error: triggers.getError(initialTrigger)
+    }
 
     this.handleTwoFA = this.handleTwoFA.bind(this)
     this.handleError = this.handleError.bind(this)
     this.handleSuccess = this.handleSuccess.bind(this)
+    this.handleLoginSuccess = this.handleLoginSuccess.bind(this)
 
     this.launch = this.launch.bind(this)
   }
 
-  launch() {
-    const { client, onLaunch, trigger } = this.props
+  launch(trigger) {
+    const { client, onLaunch } = this.props
     const konnectorJob = new KonnectorJob(client, trigger)
     konnectorJob
       .on(ERROR_EVENT, this.handleError)
       .on(SUCCESS_EVENT, this.handleSuccess)
+      .on(LOGIN_SUCCESS_EVENT, this.handleLoginSuccess)
       .on(TWO_FA_REQUEST_EVENT, this.handleTwoFA)
       .on(TWO_FA_MISMATCH_EVENT, this.handleTwoFA)
 
     this.setState({
       error: null,
       konnectorJob,
-      running: true
+      running: true,
+      trigger
     })
 
     if (typeof onLaunch === 'function') onLaunch(trigger)
@@ -93,12 +102,19 @@ export class TriggerLauncher extends Component {
     if (typeof onSuccess === 'function') onSuccess(trigger)
   }
 
+  async handleLoginSuccess() {
+    this.dismissTwoFAModal()
+    const { onLoginSuccess } = this.props
+    if (typeof onLoginSuccess === 'function') onLoginSuccess()
+  }
+
   handleTwoFA() {
     this.displayTwoFAModal()
   }
 
   async refetchTrigger() {
-    const { fetchTrigger, trigger } = this.props
+    const { fetchTrigger } = this.props
+    const { trigger } = this.state
     try {
       return await fetchTrigger(trigger._id)
     } catch (error) {
@@ -153,17 +169,22 @@ TriggerLauncher.propTypes = {
    */
   onSuccess: PropTypes.func,
   /**
+   * Callback to call when the konnector is logged in
+   */
+  onLoginSuccess: PropTypes.func,
+  /**
    * Callback to call when the konnector job fails
    */
   onError: PropTypes.func,
   /**
-   * Indicates if trigger is already runnning
+   * Indicates if trigger is already runnning.
+   * Useful at the initial render of the page when we know a job is already running.
    */
   submitting: PropTypes.bool,
   /**
-   * Trigger to launch
+   * Optionnal trigger in its initial state
    */
-  trigger: PropTypes.object.isRequired
+  initialTrigger: PropTypes.object
 }
 
 export default withLocales(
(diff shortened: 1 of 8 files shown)


packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -93,12 +102,19 @@ export class TriggerLauncher extends Component {
     if (typeof onSuccess === 'function') onSuccess(trigger)
   }
 
+  async handleLoginSuccess() {
+    this.dismissTwoFAModal()
```
Crash--: should we check if we have this modal displayed before dismissing it? It could save one render call

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -22,9 +24,9 @@ import KonnectorJob, {
  *
  * ### Example
  * ```js
- * <TriggerLauncher trigger={trigger}>
- *   {(launch, running) => (
- *     <Button onClick={launch} disabled={running}
+ * <TriggerLauncher initialTrigger={trigger}>
+ *   {(launch, running, trigger) => (
+ *     <Button onClick={() => launch(trigger)} disabled={running}
```
Crash--:
```suggestion
 *     <Button onClick={() => launch(trigger)} disabled={running} />
```

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -22,9 +24,9 @@ import KonnectorJob, {
  *
  * ### Example
  * ```js
- * <TriggerLauncher trigger={trigger}>
- *   {(launch, running) => (
- *     <Button onClick={launch} disabled={running}
+ * <TriggerLauncher initialTrigger={trigger}>
+ *   {(launch, running, trigger) => (
```
Crash--: Here, the trigger is the updated one?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -32,59 +30,30 @@ const MODAL_PLACE_ID = 'coz-harvest-modal-place'
...
-    if (this.jobWatcher) this.jobWatcher.handleSuccess()
-  }
-
   /**
    * Ensure that a trigger will exist, with valid destination folder with
    * permissions and references
    * @return {Object} Trigger document
    */
-  async ensureTrigger() {
+  async ensureTrigger(account) {
```
Crash--: update jsdoc with the new arg

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -144,65 +113,14 @@ export class TriggerManager extends Component {
   async handleOAuthAccountId(accountId) {
     const { findAccount } = this.props
     try {
+      this.setState({ error: null, status: RUNNING })
```
Crash--: are we sure to want to unset the error state? Can't we have something inside?

packages/cozy-harvest-lib/test/components/TriggerLauncher.spec.js
```diff
@@ -239,14 +251,43 @@ describe('TriggerLauncher', () => {
...
+        <TriggerLauncher {...props} onLoginSuccess={onLoginSuccess}>
+          {({ launch, running }) => <Child launch={launch} running={running} />}
+        </TriggerLauncher>
+      )
+
+      wrapper.instance().launch(trigger)
+      wrapper.instance().handleLoginSuccess()
+      expect(onLoginSuccess).toHaveBeenCalled()
+    })
+  })
```
Crash--: since `TriggerLauncher` is now retuning also a `trigger` should we add a test for it?

Crash--: Also, since we don't call `handleLoginSuccess` by hand in our code, can we add a test for our real case? ie dispatching the event and check if `handleLoginSuccess` is called or something similar?

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
y-lohse: Yes.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
gregorylegarec: At this point we received an account id from stack and we are going to run, or re-run the konnector. We do not want to show an error anymore in the TriggerManager, if any.

packages/cozy-harvest-lib/test/components/TriggerLauncher.spec.js
gregorylegarec: It looks like testing the event dispatching library we are using here, MicroEE. Maybe we should test KonnectorJob instead.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
y-lohse: IF we had something in this error state, it would be from the previous attempt to connect the account. Since a new attempt just started, yes, we really want to unset it :)

packages/cozy-harvest-lib/test/components/TriggerLauncher.spec.js
Crash--: This is not the implementation of the event I want to test. I just want to test, that if the flow goes right, we call `handleLoginSuccess` automatically. So yeah, it's basically KonnectorJob

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
Crash--: Thanks

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
y-lohse: I said 👍 but doe sit actually save a render call ? If the state is the same I don't think it will trigger an update ?

Crash--: It will call the render method and execute it. But it'll not re-render the dom if nothing change (but will call chidren render)

ptbrowne: To build on @Crash-- answer, TriggerLauncher is not a PureComponent so it does not have the shouldComponentUpdate check on props/state that would bypass the render if state/props had not changed.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -16,75 +15,45 @@ import cron from '../helpers/cron'
...
 const IDLE = 'IDLE'
 const RUNNING = 'RUNNING'
-const RUNNING_TWOFA = 'RUNNING_TWOFA'
 
 const MODAL_PLACE_ID = 'coz-harvest-modal-place'
 
 /**
- * Manage a trigger for a given konnector, from account edition to trigger
- * creation and launch.
+ * Displays the login form and on submission will create the account, triggers and folders. After that it calls TriggerLauncher to run the konnector.
```
ptbrowne:
```suggestion
 * Displays the login form and on submission will create the account, triggers and folders.
 * After that it calls TriggerLauncher to run the konnector.
```

ptbrowne:
I am not knowledgeable about this bit of code but it seems that business logic is done by TriggerLauncher, as per its docstring state that it creates the account/triggers. It seems an anti-pattern to me to do that. To me, business logic should be done in non React aware code. What do you think ? I do not want to block this PR anyway, we could discuss in an issue if this sparks more discussion.

Crash--:
You right @ptbrowne (I think you mean TriggerManager instead of TriggerLauncher, right?).

We refactored a lot harvest this last few days and we want to keep it (ATM) as much backward compatible than we can. The goal is to try to isolate each part of the process without rewriting the all thing (as we're not yet aware of all the case we have to deal with). 

We only start to have a good vision of the role of harvest and how we can enhance it. I think we can write some documentation to explain that in order to share what we found out and what we think for the future.

gregorylegarec:
@ptbrowne That was already a discussion when we first merged the TriggerManager. Clearly there is too much business logic in it. What we actually need is almost only the `ensureTrigger` method. This method should be provided by a dedicated object. We talked about this with @y-lohse and where wondering how to name it : TriggerEnsurer ? TriggerBuilder ?

ptbrowne:
> We only start to have a good vision of the role of harvest and how we can enhance it.

:+1:

> I think we can write some documentation to explain that in order to share what we found out and what we think for the future.

I think documentation might not be the right place, I think directly writing this in the docstrings of the components is better as the next developer will this directly.

```
TODO it seems that TriggerLauncher is doing too much business logic, we should try to split it.
```

ptbrowne:
> @ptbrowne That was already a discussion when we first merged the TriggerManager. Clearly there is too much business logic in it. What we actually need is almost only the ensureTrigger method. This method should be provided by a dedicated object. We talked about this with @y-lohse and where wondering how to name it : TriggerEnsurer ? TriggerBuilder ?

I don't see why we would need a `class` or a dedicated object here. It seems we already have helpers methods dedicated to triggers in `helpers/triggers`, I think the ensureTrigger method could live there provided it takes `cozyClient`, `account` and `trigger` as arguments.

y-lohse:
Merging this, another PR incoming to remove the extra `dismiss2FA` and fixing some other issues with 2FA.

PR #742, opened 2019-08-28
Harvest: Make 2FA work again

gregorylegarec:
This PR follows #736 and fixes a few issues.

We peered with @y-lohse and made also a few changes/refactoring to improve code quality. There is still a little work to simplify the whole process.

Plus a little cleaning and refactoring

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
@@ -46,7 +46,6 @@ export class TriggerLauncher extends Component {
     }
 
     this.handleTwoFA = this.handleTwoFA.bind(this)
-    this.handleTwoFACode = this.handleTwoFACode.bind(this)
     this.handleError = this.handleError.bind(this)
     this.handleSuccess = this.handleSuccess.bind(this)
     this.handleLoginSuccess = this.handleLoginSuccess.bind(this)
@@ -115,12 +114,6 @@ export class TriggerLauncher extends Component {
   handleTwoFA() {
     this.displayTwoFAModal()
   }
-
-  async handleTwoFACode(code) {
-    this.setState({ isSubmittingTwoFA: true })
-    return this.konnectorJob.sendTwoFACode(code)
-  }
-
   async refetchTrigger() {
     const { fetchTrigger } = this.props
     const { trigger } = this.state
@@ -152,11 +145,7 @@ export class TriggerLauncher extends Component {
           <TwoFAModal
             dismissAction={this.dismissTwoFAModal}
             into="coz-harvest-modal-place"
-            konnectorSlug={triggers.getKonnectorSlug(trigger)}
-            hasFailed={this.konnectorJob.isTwoFARetry()}
-            isSubmitting={this.konnectorJob.isTwoFARunning()}
-            onTwoFACodeSubmit={this.handleTwoFACode}
-            provider={this.konnectorJob.getTwoFACodeProvider()}
+            konnectorJob={this.konnectorJob}
           />
         )}
       </div>
--- a/packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
@@ -13,8 +13,7 @@ import KonnectorJob, {
   ERROR_EVENT,
   SUCCESS_EVENT,
   LOGIN_SUCCESS_EVENT,
-  TWO_FA_REQUEST_EVENT,
-  TWO_FA_MISMATCH_EVENT
+  TWO_FA_REQUEST_EVENT
 } from '../models/KonnectorJob'
 
 /**
@@ -45,7 +44,8 @@ export class TriggerLauncher extends Component {
       error: triggers.getError(initialTrigger)
     }
 
-    this.handleTwoFA = this.handleTwoFA.bind(this)
+    this.dismissTwoFAModal = this.dismissTwoFAModal.bind(this)
+    this.displayTwoFAModal = this.displayTwoFAModal.bind(this)
     this.handleError = this.handleError.bind(this)
     this.handleSuccess = this.handleSuccess.bind(this)
     this.handleLoginSuccess = this.handleLoginSuccess.bind(this)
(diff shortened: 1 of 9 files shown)


packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
```diff
@@ -28,20 +30,27 @@ export class TwoFAForm extends PureComponent {
...
 
   fetchIcon() {
     const { client } = this.context
-    const { konnectorJob } = this.props
+    const { konnectorSlug } = this.props
     return client.stackClient.getIconURL({
       type: 'konnector',
-      slug: konnectorJob.getKonnectorSlug()
+      slug: konnectorSlug
     })
```
ptbrowne: When the TwoFaModal is displayed, we already have the konnector icon displayed in the modal behind. So I think we could take the icon from there and display it directly in TwoFAModal. Technically, it would be better IMHO because it would remove this logic from the TwoFAModal and save a round trip to the server to fetch the icon. Functionally, it would also be better since the icon would not be reloaded : I already have seen the loading animation of the icon in the modal and tought "why is it loading ? we already have it behind" ?

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -141,9 +151,13 @@ export class TriggerLauncher extends Component {
         })}
         {showTwoFAModal && (
           <TwoFAModal
-            konnectorJob={konnectorJob}
             dismissAction={this.dismissTwoFAModal}
             into="coz-harvest-modal-place"
+            konnectorSlug={triggersModel.getKonnectorSlug(trigger)}
```
ptbrowne: I think the name triggersModel is deceptive since we do not have an instance of a model here. Maybe triggerLib ? triggers ?

packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
```diff
@@ -91,8 +100,14 @@ export class TwoFAForm extends PureComponent {
...
-TwoFAForm.contextTypes = {
-  client: PropTypes.object
+TwoFAModal.propTypes = {
+  client: PropTypes.object,
+  konnectorSlug: PropTypes.string.isRequired,
+  hasFailed: PropTypes.bool,
+  isSubmitting: PropTypes.bool,
+  onTwoFACodeSubmit: PropTypes.func.isRequired,
+  /** Provider between default, sms or email */
+  provider: PropTypes.string.isRequired
```
ptbrowne: PropTypes.oneOf(['sms', 'email']).isRequired ?

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -141,9 +151,13 @@ export class TriggerLauncher extends Component {
...
+            hasFailed={this.konnectorJob.isTwoFARetry()}
+            isSubmitting={this.konnectorJob.isTwoFARunning()}
+            onTwoFACodeSubmit={this.handleTwoFACode}
+            provider={this.konnectorJob.getTwoFACodeProvider()}
```
ptbrowne: Why did we remove konnectorJob as prop and created new props ? It seems that by doing so, we introduce complexity and a new vocabulary that might not be necessary. If I look at the code of TwoFAModal I see `hasFailed` and I think what does `hasFailed` mean, and I must come here to see that `isTwoFaRetry` has been transformed into `hasFailed` ? Why introducing a layer of translation here ?

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -112,6 +116,11 @@ export class TriggerLauncher extends Component {
     this.displayTwoFAModal()
   }
 
+  async handleTwoFACode(code) {
+    this.setState({ isSubmittingTwoFA: true })
```
ptbrowne: It seems to me that we are duplicating the submittingTwoFa state, at some place we use isTwoFaRunning() and at another place we use `isSubmittingTwoFa` here ?
BTW is submitting and running the same thing ? If it is, let's use the same word, it will be clearer.

ptbrowne: It seeems `isSubmittingTwoFA` is not used here and only used to trigger a re-render.

Maybe we need an observable on konnectorJob ? `konnectorJob.on(change, this.forceUpdate)` ?

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
ptbrowne: Talking with Yannick, I understand that the need is to prevent TwoFAModal to call methods on konnectorJob. An alternative way would be to have a `.freeze()`, `.dumb()` method on konnectorJob that would pass a dumbed down version of konnectorJob (without the methods) to TwoFAModal.

My impression here and what I want to prevent is : I do not want, in the event of an evolution of konnectorJob or of the needs of TwoFAModal to have to change TriggerLauncher. I just want to chagne the code of TwoFAModal or konnectorJob.

packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
y-lohse: I agree. @Crash-- do you think we could integrate this change in our routing PR? We already have a component that takes care of rendering the konnector icon there, and I believe it only fetches the icon once.

Crash--: I think we can, let's create an issue. I don't want to add anything else in the upcoming PR. ATM we're unable to finish it already ^^

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
gregorylegarec: I used to use triggers 👍

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
gregorylegarec: Well spotted @ptbrowne !

Actually the best solution is to pass the konnectorJob and subscribe to its own event mechanism. It is an evidence for us today but yesterday we were a little bit both tired at the end of the day.

PR #744, opened 2019-08-29
fix(cozy-doctypes): Check that account number exist before matching it

drazik:
We have a case in which the account's `number` property is undefined on production, so we must handle this case.


ptbrowne:
Can you add a test ? 👼

drazik:
I am trying to see why this user has no number on an account so I can add a real useful test ;)

PR #750, opened 2019-08-30
feat: Deals with accounts edition

edas:
- if a cipher is already linked to this account, try to use it
- otherwise, try to get a cipher with the same username and password
which match the requested uri
- otherwise, try to get the first cipher matching username and uri
- otherwise create a new cipher

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -149,7 +149,7 @@ export class TriggerManager extends Component {
       let originalCipher = null
       if (isUpdate && konnectorURI) {
         const search = { username: login, password, uri: konnectorURI }
-        originalCipher = await accounts.findCipher(account, vaultClient, search)
+        originalCipher = await accounts.getVaultCipher(account, vaultClient, search)
       }
 
       let cipherData
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -5,7 +5,7 @@ import flow from 'lodash/flow'
 
 import { withMutations, withClient } from 'cozy-client'
 import { CozyFolder as CozyFolderClass } from 'cozy-doctypes'
-import { VaultUnlocker, withVaultClient } from 'cozy-keys-lib'
+import { VaultUnlocker, withVaultClient, CipherType } from 'cozy-keys-lib'
 
 import AccountForm from './AccountForm'
 import OAuthForm from './OAuthForm'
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -94,6 +96,55 @@ export const mergeAuth = (account, authData) => ({
   auth: merge(account.auth, authData)
 })
 
+/**
+ * Gets the vault cipher relationship to an account
+ * @param  {object} account   io.cozy.accounts document
+ * @return {string}           cipher uuid
+ */
+export const getVaultCipherId = account =>
+  account.relationships &&
```
y-lohse: lodash.get is pretty good for those cases: `get(account, 'relationships.vaultCipher._id')`

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -94,6 +96,55 @@ export const mergeAuth = (account, authData) => ({
...
+  account.relationships.vaultCipher &&
+  account.relationships.vaultCipher._id
+
+/**
+ * Gets the linked cipher in vault
+ * If none is linked, search on that match
```
y-lohse: This function does 2 pretty different things, do you think it would make sense to split it into 2 separate functions?

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -94,6 +96,55 @@ export const mergeAuth = (account, authData) => ({
...
+    // there is an explicit existing link
+    const cipher = await vaultClient.get(id)
+    if (cipher) return cipher
+  }
+  if (username && uri) {
+    // there is no link or the link is broken
+    // but there may be a existing cipher with the same credentials
+    const type = CipherType.Login
+    const search = { type, username, uri }
+    const all = vaultClient.getAllDecryptedFor(search)
```
y-lohse: `getAllDecryptedFor ` isn't async?

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -94,6 +96,55 @@ export const mergeAuth = (account, authData) => ({
...
+    if (all.length > 0) {
+      const all = all.find(v => v.login.password == password)
```
y-lohse: That's a second constant called `all`, doesn't this throw an error?

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -94,6 +96,55 @@ export const mergeAuth = (account, authData) => ({
...
+ * If none is linked, search on that match
+ * @param account - io.cozy.accounts document
+ * @param {WebVaultClient} vaultClient
+ * @param {object} search criterias
+ * @param {string} search.username
+ * @param {string} search.password
+ * @param {string} search.uri
+ * @return {CipherView|null}
+ */
+export const getVaultCipher = async (
```
y-lohse: I'm not sure this is the right file for this function. This is essentially a Model file for `io.cozy.accounts`, I don't think Vault helper functions should be here.

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -94,6 +96,55 @@ export const mergeAuth = (account, authData) => ({
...
+  account,
+  vaultClient,
+  { username, password, uri } = {}
+) => {
+  const id = getVaultCipherId(account)
+  if (id) {
+    // there is an explicit existing link
+    const cipher = await vaultClient.get(id)
+    if (cipher) return cipher
```
y-lohse: If teh cipher can't be fond, shoud we re-attach the account to the new cipher?

edas: I think so

packages/cozy-harvest-lib/src/helpers/accounts.js
edas: You could see that as an helper for ciphers (it fetches ciphers) or an helper for accounts (it uses an account as first parameter and use internal data from account, public functions from ciphers). Both ways are fine with me.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -141,18 +141,40 @@ export class TriggerManager extends Component {
...
+      if (isUpdate) {
+        const id = accounts.getVaultCipherId(account)
+        const search = {
           username: login,
-          password,
-          uris: konnectorURI ? [{ uri: konnectorURI, match: 0 }] : []
+          uri: konnectorURI,
+          type: CipherType.Login
         }
+        const sort = [view => view.login.password === password, 'revisionDate']
```
y-lohse: Nit: this looks a bit strange for a sort, maybe worth a comment to explain what we're doing.

PR #776, opened 2019-09-09
feat: Add methods to Application Doctype

Crash--:
Rather than having to namespace those methods, let's use our Doctype ;)

### DIFF
--- a/packages/cozy-doctypes/src/Application.js
+++ b/packages/cozy-doctypes/src/Application.js
@@ -1,8 +1,51 @@
 const Document = require('./Document')
 
 const APP_DOCTYPE = 'io.cozy.apps'
+const STORE_SLUG = 'store'
 
-class Application extends Document {}
+class Application extends Document {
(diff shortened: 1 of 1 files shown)


packages/cozy-doctypes/src/Application.js
```diff
@@ -1,8 +1,51 @@
...
+  /**
+   * Return Store URL where an app/konnector can be installed / updated
+   * @param  {Array}  [appData=[]]   Apps data, as returned by endpoint /apps/ or
+   * /konnectors/
+   * @param  {Object} [app={}] AppObject
+   * @return {String}                URL as string
+   */
+  static getStoreInstallationURL(appData = [], app = {}) {
+    if (!app.slug) {
+      throw new Error('Expected app / konnector withe defined slug')
```
drazik:
```suggestion
      throw new Error('Expected app / konnector with the defined slug')
```

PR #781, opened 2019-09-11
[harvest] Dismiss vault unlock screen

y-lohse:
So far, it wasn't possible to go away from the unlock screen.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
@@ -30,9 +30,7 @@ export class EditAccountModal extends Component {
       error: false
     }
 
-    this.handleAccountEditJobSuccess = this.handleAccountEditJobSuccess.bind(
-      this
-    )
+    this.redirectToAccount = this.redirectToAccount.bind(this)
   }
 
   componentDidMount() {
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -227,7 +227,8 @@ export class TriggerManager extends Component {
       konnector,
       running: triggerRunning,
       showError,
-      modalContainerId
+      modalContainerId,
+      onVaultDismiss
```
Crash--: can you add it to the propTypes ? Is it mandatory ?

packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
```diff
@@ -131,7 +129,8 @@ export class EditAccountModal extends Component {
               account={account}
               konnector={konnector}
               initialTrigger={trigger}
-              onSuccess={this.handleAccountEditJobSuccess}
+              onSuccess={this.redirectToAccount}
+              onVaultDismiss={this.redirectToAccount}
```
Crash--: I know this is not the goal of this PR but I'm wondering if we can move all the `keys` related stuffs behind a flag?

y-lohse: I think we need to do this one of these days yes. @edas what do you think? You seemed to say it wasn't too high-priority?

PR #782, opened 2019-09-12
[Harvest] Handle cipher selection

drazik:
When the user clicks on a cipher, it is used in the account form.

There are two cases:

* If all required fields are present in the cipher, then the account form is automatically submitted
* Otherwise, the user can edit the missing fields and submit the form manually

### DIFF
--- a/packages/cozy-doctypes/src/Account.js
+++ b/packages/cozy-doctypes/src/Account.js
@@ -57,7 +57,14 @@ class Account extends Document {
           ...customFields
         },
         value => Boolean(value)
-      )
+      ),
+      relationships: {
+        vaultCipher: {
+          _id: cipher.id,
+          _type: 'com.bitwarden.ciphers',
+          _protocol: 'bitwarden'
+        }
+      }
     }
   }
 }
--- a/packages/cozy-doctypes/src/Account.js
+++ b/packages/cozy-doctypes/src/Account.js
@@ -1,4 +1,5 @@
 const Document = require('./Document')
+const pickBy = require('lodash/pickBy')
 
 const ACCOUNTS_DOCTYPE = 'io.cozy.accounts'
 
@@ -27,6 +28,36 @@ class Account extends Document {
       for (const fieldName of probableLoginFieldNames) {
         if (account.auth[fieldName]) return account.auth[fieldName]
       }
+    }
+  }
+
+  /**
+   * Create an account document from a vault cipher
+   *
+   * @param {Object} cipher
+   *
+   * @returns {Object}
+   */
+  static fromCipher(cipher) {
+    if (!cipher) {
+      throw new Error('Cipher is required')
+    }
+
+    const customFields = (cipher.fields || []).reduce((fields, field) => {
+      fields[field.name] = field.value
+
+      return fields
+    }, {})
+
+    return {
+      auth: pickBy(
+        {
+          login: cipher.login.username,
+          password: cipher.login.password,
+          ...customFields
+        },
+        value => Boolean(value)
+      )
     }
   }
 }
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -136,11 +136,24 @@ export class TriggerManager extends Component {
     }
   }
 
+  /**
+   * Get the ID of the cipher selected by the user in the list
+   *
+   * @returns {string} the cipher ID
+   */
   getSelectedCipherId() {
     const { selectedCipher } = this.state
     return selectedCipher && selectedCipher.id
   }
 
+  /**
+   * Create a new cipher and return its ID
+   *
+   * @param {string} login - the login to register in the new cipher
+   * @param {string} password - the password to register in the new cipher
+   *
+   * @returns {string} the cipher ID
+   */
   async getNewCipherId(login, password) {
     const { vaultClient, konnector } = this.props
 
@@ -164,6 +177,14 @@ export class TriggerManager extends Component {
     return cipher.id
   }
 
+  /**
+   * Find an existing cipher for the account and return its ID
+   *
+   * @param {string} login - the login to set in the cipher
+   * @param {string} password - the password to set in the cipher
+   *
+   * @returns {string} the cipher ID
+   */
   async getExistingCipherIdForAccount(login, password) {
     const { account } = this.state
     const { vaultClient, konnector } = this.props
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -48,7 +48,7 @@ export class TriggerManager extends Component {
       account,
       error: null,
       status: IDLE,
-      step: 'ciphersList',
+      step: account ? 'accountForm' : 'ciphersList',
       selectedCipher: null
     }
   }
@@ -377,9 +377,11 @@ export class TriggerManager extends Component {
           )}
           {step === 'accountForm' && (
             <>
-              <BackButton onClick={this.showCiphersList}>
-                {t('triggerManager.backToCiphersList')}
-              </BackButton>
+              {!account && (
+                <BackButton onClick={this.showCiphersList}>
+                  {t('triggerManager.backToCiphersList')}
+                </BackButton>
+              )}
               <AccountForm
                 account={account || this.cipherToAccount(selectedCipher)}
                 error={error || triggerError}
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -313,7 +313,7 @@ export class TriggerManager extends Component {
     const { fields } = konnector
     const sanitizedFields = manifest.sanitizeFields(fields)
     const account = this.cipherToAccount(selectedCipher)
-    const values = manifest.getDefaultedValues(konnector, account)
+    const values = manifest.getFieldsValues(konnector, account)
 
     const requiredFields = Object.entries(sanitizedFields)
       .filter(([, value]) => value.required)
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -2,7 +2,6 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import get from 'lodash/get'
 import flow from 'lodash/flow'
-import intersection from 'lodash/intersection'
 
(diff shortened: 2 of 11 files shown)


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -131,11 +135,111 @@ export class TriggerManager extends Component {
...
+    cipherData.login.username = login
+    cipherData.login.password = password
+
+    const cipher = await vaultClient.createNewCipher(cipherData, originalCipher)
+    await vaultClient.saveCipher(cipher)
+
+    return cipher.id
+  }
+
+  async getCipherId(login, password) {
```
y-lohse: We can remove this function now.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -226,10 +299,38 @@ export class TriggerManager extends Component {
...
+    const { fields } = konnector
+    const sanitizedFields = manifest.sanitizeFields(fields)
+    const account = this.cipherToAccount(selectedCipher)
+    const values = manifest.getDefaultedValues(konnector, account)
+
+    const requiredFields = Object.entries(sanitizedFields)
+      .filter(([, value]) => value.required)
+      .map(([key]) => key)
+
+    const requiredFieldsInValues = intersection(
```
y-lohse: How about `const hasValuesForRequiredFields = intersction().length === requiredFields.length`, and then use `hasValuesForRequiredFields` in the condition below?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -276,11 +400,13 @@ export class TriggerManager extends Component {
           )}
           {step === 'accountForm' && (
             <>
-              <BackButton onClick={this.showCiphersList}>
-                {t('triggerManager.backToCiphersList')}
-              </BackButton>
+              {!account && (
```
y-lohse: Maybe we can be more explicit here by introducing a `isUpdate = !account` variable?

packages/cozy-harvest-lib/src/helpers/manifest.js
```diff
@@ -193,9 +193,20 @@ export const sanitize = (manifest = {}) =>
       }
     : manifest
 
+const getDefaultedValues = (konnector, account) => {
```
y-lohse: How about `getFieldValues` ?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -131,11 +135,111 @@ export class TriggerManager extends Component {
       this.handleError(error)
     }
   }
+
+  getSelectedCipherId() {
+    const { selectedCipher } = this.state
+    return selectedCipher && selectedCipher.id
+  }
+
+  async getNewCipherId(login, password) {
```
Crash--: can you add the jsdoc ?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -131,11 +135,111 @@ export class TriggerManager extends Component {
...
+    const konnectorName = get(konnector, 'name') || get(konnector, 'slug')
+
+    const cipherData = {
+      id: null,
+      type: 1,
+      name: konnectorName,
+      login: {
+        username: login,
+        password,
+        uris: konnectorURI ? [{ uri: konnectorURI, match: 0 }] : []
```
Crash--: what is match:0 ?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -238,6 +339,29 @@ export class TriggerManager extends Component {
...
+      auth: pickBy(
+        {
+          login: cipher.login.username,
+          password: cipher.login.password,
+          ...customFields
+        },
+        value => Boolean(value)
+      )
+    }
+  }
```
Crash--: Should we create a `Cipher` helper / doctype?

packages/cozy-harvest-lib/src/helpers/manifest.js
drazik: Yep I like it, thanks

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
drazik: Maybe @y-lohse or @edas have the answer? I just moved this part of the code in a refactor, and have no idea why we have this `match: 0` here actually.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
drazik: Or this could be part of the `Account` one? Like `Account::fromCipher`?

Crash--: Deal :)

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -131,11 +135,85 @@ export class TriggerManager extends Component {
       this.handleError(error)
     }
   }
+
+  /**
+   * Get the ID of the cipher selected by the user in the list
+   *
+   * @returns {string} the cipher ID
```
Crash--: false | string

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
y-lohse: Right, I missed that. It's the type of `uri` that we are using — full URL match, regular expression, etc. It's an ENUM, `0` presumably means "Full URL match" (@edas can you confirm?).

I think we can replace this with a more meaningful constant.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -131,11 +135,85 @@ export class TriggerManager extends Component {
...
+
+  /**
+   * Find an existing cipher for the account and return its ID
+   *
+   * @param {string} login - the login to set in the cipher
+   * @param {string} password - the password to set in the cipher
+   *
+   * @returns {string} the cipher ID
+   */
+  async getExistingCipherIdForAccount(login, password) {
```
Crash--: Should we pass the `account` in params? and move this method to the account model? 

Also, we say `Find an existing cipher` but it seems that we create a `newCypher` 🤔

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
edas: from the jslib/enums/uriMatchType.ts: 

```typescript
export enum UriMatchType {
    Domain = 0,
    Host = 1,
    StartsWith = 2,
    Exact = 3,
    RegularExpression = 4,
    Never = 5,
}
``` 

0 means "match only with the domain part", which is the default.

Agree that we should probably have a constant here. I can export the enum from cozy-keys-lib so you can reuse it. Would it make sense ?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -131,11 +135,111 @@ export class TriggerManager extends Component {
...
+
+  async getNewCipherId(login, password) {
+    const { vaultClient, konnector } = this.props
+
+    const konnectorURI = get(konnector, 'vendor_link')
+    const konnectorName = get(konnector, 'name') || get(konnector, 'slug')
+
+    const cipherData = {
+      id: null,
+      type: 1,
```
edas:
```suggestion
      type: CipherType.Login,
```

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -226,10 +273,37 @@ export class TriggerManager extends Component {
...
+    const account = this.cipherToAccount(selectedCipher)
+    const values = manifest.getFieldsValues(konnector, account)
+
+    const requiredFields = Object.entries(sanitizedFields)
+      .filter(([, value]) => value.required)
+      .map(([key]) => key)
+
+    const hasValuesForRequiredFields =
+      intersection(Object.keys(values), requiredFields).length ===
+      requiredFields.length
```
Crash--: Should we move those previous lines to an helper? It seems that they are not a React related stuff, but more "konnector/account"?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
drazik: Actually the `createCipher` seems not to be so well-named. It's more like a `createOrUpdate`

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
drazik: See https://github.com/cozy/cozy-libs/pull/782/commits/712456d95d2e461054896280a139ce713d7403f1

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
edas: In https://github.com/cozy/cozy-keys-lib/pull/25

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
drazik: Done in https://github.com/cozy/cozy-libs/pull/782/commits/5a29d9fb712937e6095b2ce0b2d92a86f6aee79f

packages/cozy-doctypes/src/Account.js
```diff
@@ -29,6 +30,36 @@ class Account extends Document {
...
+    const customFields = (cipher.fields || []).reduce((fields, field) => {
+      fields[field.name] = field.value
+
+      return fields
+    }, {})
+
+    return {
+      auth: pickBy(
+        {
+          login: cipher.login.username,
```
edas: Il manque le relationships avec le cipher justement

drazik: En fait cette méthode sert à transformer le Cipher en `io.cozy.accounts` compatible avec le composant `AccountForm`. Dans ce cas on n'a besoin que de la propriété `auth`. L'ajout de la relationship est faite par un helper [`setVaultCipherRelationship`](https://github.com/cozy/cozy-libs/pull/782/files#diff-436ec256172e55116b84d1c18c97218aR248), qui pourrait effectivement être déplacé dans cozy-doctypes aussi.

edas: Ok mais là tu fais une méthode avec un nom générique dans une classe utilitaire générique, et elle renvoie un contenu incomplet. Du coup soit tu complètes soit tu mets un nom qui correspond mieux

PR #793, opened 2019-09-16
feat: Show account name more prominently

y-lohse:
In the current design, the name of the service is the primary text, but it should be editable. Since this feature won't be available soon, it's better to use the login as primary text.

Capture d'écran 2019-09-16 14 27 18


edas:
ok for me but such changes should be reviewed by @joel-costa

y-lohse:
I discussed it with him before hand :)

PR #796, opened 2019-09-17
feat(cozy-doctypes): Add a generic filename generation on conflict

Crash--:
Sometimes, we need to erase on conflict, but sometimes we need to rename. 

This is the method we will use on rename.

### DIFF
--- a/packages/cozy-doctypes/src/File.js
+++ b/packages/cozy-doctypes/src/File.js
@@ -125,6 +125,29 @@ class CozyFile extends Document {
       throw error
     }
   }
+  /**
+   * Method to generate a new filename if there is a conflict
+   *
+   * @param {String} filenameWithoutExtension A filename without the extension
+   * @return {String} A filename with the right suffix
+   */
+  static generateNewFileNameOnConflict(filenameWithoutExtension) {
+    //Check if the string ends by _1
+    const regex = new RegExp('(_)([0-9]+)$')
+    const matches = filenameWithoutExtension.match(regex)
+    if (matches) {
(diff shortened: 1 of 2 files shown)


packages/cozy-doctypes/src/File.js
```diff
@@ -125,6 +125,29 @@ class CozyFile extends Document {
       throw error
     }
   }
+  /**
+   * Method to generate a new filename if there is a conflict
+   *
+   * @param {String} filenameWithoutExtension A filename without the extension
+   * @return {String} A filename with the right suffix
+   */
+  static generateNewFileNameOnConflict(filenameWithoutExtension) {
```
edas: This means that if you want to save "hello.png" and there is a "hello.doc" you willl end to save a "hello_1.png". Is this really what we want ?

Crash--: If you call this method where there is no name conflict, I can't do anything for you... 

This method should only be called when `statByPath` returns a something. 

Actually this `generateNewFileNameOnConflict` is the very low method, higher function will come pretty soon

packages/cozy-doctypes/src/File.js
```diff
@@ -125,6 +125,29 @@ class CozyFile extends Document {
...
+    //Check if the string ends by _1
+    const regex = new RegExp('(_)([0-9]+)$')
+    const matches = filenameWithoutExtension.match(regex)
+    if (matches) {
+      let versionNumber = matches[2]
```
y-lohse: Nit: I would throw in a `parseInt()` for good measure here, but that's just me.

PR #798, opened 2019-09-18
Revert "feat: Share ciphers with the cozy org"

drazik:
This reverts commit aa0ae01e12a501e176b0f36f55e3c1a327cbbe86.

Since the `createNewCozySharedCipher` method is not yet available in
cozy-keys-lib, using it make the trigger creation process crash unless
you use a linked cozy-keys-lib on the `feat/share-cozy` branch and a
cozy-stack built on the `password-manager` branch. Also, this commit is
present in the cozy-lib's `feat/share-cozy` branch. So it seems this
commit landed in the `feat/pwmanager` by mistake. Reverting it for now,
so we can re-apply it later when everything is ready.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -180,11 +180,10 @@ export class TriggerManager extends Component {
         }
       }
 
-      const cipher = await vaultClient.createNewCozySharedCipher(
+      const cipher = await vaultClient.createNewCipher(
         cipherData,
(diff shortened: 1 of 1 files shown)


edas:
Let's merge https://github.com/cozy/cozy-keys-lib/pull/29 instead of reverting this one

drazik:
That would also work. I thought it wasn't quite ready yet. But if it is, OK :)

drazik:
https://github.com/cozy/cozy-keys-lib/pull/29 has been merged so we don't need to revert anymore.

PR #801, opened 2019-09-19
[Harvest] feat: Update accounts service

y-lohse:
This is the main function for the service that will update `io.cozy.accounts` passwords when a change is detected in the vault. 

It takes a cipher document from couch, decrypts the password using the org key, and updates any `io.cozy.accounts` that use this cipher.

Requires https://github.com/pulls but can already be reviewed :)

### DIFF
--- a/packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
+++ b/packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
@@ -1,4 +1,5 @@
 import SymmetricCryptoKey from 'cozy-keys-lib/transpiled/SymmetricCryptoKey'
+import EncryptionType from 'cozy-keys-lib/transpiled/EncryptionType'
 import get from 'lodash/get'
 import set from 'lodash/set'
 import unset from 'lodash/unset'
@@ -30,7 +31,7 @@ const updateAccountsPassword = async (
   const encryptedUsername = get(bitwardenCipherDocument, 'login.username')
   const bitwardenCipherId = get(bitwardenCipherDocument, '_id')
 
-  const orgKeyEncType = 2
+  const orgKeyEncType = EncryptionType.AesCbc256_HmacSha256_B64
   const orgKey = new SymmetricCryptoKey(
     vaultClient.Utils.fromB64ToArray(cozyKeys.organizationKey),
     orgKeyEncType
--- a/packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
+++ b/packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
@@ -1,0 +1,61 @@
+import SymmetricCryptoKey from 'cozy-keys-lib/transpiled/SymmetricCryptoKey'
+import get from 'lodash/get'
+import set from 'lodash/set'
+import unset from 'lodash/unset'
+
+const updateAccountsPassword = async (
+  cozyClient,
+  vaultClient,
+  bitwardenCipherDocument
+) => {
+  const cozyKeys = await cozyClient
+    .getStackClient()
+    .fetchJSON('GET', '/bitwarden/organizations/cozy')
+
+  const encryptedPassword = get(bitwardenCipherDocument, 'login.password')
+  const bitwardenCipherId = get(bitwardenCipherDocument, '_id')
+
+  const [encTypeAndIv, data, mac] = encryptedPassword.split('|')
+  const [encTypeString, iv] = encTypeAndIv.split('.')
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
```diff
@@ -0,0 +1,61 @@
...
+          _id: bitwardenCipherId,
+          _type: 'com.bitwarden.ciphers'
+        }
+      })
+      .indexFields(['relationships.vaultCipher._id'])
+  )
+
+  await Promise.all(
+    accounts.data.map(account => {
+      set(account, 'auth.password', decryptedPassword)
```
nono: I think you also need to decrypt the `username` from the cipher and put it in the `auth.login` field. As I understand, login and password are encrypted together in a single blob (the `credentials_encrypted`), and if the password is sent without the login, the login will be lost after the update.

nono: Well, I have read again that part of the code in the stack, and it looks like the login+password are encrypted in the `credentials_encrypted` field, but the login is also kept in clear (I don't understand why it has been made this way). So, it should work even you don't set `auth.login` in your code.

packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
```diff
@@ -0,0 +1,61 @@
...
+const updateAccountsPassword = async (
+  cozyClient,
+  vaultClient,
+  bitwardenCipherDocument
+) => {
+  const cozyKeys = await cozyClient
+    .getStackClient()
+    .fetchJSON('GET', '/bitwarden/organizations/cozy')
+
+  const encryptedPassword = get(bitwardenCipherDocument, 'login.password')
```
edas: je me dis que ça serait bien de mettre aussi à jour le username. Il m'arrive rarement de le changer mais ça arrive quand même

packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
y-lohse: Yeah it's also what I observed, but @edas made the same comment below, so I'll update the login with the cipher just in case.

packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
```diff
@@ -0,0 +1,77 @@
+import SymmetricCryptoKey from 'cozy-keys-lib/transpiled/SymmetricCryptoKey'
+import get from 'lodash/get'
+import set from 'lodash/set'
+import unset from 'lodash/unset'
+
+const decryptString = (encryptedString, vaultClient, orgKey) => {
```
edas: Pas bloquant mais tu aurais pu importer CipherString et laisser CipherString faire le découpage de la chaîne chiffrée. Peut-être même du coup utiliser decryptUtf8 plutôt que directement aesDecryptToUtf8.

Pas bloquant mais ça peut permettre d'éviter d'avoir du code à nous en doublon de ce qui est déjà fait dans la jslib

packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
```diff
@@ -0,0 +1,77 @@
...
+) => {
+  const cozyKeys = await cozyClient
+    .getStackClient()
+    .fetchJSON('GET', '/bitwarden/organizations/cozy')
+
+  const encryptedPassword = get(bitwardenCipherDocument, 'login.password')
+  const encryptedUsername = get(bitwardenCipherDocument, 'login.username')
+  const bitwardenCipherId = get(bitwardenCipherDocument, '_id')
+
+  const orgKeyEncType = 2
```
edas:
```suggestion
  const orgKeyEncType = EncryptionType.AesCbc256_HmacSha256_B64
```

avec l'import qui va bien

y-lohse:
This PR now needs https://github.com/cozy/cozy-keys-lib/pull/36 — but aside from that we should be good!

edas:
> This PR now needs [cozy/cozy-keys-lib#36](https://github.com/cozy/cozy-keys-lib/pull/36) — but aside from that we should be good!

done

packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
edas: Vu ensemble. Passer par CipherString aurait nécessité de poser la clef dans le container dans window

PR #807, opened 2019-09-23
Email layouts

ptbrowne:
Goal: 

- We want the ability to send notifications using built-in layouts from apps/konnectors.
- In the future, the stack will be responsible for holding the
templates, applications will have to send only the parts of the template and the stack will do the full templating + mjml-to-html conversion.

Existing: Banks already had template mechanisms for sending notifications

What has been done: Extraction and reworking of the code from Banks.
 
- Provides built-in layout for use in application and konnectors.
- Opiniated way to organize code to send a notification 
- Built in helpers (t, webLink, universalLink, palette)
- CSS used for style uses the cozy-ui palette
- It will be easy to override colors inside the email

README.md contains more description and information.


packages/cozy-notifications/README.md
```diff
@@ -0,0 +1,190 @@
...
+}
+
+MyNotification.template = require('./template.hbs')
+```
+
+## Templates
+
+Templates used in notification views are based on [`handlebars`](https://handlebarsjs.com/)
+and [`handlebars-layouts`](https://www.npmjs.com/package/handlebars-layouts). This makes
+it   simple to create notifications that are based on built-in templates.
```
Crash--: Should we also talk about MJML? Since we're using a few mjml tags in the exemple?

packages/cozy-notifications/src/style.css
```diff
@@ -0,0 +1,102 @@
+h1, h2, h3, h4, h5, h6 {
+  display:block;
+  margin:0;
+  padding:0;
+}
+
+.tableRow {
+  border-bottom: 1px #d6d8da solid;
```
Crash--: there is no way to at least use the UI palette?

packages/cozy-notifications/src/style.css
```diff
@@ -0,0 +1,102 @@
...
+}
+
+.header__desc {
+  white-space: nowrap;
+  text-align: right;
+}
+.header__desc__title {
+  color: #32363f;
+  font-weight: bold;
+  font-size: 16px;
```
Crash--: should it be in rem in order to be responsive?

packages/cozy-notifications/src/urls.js
```diff
@@ -0,0 +1,57 @@
+// Copied from ui/react/AppLinker/native
+
+const getUniversalLinkDomain = cozyUrl => {
+  const url = new URL(cozyUrl)
+  if (url.host.includes('.mycozy.cloud')) {
+    return 'cozy://'
```
Crash--: https://links.mycozy.cloud ?

packages/cozy-notifications/src/urls.js
```diff
@@ -0,0 +1,57 @@
+// Copied from ui/react/AppLinker/native
```
Crash--: and edited 

Can you send your edit upstream?

packages/cozy-notifications/src/mjml.js
```diff
@@ -0,0 +1,15 @@
+import mjml2html from 'mjml'
```
Crash--: There is no need for our custom cozy-mjml?

packages/cozy-notifications/src/style.css
Crash--: Can't we add new component to our cozy-mjml lib?

packages/cozy-notifications/README.md
```diff
@@ -0,0 +1,190 @@
...
+```
+class MyNotification {
+  async buildData() {
+    return {
+      // data that will be used in the template
+    }
+  }
+
+  getPushContent() {
+    return 'push content'
```
Crash--: can it be also an html template? Or should it be only a string? TBH I really see how to use this lib to send an email, but not to send a push notif

ptbrowne: Yes, I think the doc can be improved on this point. The title used for the notification will be the result of getTitle(), and the content will be the result of getPushContent().

packages/cozy-notifications/README.md
ptbrowne: Done

packages/cozy-notifications/README.md
ptbrowne: Done

packages/cozy-notifications/src/style.css
ptbrowne: Em is not well supported by email-client, I could not find anything about Rem but since it is more recent, I would expect to have even less support. In emails, it is better to stick to the tried and true `px`.

packages/cozy-notifications/src/urls.js
ptbrowne: PR in UI: https://github.com/cozy/cozy-ui/pull/1172

I still have to copy the file in cozy-notifications since :

1. cozy-ui is shipped with es6 modules, that makes it unsuitable for consumption in connectors (for now, we have to discuss on how to remedy that)

2. the universalLink mechanism relies on overriding the value at build time but we will not do that for konnectors

I made an issue here : https://github.com/cozy/cozy-ui/issues/1173

packages/cozy-notifications/src/style.css
ptbrowne: cozy-mjml is not structured as a lib for now, its entrypoint is a binary. I agree that we should probably merge the two libs, I am not sure if it is the right time though.

I am going to try to see if it is easy to add a new component but since the original goal of this PR was to extract the code from banks to be able to use it in konnectors, I think that if I do not succeed quickly, I will postpone this task. Would you be ok with that ?

packages/cozy-notifications/src/urls.js
ptbrowne: Done

packages/cozy-notifications/src/mjml.js
ptbrowne: See https://github.com/cozy/cozy-libs/pull/807#discussion_r327967036

packages/cozy-notifications/src/style.css
Crash--: I didn't know that this lib was taken from banks ! 

So let's not use cozy-mjml for now, but is it possible to harmonize this lib? For instance it seems that in our exemple here our cozy-logo is 100px width, and there https://github.com/cozy/cozy-libs/blob/master/packages/cozy-mjml/src/components/MJHeader.js#L19 it seems to be 129px ?

ptbrowne: I fixed the example :)

ptbrowne: > there is no way to at least use the UI palette?

Done in https://github.com/cozy/cozy-libs/pull/807/commits/ba7e1dc7e614ad2876d0c8bf00bc9a726b4f623f

PR #813, opened 2019-09-26
fix(cozy-harvest-lib): Update selected cipher if needed

drazik:
When the user selects a cipher with incorrect or missing data, he has to
fill the missing fields. We were not updating the cipher in this scenario.


ptbrowne:
No test changed ?

PR #816, opened 2019-09-27
feat: Mark the package as sideEffects free

ptbrowne:
This enables tree shaking.

See https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free
for more information

### DIFF
--- a/packages/cozy-harvest-lib/package.json
+++ b/packages/cozy-harvest-lib/package.json
@@ -61,5 +61,6 @@
     "cozy-device-helper": "1.7.1",
     "cozy-realtime": "^3.1.0",
     "cozy-ui": "^22.12.0"
-  }
+  },
+  "sideEffects": false
 }


Crash--:
Nice catch! 

Just a simple question: if a package emits a few events (with microee for instance), should it be considered as having side effect?

ptbrowne:
> Just a simple question: if a package emits a few events (with microee for instance), should it be considered as having side effect?

It depends if your package does that at import time. If your package emits events at import time, it is not side effect free.

It will be clearer with an example

app.js

```
import { lib1 } from 'lib'
```

lib.js

```
export { default as lib1 } from './lib1'
export { default as lib2 } from './lib2'
```

Here the idea of tree shaking is detecting that we do not need lib2 therefore we can remove its `export` from 'lib.js'.

Let's look at lib2 now :

lib2.js

```
emitEvent('toto')
```

Here, if we remove lib2, the event on which might rely someone is not triggered, therefore the lib.js has not the same behavior before being tree shaked and after.

If lib2 does only export functions and does not has any effect on the exterior, it can be considered side-effect free.

lib2.js

```
export const emit = () => emitEvent()
```

PR #820, opened 2019-10-01
feat: Schedule non-cozy packages updates monthly

ptbrowne:
Have updates for non cozy packages be updated monthly.


ptbrowne:
I am not sure it is worth the additional noise, I cannot recall of a time where I needed a functionality from a non-cozy package that had not been released more than one month ago.

drazik:
OK. We can still upgrade by hand if we need something quickly :+1:

PR #822, opened 2019-10-02
feat: Export a bound getErrorLocale

ptbrowne:
For easy usage in apps/konnectors, export a getErrorLocale bound
to the t of harvest-lib.


ptbrowne:
I had forgotten to push `index.js` that implemented `getBoundT` (because locales/ files are git-ignored). I used node-polyglot directly there, this is why it is a real dependency, not a dep one.

PR #825, opened 2019-10-03
[Harvest] Redirect to the only account

y-lohse:
If there's only one account, we don't want to show the account selection screen.


packages/cozy-harvest-lib/src/components/Routes.jsx
```diff
@@ -10,17 +10,30 @@ import NewAccountModal from './NewAccountModal'
...
-              <AccountsListModal konnector={konnector} accounts={accounts} />
-            )}
+            render={() => {
+              if (accounts.length === 0) {
+                history.push('./new')
+                return null
+              } else if (accounts.length === 1) {
+                history.push(`./accounts/${accounts[0].account._id}`)
+                return null
+              } else {
```
Crash--: Seriously? Is it working? No redirecting loop even without '/' in the first place?

PR #828, opened 2019-10-09
Harvest/404

y-lohse:
Following up on https://github.com/cozy/cozy-home/pull/1350, this PR catches non existing routes inside the harvest modal and redirects them to the root of the modal. To do this, I had to use a `Switch` component so that the fallback redirect is only rendered if nothing else matches. I've also made sure that an inexistent account id renders something.

Still not perfect, if the slug in the URL doesn't match a konnector at all we don't display anything nice, but that's a problem for cozy-home.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/Routes.jsx
+++ b/packages/cozy-harvest-lib/src/components/Routes.jsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Route, Redirect, withRouter } from 'react-router'
+import { Route, withRouter } from 'react-router'
 import trimEnd from 'lodash/trimEnd'
 import Modal from 'cozy-ui/transpiled/react/Modal'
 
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/Routes.jsx
```diff
@@ -10,83 +10,86 @@ import NewAccountModal from './NewAccountModal'
...
-          />
-          <Route
-            path={`${konnectorRoot}/accounts/:accountId/success`}
-            exact
-            render={({ match }) => {
-              return (
-                <KonnectorSuccess
+const Routes = ({ konnectorRoot, konnector, location, history, onDismiss }) => {
+  // we need to make sure the path ends with a / for relative links to work
+  if (/\/$/.test(location.pathname) === false) {
```
Crash--: I find `location.pathname.endsWith("/")` much more easier to read

PR #829, opened 2019-10-10
Harvest/fix conditional folder

y-lohse:
We don't want to show the folder link for konnectors that don't load files.

(PR also includes a tiny fix for the link button that had wrong margins on mobile. The `full` extension cancels out the margins from the `Stack` component, we don't want that.)


Crash--:
> The full extension cancels out the margins from the Stack component, we don't want that.)

Should we fix this issue instead of removing it?

y-lohse:
I'll talk about it in the meeting next tuesday. We should fix it in UI, but it's a breaking change that's definitely going to break things, and we use lots of buttons.

PR #830, opened 2019-10-10
Harvest/fixes

y-lohse:
Some micro fixes on harvest, the commits messages explain what's going on. The only one that's maybe not quite clear is the removed error message, it seems it was an error to add it, since there's another component just next to it that displays the error.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountModal.jsx
@@ -140,7 +140,7 @@ export class AccountModal extends Component {
               trigger={trigger}
               account={account}
               onAccountDeleted={onDismiss}
-              addAccount={() => history.push('../new')}
+              addAccount={() => history.push('../../new')}
             />
           )}
         </ModalContent>
--- a/packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
+++ b/packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
@@ -20,7 +20,7 @@ export class LaunchTriggerCard extends Component {
         <TriggerLauncher
           {...pick(this.props, Object.keys(DumbTriggerLauncher.propTypes))}
         >
-          {({ error, launch, running, trigger }) => {
+          {({ launch, running, trigger }) => {
             const lastSuccessDate = triggers.getLastSuccessDate(trigger)
             return (
               <div className="u-flex u-flex-column-s">
@@ -60,11 +60,6 @@ export class LaunchTriggerCard extends Component {
                   </li>
                 </ul>
                 <div>
-                  {error && (
-                    <Text className="u-error u-mb-1">
-                      {t('card.launchTrigger.error')}
-                    </Text>
-                  )}
                   <Button
                     label={t('card.launchTrigger.button.label')}
                     icon={<Icon focusable="false" icon="sync" spin={running} />}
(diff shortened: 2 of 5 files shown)


packages/cozy-harvest-lib/src/components/AccountModal.jsx
```diff
@@ -140,7 +140,7 @@ export class AccountModal extends Component {
               trigger={trigger}
               account={account}
               onAccountDeleted={onDismiss}
-              addAccount={() => history.push('../new')}
+              addAccount={() => history.push('../../new')}
```
ptbrowne: It seems a bit brittle to use relative routes like that, do you think it would be possible/better to use an absolute path ?

ptbrowne: A test for this would help prevent errors.

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
```diff
@@ -60,11 +60,6 @@ export class LaunchTriggerCard extends Component {
                   </li>
                 </ul>
                 <div>
-                  {error && (
-                    <Text className="u-error u-mb-1">
-                      {t('card.launchTrigger.error')}
-                    </Text>
-                  )}
```
Crash--: Does that mean that we already display the error somewhere? If yes, where?

packages/cozy-harvest-lib/src/components/AccountModal.jsx
y-lohse: > It seems a bit brittle to use relative routes like that, do you think it would be possible/better to use an absolute path ?

I honestly don't think it's more fragile than using absolute paths. In this case we changed the structure of the routes and forgot to change a link… regardless of absolute or not this would have been broken.

While it's possible to use absolute routes, I'm convinced it's in fact not better. Harvest is like procedures, it has its own routes that are mounted on a "root" route porovided by the application. In procedures we used absolute routes, which means that every bit of code that does a redirect needs to be aware of the "root url" provided by the app. With relatives routes, components only need to be aware of their location within harvest, which I think makes more sense and find easier to maintain so far.

ptbrowne: > it has its own routes that are mounted on a "root" route porovided by the application. In procedures we used absolute routes, which means that every bit of code that does a redirect needs to be aware of the "root url" provided by the app. 

OK interesting, I had not thought about the "root URL" problem. This is true that components should not know about this. An alternative: in the context of harvest, it might make more sense for components to be passed a scoped `history`, or a `history` that would be aware of the "mount point" of harvest.

```
history.push('<harvest-root>/new')
```

This is not in the context of the PR but I am interested in your thoughts about this.

y-lohse: I… quite like that idea? As a dev we'd need to be aware not to use `withRouter`, but other than that it might be a better solution than relative routes. They *do* work, but the DX isn't as smooth as I expected, even with react-router v4. Your solution doesn't even require an upgrade to v4. I'll keep it in mind 👍

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
y-lohse: Yes, it's [here](https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx#L89-L91) — just above this component :)

PR #831, opened 2019-10-10
fix: Translating components

y-lohse:
There was some confusion going on regarding translations in harvest, where we would use `translate` and `withLocales`.

- `translate` only works if there is a `withLocales` further up the tree
- `withLocales` loads the translation strings, so it should only be needed once
- however harvest exposes multiple components — in order for the to work standalone, they need to use `withLocales`

So now, every component exporter by harvest is wrapped in `withLocales`, and all the others are wrapped in `translate`. I left 2 or 3 exceptions because they are components that I'm not sure how they are used and where, so we'll have to get back to them later.

### DIFF
--- a/packages/cozy-harvest-lib/README.md
+++ b/packages/cozy-harvest-lib/README.md
@@ -21,16 +21,36 @@ Just run
 yarn add cozy-harvest-lib
 ```
 
+# Importing
+
+Components imported from cozy-harvest-lib **must** be imported from the root of the package, and not from specific files:
+
+```
+// imported from the root of the package 👍
+import { TriggerManager } from 'cozy-harvest-lib'
+
+// imported from another file or folder 👎
+import TriggerManager from 'cozy-harvest-lib/TriggerManager'
+```
+
+This is because all harvest components come with their own translation context. If you really do need to import a component straight from the shipped file, there is a `withLocales` HOC that is available, so something like this would work as a fallback:
+
+```
+import TriggerManager from 'cozy-harvest-lib/TriggerManager'
+import { withLocales } from 'cozy-harvest-lib'
+
+const MyTriggerManager = withLocales(TriggerManager)
+```
+
 # Getting Started
 
 For now it is possible to instanciate a `<TriggerManager />` which will allow to edit an account and launch the trigger.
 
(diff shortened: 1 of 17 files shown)


packages/cozy-harvest-lib/README.md
```diff
@@ -21,28 +21,49 @@ Just run
...
+
+This is because all harvest components come with their own translation context. If you really do need to import a component straight from the shipped file, there is a `withLocales` HOC that is available, so something like this would work as a fallback:
+
+```
+import TriggerManager from 'cozy-harvest-lib/TriggerManager'
+import { withLocales } from 'cozy-harvest-lib'
+
+const MyTriggerManager = withLocales(TriggerManager)
+```
+
```
ptbrowne: Pour mon info, pourquoi les composants sont pas directement exportés wrappés dans withLocale?

packages/cozy-harvest-lib/README.md
```diff
@@ -21,28 +21,49 @@ Just run
...
 ReactDOM.render(
   <CozyProvider client={client}>
-    <I18n lang="en" dictRequire={lang => require(`../src/locales/${lang}`)}>
-      // Fetch konnector at some point
-      <TriggerManager konnector="konnector" onSuccessLogin={() => alert('logged in')} />
-      // other stuff
-    </I18n>
+    // Fetch konnector at some point
+    <TriggerManager
+      konnector="konnector"
```
ptbrowne: Having a string here is a bit error inducing I think. What do you think of putting `{konnector}` here ?

ptbrowne: What do you think of putting a `<Query/>` instead of the "Fetch konnector at some point" ? It would be more real, and would not necessitate much more lines ?

packages/cozy-harvest-lib/README.md
y-lohse: Sur le principe pour moi, withLocales injecte le bon contexte de traductions, alors que `translate` injecte juste une prop `t`. Et donc je trouve préférable que l'on utilise `translate` dans les composants de harvest comme on le fait partout ailleurs.

Sachant que le seul cas ou ca pose problème — import d'un composant pas exporté "officiellement" — n'est pas quelque chose que l'on veux encourager du tout sur harvest, ca me semblait mieux.

ptbrowne: OK ca marche 👍

PR #833, opened 2019-10-11
feat: Create a new cozy-hocs

Crash--:
The idea of this lib is to gather HoCs we need in our applications or in our "modules". 

Here it just an exemple that I needed a few time in our "Scan module" to detect and react (haha) on a network change.

### DIFF
--- a/packages/cozy-hocs/README.md
+++ b/packages/cozy-hocs/README.md
@@ -1,0 +1,18 @@
+# HoCs
+
+cozy-hocs is a lib that gather and expose a few HoCs
+
+## Demo
+
+```jsx
+import { withOffline } from "cozy-hocs";
+
+const MyComponent = props => {
+  if (props.isOffline) {
+    return "Offline";
+  }
+  return "Online";
+};
+
+export default withOffline(MyComponent);
+```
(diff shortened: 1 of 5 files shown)


ptbrowne:
Do you think it could live in cozy-ui ? I think we should refrain ourselves from having to much libs, it makes it difficult to follow what is where etc...

Crash--:
I don't think it has to be in UI. UI is here for our UI component. And it you want to install our modules we install one of our lib (authenticate, procedure, harvest, soon scanner). And if you want to manually connect UI and data, you can combine cozy-hocs with UI. 

For instance, I don't think `translate` is in its right place in UI. It should be in this repo. 

Let's discuss this issue on Tuesday to see what we do we that. I'm not in a hurry ;)

ptbrowne:
Can you provide some advantages of separating this much ? I think the architecture of cozy's libs is already a bit too sprawling/tentacular. I think this makes it difficult cognitevely to remember where everything is. Having a separate repo means that it makes it more difficult to make atomic changes. To me, "modules" could certainly live inside cozy-ui, it would make adoption easier. The thing we have to watch is the size of the generated CSS from cozy-ui, but I think we could exclude modules from the CSS generated for cozy-u.

PR #834, opened 2019-10-14
Harvest/ui tweaks

y-lohse:
- Introduced a variant of the modal header where the layout is vertical instead of horizontal
- Corrected margins on one of the forms

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountsListModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountsListModal.jsx
@@ -2,16 +2,25 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { withRouter } from 'react-router'
 import { ModalContent } from 'cozy-ui/transpiled/react/Modal'
+import Stack from 'cozy-ui/transpiled/react/Stack'
+import { translate } from 'cozy-ui/transpiled/react/I18n'
 import AccountsList from './AccountsList/AccountsList'
-import KonnectorModalHeader from './KonnectorModalHeader'
+import KonnectorIcon from './KonnectorIcon'
 
 class AccountsListModal extends React.Component {
   render() {
-    const { konnector, accounts, history } = this.props
+    const { konnector, accounts, history, t } = this.props
     return (
       <>
-        <KonnectorModalHeader konnector={konnector} />
         <ModalContent>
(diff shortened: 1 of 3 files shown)


Crash--:
> Introduced a variant of the modal header where the layout is vertical instead of horizontal
 
Can you share a use case? screenshot ?

packages/cozy-harvest-lib/src/components/KonnectorModalHeader.jsx
```diff
@@ -2,16 +2,28 @@ import React from 'react'
...
+          className={cx('u-w-3', 'u-h-3', vertical ? 'u-mb-half' : 'u-mr-half')}
+        >
           <KonnectorIcon konnector={konnector} />
         </div>
-        <div className="u-flex-grow-1 u-mr-half">
-          <h3 className="u-title-h3 u-m-0">{konnector.name}</h3>
+        <div
+          className={cx('u-flex-grow-1', vertical ? 'u-mb-half' : 'u-mr-half')}
+        >
+          <h3 className="u-title-h3 u-m-0">{title}</h3>
```
ptbrowne: IMHO having to introduce a new API (`vertical`) and having to conditionally put utility classes shows that we are not working at the right level here.

What we seem to want here seems to switch between a stack based layout and a row based layout, with margins between items :

```
const Wrapper = this.props.wrapper // (Stack || Row)
<Wrapper spacing='m'>
</Wrapper>
```

ptbrowne: What happened to this ?

y-lohse: Wtf, I thought I posted a comment explaining the latest changes. Must have forgotten to send it..? Adding it again.

y-lohse:
Here's a screenshot of the other style:

Capture d'écran 2019-10-17 11 16 54

I also agree in hindsight that the `vertical` API change wasn't the best. After talking with @joel-costa , it turns out that in the 2 cases I'm changing, the konnector icon and title shouldn't be a header at all, they are part of the modal's body, should scroll with it if it overflows, etc. So in the end I reverted the changes to `KonnectorModalHeader` and added the new content to the modals that needed the change.

PR #837, opened 2019-10-16
Harvest/ Maintenance information

y-lohse:
Displays a warning when a konnector is in maintenance mode.


packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx
```diff
@@ -29,128 +31,139 @@ import LaunchTriggerCard from '../cards/LaunchTriggerCard'
...
-      addAccount,
-      history,
-      t
-    } = this.props
+const KonnectorAccountTabs = ({
+  konnector,
+  trigger: initialTrigger,
+  account,
+  onAccountDeleted,
+  //TODO on``
```
edas: I understand this TODO was already there but do you know why we have one and what is left to do? Otherwise we'd better remove it.

Crash--: Yes, it's to rename the following props. `addAccount` to `onAccountAdded` I think

packages/cozy-harvest-lib/src/components/hooks/useMaintenanceStatus.jsx
```diff
@@ -0,0 +1,27 @@
...
+  const registry = new Registry({
+    client
+  })
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const appStatus = await registry.fetchApp(slug)
+      const isInMaintenance = get(appStatus, 'maintenance_activated', false)
+      const messages = get(appStatus, 'maintenance_options.messages', {})
+      setMaintenanceStatus({ isInMaintenance, messages })
```
ptbrowne: Nit: I think the hook should be in charge just of the fetching and let the data extraction (the two gets) to helpers outside the hook. This would enable those helpers (getMessages and isActive) to be put directly on the Registry class as static methods. This would also remove the need to have a new API (`isMaintenance` and `messages` to communicate with the components.

 I think this would make the hook more flexible as we would not need to have another hook if we want to access other parts of the status response.

y-lohse: Yes, agree. There isn't much else in the maintenance status for us so I also think it can stay as it is for now, but it would be a good improvement in the future.

PR #839, opened 2019-10-18
feat(Scanner): Add Scanner module

Crash--:
Create a new "Scanner" module. 

Used in Drive ATM 

- [x] Use withOffline from UI (need to be done) 
- [x] Use ExperimentalModal instead of Modal (semantic release didn't create a release with the new  style...) 
- [x] Write a few tests at least for getter


packages/cozy-scanner/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "cozy-scanner",
+  "version": "0.0.1",
+  "description": "Cozy-Scanner is used to access device comera and also provide component to categorize document",
```
ptbrowne:
```suggestion
  "description": "Cozy-Scanner provides a component to take a picture of a document and categorize it",
```

packages/cozy-scanner/src/DocumentCategory.jsx
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * DocumentCategory component
+ *
+ * The goal of this component is to display a
+ * category / type of document and also manage
+ * its associated ActionMenu since a category has
+ * several associated items.
+ *
+ * If an item from a category is selected, we display
+ * its label instead of the label from the category
```
ptbrowne: Is it still true ? Don't we show two lines now ?

packages/cozy-scanner/src/DocumentCategory.jsx
```diff
@@ -0,0 +1,125 @@
...
+} from 'cozy-ui/transpiled/react/ActionMenu'
+import palette from 'cozy-ui/transpiled/react/palette'
+import GridItem from 'cozy-ui/transpiled/react/Labs/GridItem'
+
+import CategoryGridItem from './CategoryGridItem'
+
+import styles from './styles.styl'
+
+/**
+ * DocumentCategory component
```
ptbrowne: I think this line is not necessary.

packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -0,0 +1,201 @@
...
+ * Document Qualification
+ *
+ * This screen is used to qualify a document ie:
+ *  Selecting a metadata category etc
+ *  Renaming the file
+ *
+ * When a selection is done, we call a callback from
+ * its parent
+ *
+ */
```
ptbrowne: The docstring is not attached to the component.

packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -0,0 +1,201 @@
...
+            <InputGroup
+              fullwidth
+              append={<span className="u-pr-1">{filename_extension}</span>}
+              className="u-bdrs-3"
+            >
+              <Input
+                placeholder={t('Scan.filename')}
+                value={filename}
+                onChange={event => {
+                  //If the user write something once, we don't want to rename the file automatically anymore
```
ptbrowne:
```suggestion
                  // If the user writes something once, we don't want to rename the file automatically anymore
```

packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -0,0 +1,201 @@
...
+                  if (!hasUserWrittenFileName) {
+                    this.setState({ hasUserWrittenFileName: true })
+                  }
+                  //If we left an empty value, then we reset the behavior
```
ptbrowne:
```suggestion
                  // If we left an empty value, then we reset the behavior
```

packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -0,0 +1,201 @@
...
+
+DocumentQualification.defaultProps = {
+  editFileName: false
+}
+DocumentQualification.propTypes = {
+  /**
+   * This callback is called after a select.
+   *
+   */
+  onQualified: PropTypes.func,
```
ptbrowne:
```suggestion
  /** Called after category selection */
  onQualified: PropTypes.func,
```

Then why isn't the name onCategorySection ;)

packages/cozy-scanner/src/DocumentTypeData.js
```diff
@@ -0,0 +1,293 @@
...
+    icon: 'bank',
+    file_type_ids: ['34', '35', '36']
+  },
+  {
+    id: 'theme8',
+    label: 'invoice',
+    icon: 'euro',
+    file_type_ids: ['27', '28', '29', '26', '30']
+  }
+]
```
ptbrowne: Nit: Could be a JSON. Helpful when working with jq.  Not mixing data with code :+1:

packages/cozy-scanner/src/DocumentTypeData.js
```diff
@@ -0,0 +1,293 @@
...
+        items.push(getItemById(itemId))
+      })
+    }
+  })
+  return items
+}
+
+export const getItemById = id => {
+  return items.find(item => item.id === id)
+}
```
ptbrowne: When calliing a method like that, I would not expect it to be O(n). Myabe you can build a indexed view of items ?

```
const itemsById = keyBy(items, item => item.id)
```

packages/cozy-scanner/src/DocumentTypeData.js
```diff
@@ -0,0 +1,293 @@
...
+
+export const getThemeByItem = item => {
```
ptbrowne:
```suggestion
export const getItemTheme = item => {
```

packages/cozy-scanner/src/DocumentTypeData.js
```diff
@@ -0,0 +1,293 @@
...
+  if (item.defaultTheme) {
+    return themes.find(theme => theme.id === item.defaultTheme)
+  }
+  return themes.find(theme => theme.file_type_ids.includes(item.id))
```
ptbrowne: We have to scan in O(n*m) here to find the theme of an item.

I think it would be better (more performant, and the data would be clearer) if the data was reorganized. The theme could be flattened in the items at launch time for example.

packages/cozy-scanner/src/Scanner.jsx
```diff
@@ -0,0 +1,259 @@
...
+    const { generateName } = this.props
+    const name = filename === '' ? generateName() : filename
+    this.setState({ status: SCANNER_UPLOADING, filename: name })
+    const { dirId, onConflict, onBeforeUpload, onFinish } = this.props
+    const onResolvedLocalFS = async fileEntry => {
+      fileEntry.file(
+        file => {
+          const reader = new FileReader()
+          reader.onloadend = async () => {
+            //we get the of the readAsBuffer in the `result` attr
```
ptbrowne: What ? ;)

packages/cozy-scanner/src/Scanner.jsx
```diff
@@ -0,0 +1,259 @@
...
+          onConflict,
+          metadata
+        )
+      }
+    } catch (error) {
+      if (/Not Found/.test(error)) {
+        return await this.upload(name, file, dirId, metadata)
+      }
+      throw error
+    }
```
ptbrowne: IMHO the two methods, before being extracted to FileCollection, should be extracted from the component.

packages/cozy-scanner/src/Scanner.jsx
```diff
@@ -0,0 +1,259 @@
...
+      metadata
+    })
+  }
+  onFail = message => {
+    this.setState({ loadingScreen: false })
+    console.log('failed', message)
+  }
+  startScanner = () => {
+    try {
+      //
```
ptbrowne: Extra comment

packages/cozy-scanner/src/Scanner.jsx
```diff
@@ -0,0 +1,259 @@
...
+    )
+  }
+}
+
+Scanner.defaultProps = {
+  pluginConfig: {}
+}
+/**
+ *
+ */
```
ptbrowne: Extra comment.

packages/cozy-scanner/src/ScannerQualification.jsx
```diff
@@ -0,0 +1,91 @@
...
+import { translate } from 'cozy-ui/transpiled/react/I18n'
+import Alerter from 'cozy-ui/transpiled/react/Alerter'
+
+import DocumentQualification from './DocumentQualification'
+
+import { Modal as NewModal } from 'cozy-ui/transpiled/react'
+
+import { getTracker } from 'cozy-ui/transpiled/react/helpers/tracker'
+
+const pushAnalytics = qualification => {
```
ptbrowne:
```suggestion
const trackQualification = qualification => {
```

packages/cozy-scanner/src/ScannerQualification.jsx
```diff
@@ -0,0 +1,91 @@
...
+        'Qualification',
+        qualification.label
+      ])
+    }
+  }
+}
+/**
+ * ScannerQualification component
+ *
+ * This component is used to be displayed just after taking a
```
ptbrowne:
```suggestion
 * Displayed just after taking a
```

packages/cozy-scanner/src/ScannerQualification.jsx
```diff
@@ -0,0 +1,91 @@
...
+      />
+    )
+  }
+}
+
+ScannerQualification.propTypes = {
+  /**
+   * primary action of the modal
+   *
+   */
```
ptbrowne: Can you use comments like this `/** Primary action of the modal */` ?

It is a bit weird here since we do not know that there is a modal at this point.

packages/cozy-scanner/src/locales/en.json
```diff
@@ -0,0 +1,79 @@
...
+  "Scan": {
+    "scan_a_doc": "Scan a doc",
+    "save_doc": "Save the doc",
+    "filename": "Filename",
+    "save": "Save",
+    "cancel": "Cancel",
+    "qualify": "Categorize",
+    "apply": "Apply",
+    "error": {
+      "offline": "You are currently offline and you can't use this functionnality. Try it later",
```
ptbrowne:
```suggestion
      "offline": "You are currently offline and you can't use this functionality. Please, try again later",
```

packages/cozy-scanner/src/locales/en.json
```diff
@@ -0,0 +1,79 @@
...
+      "uploading": "You are already uploading a file. Wait until the end of this upload and try again.",
```
ptbrowne:
```suggestion
      "uploading": "You are already uploading a file. Please wait until the end of this upload and try again.",
```

packages/cozy-scanner/src/locales/en.json
```diff
@@ -0,0 +1,79 @@
...
+      "generic": "Something went wrong. Please try again."
+    },
+    "successful": {
+      "qualified_ok": "You just have successfully categorized your file! "
```
ptbrowne: Sometimes we see qualified, sometimes "categorized", it seems confusing.

packages/cozy-scanner/src/DocumentQualification.jsx
Crash--: we should have this rule in our eslint / prettier config if you want one space after `//`.

packages/cozy-scanner/src/DocumentQualification.jsx
Crash--: Because now we can also changed the filename. Updated the comment

Crash--: Hmm, in fact, since I also have a callback for the onFileNameChange, I think I can remove the stuff related to filename and rename it to onCategorySelection 🤔

packages/cozy-scanner/src/DocumentTypeData.js
Crash--: Modern JS engine has optimization for this kind of behavior even if I agree that we can flattened the datas. Will see what I can do

packages/cozy-scanner/src/Scanner.jsx
Crash--: In fact there are already extracted. I need to use them ;)

packages/cozy-scanner/src/ScannerQualification.jsx
Crash--: I think I don't get your point. 

We are in a component calling `ExperimentalModal`. We know that there is a modal at this point, no?

packages/cozy-scanner/src/locales/en.json
Crash--: We now use `describe` in English

packages/cozy-scanner/src/DocumentQualification.spec.jsx
```diff
@@ -0,0 +1,85 @@
...
+  global.Date.parse = _Date.parse
+  global.Date.now = _Date.now
+}
+
+const unMockDate = () => {
+  global.Date = new Date()
+  global.Date.UTC = Date.UTC
+  global.Date.parse = Date.parse
+  global.Date.now = Date.now
+}
```
y-lohse: Why not use [`mockdate`](https://github.com/konnectors/toutatice/blob/d983393ff8132579a6849f8996a2a8d2cf7a8982/src/helpers/transpileGroupToCozy.spec.js#L8-L14) instead?

packages/cozy-scanner/src/ScannerQualification.jsx
y-lohse: Technically we're calling `ScannerQualification` when using these proptypes :)

packages/cozy-scanner/Readme.md
```diff
@@ -0,0 +1,32 @@
...
+Cozy-Scanner requires :
+
+- a cordova plugin : `cordova plugin add https://github.com/cozy/cordova-plugin-camera.git#bf519ca009eeaec3a47237ee07a09d515cc733df`
+- At least Cozy-UI 25.7
+- Cozy-Doctypes
+
+To use the component and get the style, you need to import the css in your app
+
+```js
+import "cozy-scanner/dist/stylesheet.css";
```
ptbrowne: It would be nice not to have to do that ? Why is this necessary ? Could the scanner component add this stylesheet by itself ?

packages/cozy-scanner/Readme.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+Since Cozy-Scanner uses a few `Aleter` you also have to put `<Alerter />` (from Cozy-UI)
```
ptbrowne:
```suggestion
Since Cozy-Scanner uses a few `Alerter` you also have to put `<Alerter />` (from Cozy-UI)
```

packages/cozy-scanner/babel-transformer.js
```diff
@@ -0,0 +1,3 @@
+const config = require('./babel.config')
+const { createTransformer } = require('babel-jest')
+module.exports = createTransformer(config)
```
ptbrowne: Why is this necessary ? Can you comment ? Is this because we use a custom babel config ?

packages/cozy-scanner/src/CategoryGridItem.jsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <div
+      className={classNames(
+        styles['grid-item'],
+        'u-pb-half u-ph-half u-bxz  u-ellipsis u-c-pointer',
+        {
+          [styles['border-selected']]: isSelected,
+          [styles['grid-item__selected']]: isSelected && label,
+          [styles['grid-item__selected__without_label']]: isSelected && !label,
+          [styles['border-not-selected']]: !isSelected
```
ptbrowne: It seems a `GridItem` that would care about the `isSelected` prop could be extracted here.

It seems abnormal to have to put 4 different classes here also :/

packages/cozy-scanner/src/DocumentCategory.jsx
```diff
@@ -0,0 +1,124 @@
...
+    const { onSelect } = this.props
+    if (onSelect) onSelect(item)
+  }
+  render() {
+    const { isMenuDisplayed } = this.state
+    const { category, isSelected, selectedItem, items, t } = this.props
+    return (
+      <>
+        <GridItem onClick={() => this.toggleMenu()}>
+          <CategoryGridItem
```
ptbrowne: CategoryBox maybe ?

It was unclear to me at this point that GridItem was only dealing with the Layout where CategoryGridItem deals with the border and the selection.

packages/cozy-scanner/src/DocumentCategory.spec.jsx
```diff
@@ -0,0 +1,52 @@
...
+          t={text => text}
+        />
+      </MuiCozyTheme>
+    )
+    expect(asFragment()).toMatchSnapshot()
+    //Click on the item
+    fireEvent.click(getByText('Scan.items.test'))
+    expect(asFragment()).toMatchSnapshot()
+    //Check if the second item is displayed in the ActionMenu
+    expect(getByText('Scan.items.Label2'))
```
ptbrowne: Seems like you do not expect anything here.

packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -0,0 +1,206 @@
...
+import { getItemsByCategory } from './DocumentTypeDataHelpers'
+import GridItem from 'cozy-ui/transpiled/react/Labs/GridItem'
+import styles from './styles.styl'
+
+const filename_extension = '.jpg'
+const id_filename_input = 'filename_input'
+
+/**
+ *
+ * This screen is used to qualify a document ie:
```
ptbrowne:
```suggestion
 * Used to qualify a document ie:
```

packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -0,0 +1,206 @@
...
+ *  Selecting a metadata category etc
```
ptbrowne:
```suggestion
 *   - Selecting a metadata category etc
```

packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -0,0 +1,206 @@
...
+ *  Renaming the file
```
ptbrowne:
```suggestion
 *   - Renaming the file
```

packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * When a selection is done, we call a callback from
```
ptbrowne:
```suggestion
 * When a selection is done, `props.onDescribed` is called with the selected item
```

packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -0,0 +1,206 @@
...
+   *
+   */
+  onDescribed: PropTypes.func,
+  onFileNameChanged: PropTypes.func,
+  title: PropTypes.string,
+  initialSelected: PropTypes.shape({
+    itemId: PropTypes.string,
+    categoryLabel: PropTypes.string
+  }),
+  editFileName: PropTypes.bool
```
ptbrowne: It is unclear to me from the name what this prop does.

packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -0,0 +1,206 @@
...
+}
+
+DocumentQualification.defaultProps = {
+  editFileName: false
+}
+DocumentQualification.propTypes = {
+  /**
+   * This callback is called after a qualification
+   *
+   */
```
ptbrowne: Can you use "shorthand" docstrings ? `/** Called after qualification (with what ?) */`

packages/cozy-scanner/src/DocumentQualification.spec.jsx
```diff
@@ -0,0 +1,85 @@
...
+describe('DocumentQualification', () => {
+  it('Inital render + selection + filename', () => {
```
ptbrowne:
```suggestion
  it('Initial render + selection + filename', () => {
```

packages/cozy-scanner/src/Scanner.jsx
```diff
@@ -0,0 +1,190 @@
...
+import { CozyFile } from 'cozy-doctypes'
+import { Modal } from 'cozy-ui/transpiled/react'
+
+import ScannerQualification from './ScannerQualification'
+import withOffline from 'cozy-ui/transpiled/helpers/withOffline'
+export const SCANNER_IDLE = 'idle'
+export const SCANNER_DONE = 'done'
+export const SCANNER_UPLOADING = 'uploading'
+/**
+ * LoadingScreen is used to create a better transition between
```
ptbrowne: LoadingScreen ?

packages/cozy-scanner/src/Scanner.jsx
```diff
@@ -0,0 +1,190 @@
...
+ * the native camera and the Qualification Modale.
```
ptbrowne:
```suggestion
 * the native camera and the qualification modal.
```

packages/cozy-scanner/src/Scanner.jsx
```diff
@@ -0,0 +1,190 @@
...
+          // Read the file as an ArrayBuffer
+          reader.readAsArrayBuffer(file)
+        },
+        err => {
+          this.setState({ error: err })
+          this.setState({ status: SCANNER_DONE })
+          //Since this module is pretty recent, let's have this info in sentry if needed
+          console.error('error getting fileentry file!' + err) //eslint-disable-line no-console
+        }
+      )
```
ptbrowne: This function seems pretty complex and is not tested as far as I am aware. It seems to deal with several things.

- state of the component
- reading the file
- uploading the file

Do you think we could we extract the file reading and uploading logic out the component so that the component only deal with its state and props ?

```
     this.setState({ status: SCANNER_UPLOADING, filename: name })
     try {
     ... do before upload stuff ...
      const { data: newFile } = await doUpload(imageURI)
      this.setState({ filename: newFile.name }) 
     } catch (e) {
       this.setState({ status: SCANNER_DONE, error: e }) 
     } finally {
       this.setState({ status: SCANNER_DONE }) 
      }
```

packages/cozy-scanner/src/ScannerQualification.jsx
```diff
@@ -0,0 +1,90 @@
...
+        dismissAction={dismissAction}
+        primaryText={t('Scan.save')}
+        primaryAction={async () => {
+          pushAnalytics(qualification)
+          try {
+            await onSave(qualification, filename)
+          } catch (error) {
+            Alerter.error(t('Scan.error.generic'))
+          }
+        }}
```
ptbrowne: You might want to put the handler creation outside the render so that children components can bypass render with PureComponent.

packages/cozy-scanner/src/ScannerQualification.jsx
ptbrowne: Yes for the user/reader, when reading the prop types, (s)he might not know that the implementation is through a modal. If the ScannerQualification component had `Modal` in its name, I think it would be clearer.

packages/cozy-scanner/src/DocumentQualification.spec.jsx
Crash--: Thanks

packages/cozy-scanner/Readme.md
Crash--: At the beginning I wanted to let the app dealing with the style without having to import a css file. But `scanner` use a few stylus files, and this means the app needs to be able to handle stylus file and this is something I want to avoid for the sake of simplicity to use. 

So the option was to generate a stylesheet.css file. At this moment, we have two choice : 
1/ Having each of our component importing this stylesheet 
2/ Let the app importing the stylesheet 

Again, I choose the 2 because copy / pasting a line, is simpler for DX. Having the same stylesheet imported in each component will result in a `big bundle size` (right?) or... you'll need to customize your build tools and this is something I try to avoid.

packages/cozy-scanner/babel-transformer.js
Crash--: Comment added: 

/**
 * Since we are using a custom babel config
 * we need to create a Transformer to tell
 * jest to use it
 */

packages/cozy-scanner/src/DocumentCategory.spec.jsx
Crash--: No. I expect to see the ActionMenu. And to to that, I check if I see this label

packages/cozy-scanner/src/DocumentQualification.jsx
Crash--: renamed to allowEditFilename

packages/cozy-scanner/Readme.md
ptbrowne: > Having the same stylesheet imported in each component will result in a big bundle size (right?) 

No. The stylesheet will be only once in the bundle.

packages/cozy-scanner/src/styles.styl
```diff
@@ -0,0 +1,35 @@
...
+    padding-top rem(12)
+    }
+
+.grid-item-theme {
+    font-size rem(12)
+    padding-top rem(4)
+    }
+.grid-item-label {
+    font-size rem(10)
+    }
```
ptbrowne: Closing brackets are indented, could you fix it please ;)

packages/cozy-scanner/Readme.md
Crash--: Removed the need of stylus since the CSS is pretty easy and having to import manually the stylesheet in the transpiled code was not very easy

packages/cozy-scanner/src/CategoryGridItem.jsx
```diff
@@ -4,7 +4,8 @@ import classNames from 'classnames'
 import palette from 'cozy-ui/transpiled/react/palette'
 import { Icon, IconStack } from 'cozy-ui/transpiled/react'
 
-import styles from './styles.styl'
+import styles from './stylesheet.css'
+console.log('styles', styles)
```
ptbrowne: console.log to remove

packages/cozy-scanner/src/stylesheet.css
```diff
@@ -0,0 +1,30 @@
...
+.grid-item__selected {
+  padding-top: 0.75rem;
+}
+.grid-item-theme {
+  font-size: 0.75rem;
+  padding-top: 0.25rem;
+}
+.grid-item-label {
+  font-size: 0.625rem;
+}
```
ptbrowne: This does not look specific to the scanner ? Can you add a comment so that we do not forget that this should be in UI ?

packages/cozy-scanner/src/CategoryGridItem.jsx
Crash--: already done

packages/cozy-scanner/src/stylesheet.css
```diff
@@ -0,0 +1,30 @@
...
+  is not well centered. So I can't just rem(16) ... */
+.grid-item-title {
+  margin-top: 1.188rem;
+}
+.grid-item__selected__without_label {
+  padding-top: 1rem;
+}
+.grid-item__selected {
+  padding-top: 0.75rem;
+}
```
ptbrowne: Why a selected grid-item has a different padding-top ?

Crash--: because we dont display it in the same way...

PR #843, opened 2019-10-21
fix(cozy-harvest-lib): Can't close unlock form on account creation

drazik:
When creating a new account and no account already exist, we can't go
back to `konnectorRoot/` because there is a redirect from
`konnectorRoot/` to `konnectorRoot/new`, so the unlock form is reopened
everytime. We want to go back to the root.


Crash--:
Don't know if #842 will be merged before, but we should use `pushHistory` after the merge

drazik:
> Don't know if #842 will be merged before, but we should use `pushHistory` after the merge

Yep I just saw this one too. I will wait for it to be merged and do the change.

drazik:
Closing as after rebasing feat/pwmanager on master after #842 merge is OK.

PR #844, opened 2019-10-21
Harvest/fix: Make modal top edges round

y-lohse:
Our modals have rounded edges, but they are defined on the `Modal` component. The `ModalHeader` has square edges, and giving it a background color makes them stand out. Same goes for the `BrandedModalHeader`. 

I've added this issue to our list of things-to-specify with @joel-costa for our Modal rework and found a workaround for this case (giving the whole modal a background color)

### DIFF
--- a/packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
@@ -100,9 +100,10 @@ export class EditAccountModal extends Component {
         size="small"
         closable={isMobile ? false : true}
         closeBtnColor={palette.white}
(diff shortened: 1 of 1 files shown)


packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
```diff
@@ -100,9 +100,10 @@ export class EditAccountModal extends Component {
         size="small"
         closable={isMobile ? false : true}
         closeBtnColor={palette.white}
+        className="u-bg-dodgerBlue"
```
Crash--: Can you make a TODO here or something to link this issue? :)

PR #845, opened 2019-10-21
fix: Use initial trigger running status

y-lohse:
Until a trigger is launched, the `TriggerLauncher` component uses the `initialTrigger` prop to pass down the correct information. However we were ignoring the current status, so if the `initialTrigger` said that the job was running, we were suppressing this information.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
@@ -1,5 +1,6 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
+import get from 'lodash/get'
 import { withClient, withMutations } from 'cozy-client'
 import CozyRealtime from 'cozy-realtime'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -44,7 +45,8 @@ export class TriggerLauncher extends Component {
     this.state = {
       showTwoFAModal: false,
       trigger: initialTrigger,
-      error: triggersModel.getKonnectorJobError(initialTrigger)
+      error: triggersModel.getKonnectorJobError(initialTrigger),
+      running: get(initialTrigger, 'current_state.status') === 'running'
```
ptbrowne: Why do we duplicate the status here ? Why does TriggerLauncher does not let the downstream components get the trigger status ?

ptbrowne: getStatus should be a helper inside triggerModels. This could then be moved to cozy-client.

y-lohse: > getStatus should be a helper inside triggerModels. This could then be moved to cozy-client.

Yep, good point, i'll make the change.

> Why does TriggerLauncher does not let the downstream components get the trigger status ?

Because the trigger status contained in `initialTrigger` doesn't change over time, even when the status trigger is started. TriggerLauncher is connected to cozy-realtime and makes sure downstream components always have an up-to-date status.

ptbrowne: > TriggerLauncher is connected to cozy-realtime and makes sure downstream components always have an up-to-date status.

TriggerLauncher could make sure downstream components always have an up-to-date trigger and job.

PR #846, opened 2019-10-21
fix(doctypes): Better test error

Crash--:
Since this method is in cozy-doctypes, it seems the case has changed...

### DIFF
--- a/packages/cozy-doctypes/src/File.js
+++ b/packages/cozy-doctypes/src/File.js
@@ -199,7 +199,7 @@ class CozyFile extends Document {
         )
       }
     } catch (error) {
-      if (/Not found/.test(error)) {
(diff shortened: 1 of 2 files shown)


packages/cozy-doctypes/src/File.js
```diff
@@ -199,7 +199,7 @@ class CozyFile extends Document {
         )
       }
     } catch (error) {
-      if (/Not found/.test(error)) {
+      if (/Not Found/.test(error.message)) {
```
ptbrowne: Can you add a test ? Seems like this should be extracted from the method. Thanks !

Crash--: Trying to figure out why in this case I have an object and a case issue (since I already have a failing test... testing `Not found`)

PR #847, opened 2019-10-21
refactor: Use model to determine if konnector is running

y-lohse:
Based on https://github.com/cozy/cozy-libs/pull/845#discussion_r336974353


y-lohse:
Me too generally, but I stayed consistent with the rest of the codebase :)

PR #848, opened 2019-10-21
feat: Support 2fa flow done on mobile/app

ptbrowne:
This is called the decoupled flow because the 2FA is done on another device
   than the one where the user logs in. This means that we do not need to show
   a user input and a submit button in the 2FA modal.

Image collée à 2019-10-21 16-44

- [x] The icon is working now.
- [ ] Translations on transifex

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/accounts.js
+++ b/packages/cozy-harvest-lib/src/helpers/accounts.js
@@ -5,9 +5,22 @@ import manifest from './manifest'
 
 const DEFAULT_TWOFA_CODE_PROVIDER_TYPE = 'default'
 
-const PROVIDERS = {
+export const TWOFA_PROVIDERS = {
   EMAIL: 'email',
-  SMS: 'sms'
+  SMS: 'sms',
+  APP: 'app'
+}
+
+// For some 2FA modes, we do not need user input, this is for example the
+// case for the "app" two fa where the user will open the website/app of the
+// provider and click on a notification or a button. For those modes, we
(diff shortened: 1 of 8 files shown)


packages/cozy-harvest-lib/src/locales/en.json
```diff
@@ -315,9 +315,11 @@
     "providers": {
       "default": "Enter the code sent to you by e-mail or SMS",
       "email": "Enter the code sent to you by e-mail",
-      "sms": "Enter the code sent to you by SMS"
+      "sms": "Enter the code sent to you by SMS",
+      "app": "Please check your Bank provider app to authenticate"
```
Crash--: Are we sure we will only have `Banks` provider for this use case?

ptbrowne: No I fixed this.

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -5,9 +5,22 @@ import manifest from './manifest'
...
+// For some 2FA modes, we do not need user input, this is for example the
+// case for the "app" two fa where the user will open the website/app of the
+// provider and click on a notification or a button. For those modes, we
+// does not need to show an input field with a submit button. We only have
+// to wait, the konnector should tell us when everything is OK.
+export const TWOFA_USER_INPUT = {
+  default: true,
+  email: true,
+  sms: true,
+  app: false
```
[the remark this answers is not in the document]
ptbrowne: Done

PR #849, opened 2019-10-22
feat(cozy-harvest-lib): Make TriggerManager usable standalone

drazik:
In cozy-home intents, we use the TriggerManager in a standalone manner
(ie outside of harvest routes). So it must embed the providers (like
HarvestVaultProvider and MuiCozyTheme) that are required for it to work.


y-lohse:
Good catch, but it reminds me of #831 — all the exported components need the translation context too.

If we wrap all components in their providers before exporting them like in this PR, we end up wrapping some components in multiple contexts. Instead I think it's better to do the wrapping in the index.js file, were there is a clear list of publicly available components. What do you think?

PR #854, opened 2019-10-23
Handling double OTP

ptbrowne:
- Handle 2FA flow where we have two SMS
- Try to simplify the code by:
  - Using less "mutations" that are bound to the client: prefer "client" functions
   that take the client as 1st argument. Since those functions do not need initialisation,
   they are more straightforward to use. Moreover they are stateless in the sense that
   they do not keep a reference on the client. This makes them easier to reason about
   than "mutations" closures
  - Integrate the account watcher into the KonnectorJob. Previously there were 3 classes
    dealing with events which made the code a bit too "decoupled". I think having only
    one class responsible of listening to low-levl changes on documents and providing
    higher level events is better than having 3 classes.


packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
```diff
@@ -17,19 +17,36 @@ import accounts, {
...
     this.handleTwoFAMismatch = this.handleTwoFAMismatch.bind(this)
     this.handleSubmit = this.handleSubmit.bind(this)
   }
 
   componentDidMount() {
-    this.props.konnectorJob.on(TWO_FA_MISMATCH_EVENT, this.handleTwoFAMismatch)
+    this.props.konnectorJob
+      .on(TWO_FA_MISMATCH_EVENT, this.handleTwoFAMismatch)
+      .on(TWO_FA_REQUEST_EVENT, this.handleTwoFARequest)
+  }
```
drazik: `componentWillUnmount` with event listeners removal seems to be missing.

ptbrowne:
Thanks drazik :)

PR #861, opened 2019-11-03
cozy-doctypes: Add rawNumber account matching

dohseven:
I spotted an issue in the account matching with my banking data. I have two different accounts, which have almost the same number, with 12 digits (and a letter):

- 123456789012
- 123456789012-A

In this case, both are normalized by `normalizeAccountNumber` to 123456789012, making it impossible to differentiate the two accounts when trying to reconciliate them with fetched accounts.

The work-around I propose for this case is to try to match the `rawNumber` of the accounts when almost everything fails. 

I chose this solution to avoid modifying the corner cases related to BI in `normalizeAccountNumber`. However I find it rather straight-forward, which makes me wondering if I haven't missed something, and if you didn't want to introduce a `rawNumber` matching for some reason. Is this the case?


ptbrowne:
Thanks dohseven ! We have integration tests related to this code, they are encrypted for privacy reasons (run on anonymized data from real accounts). I will try to find the time to run them to see if they are still OK after a change.

dohseven:
Thanks @ptbrowne! I tried to run the tests, but I stumbled upon this encryption. 

Maybe would it be good to add a new test for my case, in order to avoid regressions in the future? Let me know if I can help.

ptbrowne:
> Maybe would it be good to add a new test for my case, in order to avoid regressions in the future? Let me know if I can help.

Yes I was planning to add a test with your example, do you mind posting the two anonymized accounts the test is failing on, with the expected results ?

dohseven:
Anonymized accounts from io.cozy.bank.accounts:
```
{
 "id": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
 "key": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
 "value": {
  "rev": "yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy"
 },
 "doc": {
  "_id": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
  "_rev": "yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy",
  "balance": 1337,
  "cozyMetadata": {
   "updatedAt": "2019-11-03T15:43:56.428Z"
  },
  "institutionLabel": "Fortuneo Banque",
  "label": "Test account 1",
  "number": "123456789012",
  "rawNumber": "123456789012",
  "type": "Market",
  "vendorId": "123456789012"
 }
}

{
 "id": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
 "key": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
 "value": {
  "rev": "yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy"
 },
 "doc": {
  "_id": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
  "_rev": "yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy",
  "balance": 1337,
  "cozyMetadata": {
   "updatedAt": "2019-11-03T15:43:56.415Z"
  },
  "institutionLabel": "Fortuneo Banque",
  "label": "Test account 2",
  "number": "123456789012",
  "rawNumber": "123456789012-A",
  "type": "Savings",
  "vendorId": "123456789012-A"
 }
}


```

The expected results would be something like:
```
expect(normalizeAccountNumber('123456789012')).toBe('123456789012')
expect(normalizeAccountNumber('123456789012-A')).toBe('123456789012')

// This one should match Test account 1
const account1 = {
    balance: 1337,
    label: 'Test account 1',
    number: '123456789012',
    type: 'Market',
    institutionLabel: 'Fortuneo Banque'
  }

// This one should match Test account 2
const account2 = {
    balance: 1337,
    label: 'Test account 2',
    number: '123456789012',
    type: 'Savings',
    institutionLabel: 'Fortuneo Banque'
  }
```

Let me know if you need something else!

ptbrowne:
I just tested your PR and unfortunately it broke one of the unit test where raw account number is the same for accounts having either

- different type (checkings, savings)
- different currency (USD, EUR)

We have those cases for the Milleis bank for example.

Since it seems the raw number check if a bit flaky, I think we should add checks on currency and type :

```
  const matchRawNumberCurrencyType = existingAccounts.find(
    otherAccount =>
      (eqNotUndefined(account.rawNumber, otherAccount.number) ||
        eqNotUndefined(account.number, otherAccount.rawNumber)) &&
      otherAccount.type == account.type &&
      otherAccount.currency == account.currency
  )
  if (matchRawNumberCurrencyType) {
    return {
      match: matchRawNumberCurrencyType,
      method: 'rawNumber-exact-currency-type'
    }
  }
```

Thanks for your example data, it will be helpful, but I think you missed the "Test account 2" ?

dohseven:
I rebased and reworked the commit in order to add the type and currency as you suggested.

> but I think you missed the "Test account 2" ?

I've described the _Test account 2_ with the `io.cozy.bank.accounts` doc and the fetched account: what have I missed?

ptbrowne:
> I've described the Test account 2 with the io.cozy.bank.accounts doc and the fetched account: what have I missed?

Oups, the Refined Github extension mine made the first code block scrollable and I could not see the account, sorry. 

The build is red, it's only a prettier problem :)

dohseven:
This should be it!

Thanks for the coming tests :)

dohseven:
@ptbrowne Anything else I can do to help you with the tests?

ptbrowne:
> @ptbrowne Anything else I can do to help you with the tests?

Thanks for your proposal, I just needed to find a little bit of time. I've updated the tests :+1:

ptbrowne:
@drazik Could you run the tests on your end to see if everything is ok ?

ptbrowne:
Thanks dohseven !

PR #862, opened 2019-11-04
feat: Sort flags

ptbrowne:
Return list of flags sorted.

### DIFF
--- a/packages/cozy-flags/src/flag.js
+++ b/packages/cozy-flags/src/flag.js
@@ -74,7 +74,7 @@ const flag = function() {
 }
 
 export const listFlags = () => {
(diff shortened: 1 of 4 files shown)


packages/cozy-flags/src/flag.js
```diff
@@ -74,7 +74,7 @@ const flag = function() {
 }
 
 export const listFlags = () => {
-  return store.keys()
+  return store.keys().sort()
```
Crash--: Should we take care of possible Uppercase flag?

ptbrowne: We do not have the case, I don't think it is a problem :)

PR #868, opened 2019-11-13
feat(cozy-doctypes): Add generateFileName for a version

Crash--:
cozy-stack now handles file's version. 
cozy-client adds now the ability to download a specific version (https://github.com/cozy/cozy-client/pull/564) 

This PR add a new helper to generate the default Filename for a specific version aka : 
`filename-revisionDate.extension`

### DIFF
--- a/packages/cozy-doctypes/src/File.spec.js
+++ b/packages/cozy-doctypes/src/File.spec.js
@@ -233,6 +233,20 @@ describe('File model', () => {
     })
   })
 
+  describe('generateFileNameForRevision', () => {
+    it('should generate the right file name for a revision', () => {
+      const MOCKED_DATE = '2018-01-01T12:00:00.210Z'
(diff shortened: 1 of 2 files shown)


packages/cozy-doctypes/src/File.spec.js
```diff
@@ -233,6 +237,27 @@ describe('File model', () => {
...
+      MockDate.set(MOCKED_DATE)
+    })
+
+    afterEach(() => {
+      jest.restoreAllMocks()
+      MockDate.reset()
+    })
+    it('should generate the right file name for a revision', () => {
+      const expectedFilename = 'test_01 January - 12h00.pdf'
+      const date = new Date(MOCKED_DATE)
```
ptbrowne: Nit: I am not sure why you MockDate and then reset it, since you pass directly the date to `new Date`, it does not seem necessary.

PR #869, opened 2019-11-13
feat: Create or update can handle duplicates via strategies

ptbrowne:
Before, create or update would throw if more than 1 existing document was
found through the idAttributes selector. It results in errors for konnectors
for which 2 documents have been inserted by error (for example if the vendor returned two documents with the same vendorId).

After discussion with @edas, we resolved to removing those duplicates. To kept backward compatibility and to explicit this behavior, I went with user configurable strategies in case of duplicates.

- The default strategy is to throw an error when a duplicate is found (same behavior as before)
- The "remove" strategy `createOrUpdate(doc, { handleDuplicate: "remove" })` will remove the most ancient duplicate documents and will keep only the most recent

### DIFF
--- a/packages/cozy-scanner/CHANGELOG.md
+++ b/packages/cozy-scanner/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.1.1](https://github.com/cozy/cozy-libs/compare/cozy-scanner@0.1.0...cozy-scanner@0.1.1) (2019-11-12)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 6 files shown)


edas:
Oh, j'avais cru comprendre qu'on était dans le code du connecteur. Dans le code de la lib ça se discute plus comme comportement (mais en même temps je ne vois pas quel autre comportement serait intelligent à avoir, donc si quelqu'un en voit un autre il sera temps de le coder sous forme d'option à ce moment là)

packages/cozy-scanner/CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
...
+
+
+### Bug Fixes
+
+* **scanner:** Fix publish lang ([13e3375](https://github.com/cozy/cozy-libs/commit/13e3375))
+
+
+
+
+
```
Crash--: weird :o

PR #874, opened 2019-11-26
chore: Remove preact

ptbrowne:
Preact and preact-modal were indicated as dependencies here.

Preact and preact-modal are used by cozy-ui as peer dependencies,
so they do not need to be included here. Saw the bug in Banks
where preact-portal was installed even if it's been a long time
that we do not use preact-portal.

It would actually cause a build bug since the new version of cozy-ui that inverted the USE_REACT flag to USE_PREACT. 

1. Since process.env.USE_PREACT is not changed to
false, the require('preact-portal') was executed
2. It would be found in the node_modules but would fail on its require('preact').
3. So even a try/catch around the require('preact-portal') would not
prevent the build error (since preact-portal could be found).

So if you have preact or preact-portal in your node_modules, you MUST use USE_PREACT env var.

### DIFF
--- a/packages/cozy-harvest-lib/package.json
+++ b/packages/cozy-harvest-lib/package.json
@@ -47,7 +47,7 @@
     "cozy-client": "7.13.0",
     "cozy-device-helper": "^1.8.0",
     "cozy-realtime": "3.1.0",
-    "cozy-ui": "27.6.2",
+    "cozy-ui": "27.11.2",
     "enzyme": "3.10.0",
     "enzyme-adapter-react-16": "1.15.1",
     "identity-obj-proxy": "3.0.0",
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/package.json
```diff
@@ -29,12 +29,12 @@
     "@babel/runtime": "^7.5.2",
     "@material-ui/core": "3",
     "cozy-doctypes": "^1.69.0",
```
Crash--: Why do you add UI as a dep? It should remains only dev and peerdep nop?

ptbrowne: Fixed, thanks

PR #876, opened 2019-11-29
Fix icons

y-lohse:
The last PR introduced a different svg loader for babel, which broke the existing svg integrations.


y-lohse:
There are 2 commits, but I'll add the reference here, we'll be able to trace it back via github. The breaking commit was 7444dd5ca0479829f4c5ea8bb1812605384c95f2.

PR #877, opened 2019-11-29
fix: Use data in vaultCipher relationship

Gara64:
See https://github.com/cozy/cozy-doctypes/blob/master/docs/README.md#external-relationships

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/accounts.js
+++ b/packages/cozy-harvest-lib/src/helpers/accounts.js
@@ -113,7 +113,7 @@ export const mergeAuth = (account, authData) => ({
  * @return {string}           cipher uuid
  */
 export const getVaultCipherId = account =>
-  get(account, 'relationships.vaultCipher._id')
+  get(account, 'relationships.vaultCipher.data[0]._id')
 
 /**
  * Adds or updates a vault cipher relationship to an account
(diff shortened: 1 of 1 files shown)


Gara64:
@y-lohse I'm already working on a migration job on the stack side, which will create a cipher for each account and add the relationship in the account

y-lohse:
> @y-lohse I'm already working on a migration job on the stack side, which will create a cipher for each account and add the relationship in the account

I meant a migration for instances that have the relationship without `data`, for which things will break after this change. AFAIK it's only test instances though.

Gara64:
> > @y-lohse I'm already working on a migration job on the stack side, which will create a cipher for each account and add the relationship in the account
> 
> I meant a migration for instances that have the relationship without `data`, for which things will break after this change. AFAIK it's only test instances though.

Yes, it's only testing instances for now, so I think it's safe :)

ptbrowne:
> @y-lohse I'm already working on a migration job on the stack side, 

Is this a migration that will run on each Cozy ? Wouldn't it be simpler at least for deployment to do it with `ach batch` ?

Gara64:
> > @y-lohse I'm already working on a migration job on the stack side,
> 
> Is this a migration that will run on each Cozy ? Wouldn't it be simpler at least for deployment to do it with `ach batch` ?

In this case we need the private key to decrypt credentials accounts, which should remain on the worker side, so it makes more sense to use a [migration job](https://docs.cozy.io/en/cozy-stack/workers/#migrations)

ptbrowne:
TIL migration jobs, thanks 🎉

Gara64:
ping @drazik is it ok for you?

drazik:
Yes. Sorry I missed the first ping :(

PR #882, opened 2019-12-04
feat: We can pass contentType if needed

Crash--:
Two things : 
- Forgot to apply metadata / qualification during the upload... 
- Add an argument to pass contentType to the upload method

### DIFF
--- a/packages/cozy-doctypes/src/File.js
+++ b/packages/cozy-doctypes/src/File.js
@@ -1,5 +1,4 @@
 const trimEnd = require('lodash/trimEnd')
-
 const Document = require('./Document')
 
 const FILENAME_WITH_EXTENSION_REGEX = /(.+)(\..*)$/
@@ -171,12 +170,16 @@ class CozyFile extends Document {
    * @param {ArrayBuffer} file data
    * @param {String} dirId dir id where to upload
    * @param {String} conflictStrategy Actually only 2 hardcoded strategies 'erase' or 'rename'
+   * @param {Object} metadata An object containing the metadata to attach
+   * @param {String} contentType content type of the file
    */
   static async uploadFileWithConflictStrategy(
     name,
     file,
     dirId,
-    conflictStrategy
+    conflictStrategy,
+    metadata,
+    contentType
   ) {
     const filesCollection = this.cozyClient.collection('io.cozy.files')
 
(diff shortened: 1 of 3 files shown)


packages/cozy-doctypes/src/File.js
```diff
@@ -171,12 +170,16 @@ class CozyFile extends Document {
    * @param {ArrayBuffer} file data
    * @param {String} dirId dir id where to upload
    * @param {String} conflictStrategy Actually only 2 hardcoded strategies 'erase' or 'rename'
+   * @param {Object} metadata An object containing the wanted metadata to attach
```
ptbrowne:
```suggestion
   * @param {Object} metadata An object containing the metadata to attach
```

packages/cozy-doctypes/src/File.js
```diff
@@ -171,12 +170,16 @@ class CozyFile extends Document {
...
+   * @param {String} contentType content type of the file
    */
   static async uploadFileWithConflictStrategy(
     name,
     file,
     dirId,
-    conflictStrategy
+    conflictStrategy,
+    metadata,
+    contentType
```
ptbrowne: With 6 arguments it starts to be hard to remember the right order of argument. Have you considered using option objects ? For example the file options could be an object and the conflict strategy could be a second argument. I think it would also be easier when passing the file options down to other functions.

packages/cozy-scanner/src/ScannerUpload.js
```diff
@@ -1,19 +1,21 @@
...
+ * @param {String} contentType Content-Type of the file
  */
 export const doUpload = async (
   imageURI,
   qualification,
   name = '',
   dirId,
-  onConflict
+  onConflict,
+  contentType
```
ptbrowne: Number of argument here also makes remembering the order difficult.

packages/cozy-doctypes/src/File.js
Crash--: Yep, this is planned when migrating to cozy-client (I write a todo there : https://github.com/cozy/cozy-client/pull/571#issue-344568157). I don't want to make a Breaking Change right now

packages/cozy-scanner/src/ScannerUpload.js
```diff
@@ -1,19 +1,21 @@
 import { CozyFile } from 'cozy-doctypes'
 
 /**
- *
+ * @todo this method sould be extracted to Cozy-Client itself
```
y-lohse: Nit, but this shouldn't be a docstring comment imo 😀

PR #883, opened 2019-12-06
feat: Use conventional display name for HOC

ptbrowne:
Breaking change as snapshots will fail.


Crash--:
> **Breaking Changes** should start with the word `BREAKING CHANGE:`

I think you need at least to add `:`

PR #885, opened 2019-12-11
chore: Use flexible dependencies for React

edas:
When we ask for specific versions of React, this
may load React multiple times and break the hooks
behaviour une React. Loading React ^16.12.0 unless
needed.

We had some issues on Travis to run the tests: 
- We excluded /dist to be tested by cozy-scanner
- We increase node memory limit on Travis 
- We have to run tests sequentially 
- Change tests to `wait` since it takes more time on travis and otherwise it timeout 

since I don't know why we're having issues on this PR, I just opened https://github.com/cozy/cozy-libs/pull/886 to see if we have problems on current master


ptbrowne:
> Cozy-scanner has a fixed version because of @testing-library/react

I don't understand this reason :/ Could you explain a bit more ?

packages/cozy-scanner/src/DocumentCategory.spec.jsx
```diff
@@ -47,6 +47,6 @@ describe('DocumentCategory', () => {
     expect(onSelect).toBeCalledWith({
       itemId: '2',
       categoryLabel: 'test'
-    })
+    }) */
```
ptbrowne: Is this intentional ?

edas:
> I don't understand this reason :/ Could you explain a bit more ?

@testing-library/react does use the exact 16.9.0 version of React. We'll test if it's still compatible with newer versions once we resolve the failing test.

WIP (and I missed the WIP tag in the PR, sorry)

ptbrowne:
Le changeemen de parallel à 1 change la durée des tests ?

edas:
En local je n'ai pas compté mais je n'ai pas vu de différence qui m'ait sauté à la figure.

Sur Travis c'est lent de toutes façons, je pense que je n'aurais pas vu la différence même si elle était significative.

PR #902, opened 2020-01-02
[cozy-realtime] Add cozy-client plugin

drazik:
This plugin can be used with the cozy-client's `registerPlugin` method:

```js
import CozyClient from 'cozy-client'
import { RealtimePlugin } from 'cozy-realtime'

  uri: 'http://cozy.tools:8080',
  token: 'sometoken'
})

client.registerPlugin(RealtimePlugin)
```

This will be useful to pass a client with cozy-realtime already attached and initialized to components that needs it (for example cozy-ui's ContactsModalList, see https://github.com/cozy/cozy-ui/pull/1326)


packages/cozy-realtime/src/RealtimePlugin.js
```diff
@@ -0,0 +1,23 @@
+import CozyRealtime from './CozyRealtime'
+
+class RealtimePlugin {
```
Crash--: With a small JSDoc to explain that this a CozyClient's plugin ;)

PR #905, opened 2020-01-06
fix: Allow oauth without scope

doubleface:
The goal here is to avoid the generated oauth popup url to get : scope=undefined parameter, which is not supported by enedis webservice for example.

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/oauth.js
+++ b/packages/cozy-harvest-lib/src/helpers/oauth.js
@@ -76,6 +76,35 @@ export const handleOAuthResponse = () => {
 }
 
 /**
+ * Generate oauth popup url
+ * @param  {string} cozyUrl cozy url
+ * @param  {string} accountType connector slug
+ * @param  {string} oAuthStateKey localStorage key
+ * @param  {Object} oAuthConf connector manifest oauth configuration
+ * @param  {string} nonce unique nonce string
+ */
+export const getOAuthUrl = ({
+  cozyUrl,
+  accountType,
+  oAuthStateKey,
+  oAuthConf,
+  nonce
+}) => {
+  let oAuthUrl = `${cozyUrl}/accounts/${accountType}/start?state=${oAuthStateKey}&nonce=${nonce}`
+  if (
+    oAuthConf.scope !== undefined &&
+    oAuthConf.scope !== null &&
+    oAuthConf.scope !== false
+  ) {
+    const urlScope = Array.isArray(oAuthConf.scope)
+      ? oAuthConf.scope.join('+')
+      : oAuthConf.scope
+    oAuthUrl += `&scope=${urlScope}`
+  }
+  return oAuthUrl
+}
+
+/**
  * Initializes client OAuth workflow by storing the current information about
  * account type in localStorage. Generates the OAuth URL to stack endpoint,
  * passing the localStorage key as state in query string.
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/helpers/oauth.js
```diff
@@ -75,6 +75,20 @@ export const handleOAuthResponse = () => {
...
+ * Generate oauth popup url
+ * @param  {string} cozyUrl cozy url
+ * @param  {string} accountType connector slug
+ * @param  {string} oAuthStateKey localStorage key
+ * @param  {Object} oAuthConf connector manifest oauth configuration
+ * @param  {string} nonce unique nonce string
+ */
+export const getOAuthUrl = (cozyUrl, accountType, oAuthStateKey, oAuthConf, nonce) => {
+  let oAuthUrl = `${cozyUrl}/accounts/${accountType}/start?state=${oAuthStateKey}&nonce=${nonce}`
+  if (oAuthConf.scope) oAuthUrl+= `&scope=${oAuthConf.scope}`
```
edas: Some servers will require the `scope` parameter, even if empty. 

The real fix should probably be on the Enedis side (The client is allowed by the spec to send a scope parameter).

If you want a workaround here, you should at least make a difference between an empty list / empty string, and a null / undefined / false value. If you can't, you'd better define some magic value that means "do not send a scope parameter"

packages/cozy-harvest-lib/src/helpers/oauth.js
```diff
@@ -97,9 +111,7 @@ export const prepareOAuth = (client, konnector) => {
 
   const cozyUrl = client.stackClient.uri
 
-  const oAuthUrl = `${cozyUrl}/accounts/${accountType}/start?scope=${
-    oauth.scope
-  }&state=${oAuthStateKey}&nonce=${Date.now()}`
+  const oAuthUrl = getOAuthUrl(cozyUrl, accountType, oAuthStateKey, oauth, Date.now())
```
edas: There are too much parameters here. You should use an object (here and in the function above) for named parameters

packages/cozy-harvest-lib/src/helpers/oauth.js
Crash--: I think adding a scope : false / undefined can be a good compromise here :)

doubleface: Thanks, I will choose undefined

packages/cozy-harvest-lib/src/helpers/oauth.js
doubleface: I had this in mind but probably was too lazy here. Will do

packages/cozy-harvest-lib/src/helpers/oauth.js
doubleface: I finally chose an empty array as "no scope" value to have the most possibilities available

packages/cozy-harvest-lib/src/helpers/oauth.js
```diff
@@ -75,6 +75,27 @@ export const handleOAuthResponse = () => {
...
+ */
+export const getOAuthUrl = ({
+  cozyUrl,
+  accountType,
+  oAuthStateKey,
+  oAuthConf,
+  nonce
+}) => {
+  let oAuthUrl = `${cozyUrl}/accounts/${accountType}/start?state=${oAuthStateKey}&nonce=${nonce}`
+  if (!Array.isArray(oAuthConf.scope) || oAuthConf.scope.length > 0)
```
edas:
```suggestion
  if (oAuthConf.scope !== undefined || oAuthConf.scope !== false || oAuthConf.scope !== null)
```

packages/cozy-harvest-lib/src/helpers/oauth.js
```diff
@@ -75,6 +75,27 @@ export const handleOAuthResponse = () => {
...
+    oAuthUrl += `&scope=${oAuthConf.scope}`
```
edas:
```suggestion
    oAuthUrl += `&scope=${Array.isArray(oAuthConf.scope) ? oAuthConf.scope.join(' ') : oAuthConf.scope}`
```

doubleface:
@edas the stack does not parse this scope value and the goal of the scope array value was to avoid to have to escape spaces in the manifest.

It is now implemented as expected

PR #906, opened 2020-01-07
fix: Do not set lang inside templateData if it does not exist

ptbrowne:
This error would be thrown if a notification would not return
data inside buildData, this is a convention so that the notification
is not sent.

### DIFF
--- a/packages/cozy-notifications/src/notifications.js
+++ b/packages/cozy-notifications/src/notifications.js
@@ -39,11 +39,15 @@ const getBuiltInHelpersForView = notifView => {
 
 export const buildAttributes = async (notifView, options = {}) => {
   const templateData = await notifView.buildData()
-  templateData.lang = options.lang
 
-  if (notifView.shouldSend && !notifView.shouldSend(templateData)) {
+  if (
+    !templateData ||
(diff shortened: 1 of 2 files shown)


ptbrowne:
> Why not having always a shouldSend method that returns false if templateData is empty? 

ShouldSend is a method from *that can be implemented but it is not compulsory*. So we cannot guarantee that it will always return false in case of empty data.

> Also can you point me where this shouldSend is coming from? 

https://github.com/cozy/cozy-libs/search?utf8=%E2%9C%93&q=shouldSend&type=

You can see its use [in the test](https://github.com/cozy/cozy-libs/blob/1c49f3e024af0832f0138c2b6f6db24cceece919/packages/cozy-notifications/src/notifications.spec.js).

Crash--:
>    Also can you point me where this shouldSend is coming from?
https://github.com/cozy/cozy-libs/search?utf8=%E2%9C%93&q=shouldSend&type=
You can see its use in the test.

I saw that. I missed the point that the goal of `NotificationView` is to be extended by something/someone and that we can not guarantee the check. 

Thanks :)

PR #907, opened 2020-01-08
Make TriggerManager work with locked vault

drazik:
When the user never used his vault (meaning it's empty), we let him create
new accounts without unlocking it. Thus the TriggerManager have to be able
to work fine with a locked vault. In that case, we skip ciphers handling to
just create plain old accounts.

It also include a little refactor so that some missing tests could be added.

I tried to add missing tests for TriggerManager too, but it was a too heavy piece of work, so I'm going to do it in another PR coming soon.


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -271,28 +272,37 @@ export class DumbTriggerManager extends Component {
...
-      if (cipherId) {
-        await this.updateCipher(cipherId, identifier, password)
-      } else {
-        const existingCipherId = await this.getExistingCipherIdForAccount(
-          identifier,
-          password
+      if (isVaultLocked) {
+        // eslint-disable-next-line no-console
+        console.warn(
+          'Impossible to manage ciphers since vault is locked. The created io.cozy.accounts will not be linked to an com.bitwarden.ciphers'
```
y-lohse: Nit: `will not be linked to a com.bitwarden.ciphers`

PR #909, opened 2020-01-13
feat(AccountField): Make it non recognizable by browser extensio…

drazik:
We want harvest's konnector form to not be recognized by password managers browsers extensions, so they don't try to autofill it with wrong credentials guessed from the current URL (which is the cozy instance one, not the one from the vendor the user is configuring a konnector for).

To achieve this, we use several techniques:

* Don't include things like « login » or « password » in inputs `name`, `id` or `className`
* Don't use a `input type="password"`: instead we use a classic text input, but we show dots unicode characters in it
* Obfuscate the labels: this is the hacky part. Some extensions look at the label associated to the input and check if it contains something that can tell them it's a login form. To prevent them from doing this, I obfuscated the label content by adding invisible letters in the middle of the label string. This way, when checking `label.textContent`, extensions see garbage and can't compare it to known string (login, identifiant, mot de passe, password,...). Of course it totally breaks accessibility as a screen reader will not be able to read the label correctly (theorically, I still need to take some time to check it).

In the end, Bitwarden (and Cozy Extension), Lastpass and Dashlane totally ignore this form.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm/Field.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/Field.jsx
@@ -33,6 +33,7 @@ export const Field = ({ label, type, ...props }) => {
     <Component
       {...props}
       label={<ObfuscatedLabel label={label} />}
+      aria-label={label}
       type={type}
     />
   )
--- a/packages/cozy-harvest-lib/src/components/AccountForm/Field.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/Field.jsx
@@ -1,0 +1,98 @@
+import React, { useState } from 'react'
+import UIField from 'cozy-ui/transpiled/react/Field'
+
+const ObfuscateChar = () => {
+  return (
+    <span style={{ fontSize: 0 }} aria-hidden="true">
+      a
+    </span>
+  )
+}
+
+/**
+ * Obfuscate a given label by inserting chars in it.
+ * The chars inserted are not visible, but they are rendered by screen readers,
+ * which is no good for accessibility. We must find a better solution.
+ */
+const ObfuscatedLabel = ({ label }) => {
+  const chars = label.split('')
+
+  return chars.map((char, index) => (
+    <React.Fragment key={`${char}-${index}`}>
+      <ObfuscateChar />
+      {char}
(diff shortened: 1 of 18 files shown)


packages/cozy-harvest-lib/src/components/AccountForm/Field.jsx
```diff
@@ -0,0 +1,98 @@
...
+  return (
+    <span style={{ fontSize: 0 }} aria-hidden="true">
+      a
+    </span>
+  )
+}
+
+/**
+ * Obfuscate a given label by inserting chars in it.
+ * The chars inserted are not visible, but they are rendered by screen readers,
```
edas: Are  we sure this is needed if we already transform the password field in a text field ?

I am not confortable with breaking a11y

edas: If this is needed (and only  if this is needed), I would welcome an `aria-label` or something similar.

drazik: Yes it is needed. Dashlane and lastpass check the label and they seem to have a dictionary for internationalization because they match "mot de passe", "identifiant", "login" and "password".

OK I will add an aria-label

packages/cozy-harvest-lib/test/components/__snapshots__/AccountField.spec.js.snap
```diff
@@ -1,177 +1,341 @@
...
+  labelProps={Object {}}
+  name="date"
+  placeholder="fields.date.placeholder"
+  required={true}
+  secondaryLabels={Object {}}
+  side={null}
+  size="medium"
+  t={
+    [MockFunction] {
+      "calls": Array [
```
y-lohse: A shame about al of this in the snapshots :/ Not related to this PR though.

packages/cozy-harvest-lib/src/components/AccountForm/Field.jsx
drazik: See https://github.com/cozy/cozy-libs/pull/909/commits/2cf3d3101fdb8ab65746708fabcfae6a1b356328

PR #911, opened 2020-01-15
docs: Add a note about the stylesheet import

Crash--:
The readme is not very clear ATM and I don't want to rework it now. I just added a note about the import of the stylesheet import since we have to.


ptbrowne:
docs instead of doc

PR #921, opened 2020-01-21
[eslint-config-cozy-app] Add react-hooks plugin

### LINKED ISSUE #916: Add eslint-plugin-react-hooks to eslint config
Since we are more and more using hooks, I think we should add [eslint-plugin-react-hook](https://www.npmjs.com/package/eslint-plugin-react-hooks#installation) to our eslint config, so the deps of our hooks are automatically checked.

drazik:
Closes #916


Crash--:
https://github.com/cozy/cozy-libs/issues/916

PR #922, opened 2020-01-23
Refactor TriggerManager handleSubmit

ptbrowne:
Break functions and move them out of the component. Having them inside the components makes it
harder to reason about since every method can use the internal state of the component class.

This should also make testing easier since we od not have to instantiate a component to call
those functions.

It makes the logic inside handleSubmit more understandable IMHO. It is a first step towards 
having custom logic inside TriggerManager to support BI connection creation flow.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -63,6 +63,38 @@ const searchForCipher = async (vaultClient, searchOptions) => {
 }
 
 /**
+ * Create a new cipher and return its ID
+ *
+ * @param {string} login - the login to register in the new cipher
+ * @param {string} password - the password to register in the new cipher
+ *
+ * @returns {string} the cipher ID
+ */
+const createCipher = async (vaultClient, createOptions) => {
+  const { konnector, login, password } = createOptions
+  const konnectorURI = get(konnector, 'vendor_link')
+  const konnectorName = get(konnector, 'name') || get(konnector, 'slug')
+
+  const cipherData = {
+    id: null,
+    type: CipherType.Login,
+    name: konnectorName,
+    login: {
+      username: login,
+      password,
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -28,14 +26,39 @@ import triggers from '../helpers/triggers'
...
 import HarvestVaultProvider from './HarvestVaultProvider'
 
+import { createOrUpdateCipher } from '../models/cipherUtils'
+
 const IDLE = 'IDLE'
 const RUNNING = 'RUNNING'
 
 const MODAL_PLACE_ID = 'coz-harvest-modal-place'
 
+const buildOrUpdateAccount = ({ account, userData, cipher, konnector }) => {
```
Crash--: Should we move this function to `./connection/accounts` ?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -62,6 +62,38 @@ const searchForCipher = async (vaultClient, searchOptions) => {
...
+        ? [{ uri: konnectorURI, match: UriMatchType.Domain }]
+        : []
+    }
+  }
+
+  const cipher = await vaultClient.createNewCozySharedCipher(cipherData, null)
+  await vaultClient.saveCipher(cipher)
+
+  return cipher
+}
```
Crash--: same

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
ptbrowne: I still have some refactoring to do, I will do this later on, is it OK for you ?

Crash--: I have no doubt you will ;)

PR #926, opened 2020-01-27
Cozy-Sharings: Few enhancements 

Crash--:
- Better alignment 
before : 
Capture d’écran 2020-01-27 à 12 39 12
after: 
Capture d’écran 2020-01-27 à 12 39 09

- Add a tooltip info for SharedIcon and use coolGrey
Capture d’écran 2020-01-27 à 12 40 44

- Add a onFileDelete method to refresh after file deletion 
- Filter trashed document 
- Add revokeAllRecipient


packages/cozy-sharing/src/components/SharedStatus.jsx
```diff
@@ -27,7 +28,20 @@ export const SharedStatus = ({ className, docId, recipients, link, t }) => (
...
       </SharingTooltip>
     )}
-    {link && <LinkIcon />}
+    {link && (
+      <>
+        <span
+          style={{ fill: palette.coolGrey }}
+          data-tip
+          data-for={`linkfor${docId}`}
+        >
```
drazik: It is not possible to put these attributes on the LinkIcon?

packages/cozy-sharing/src/components/__snapshots__/SharedStatus.spec.js.snap
```diff
@@ -67,7 +67,7 @@ exports[`SharedStatus component should display the link if there is a link 1`] =
               wrapper="div"
             >
               <div
-                className="__react_component_tooltip place-bottom type-dark"
+                className="__react_component_tooltip place-bottom type-dark "
```
drazik: Why did it change?

Crash--: I think it depends on the nodejs version :/

packages/cozy-sharing/src/components/SharedStatus.jsx
Crash--: BG!

packages/cozy-sharing/src/components/__snapshots__/SharedStatus.spec.js.snap
drazik: OK. There's nothing blocking, I just wanted to ensure we don't miss something ;)

PR #927, opened 2020-01-27
feat: Add RefreshableSharings to expose a way to refresh sharings

Crash--:
In a few use case, we need to refresh the sharingProvider to update the infos about sharings. It's needed for instance when we delete a file or when we move a file. 

Since the stack does complicated stuffs for the sharings, we do not want ATM to mimic its behavior from a front point of view. So the refresh only call the fetch from the stack

### DIFF
--- a/packages/cozy-sharing/src/components/Recipient.jsx
+++ b/packages/cozy-sharing/src/components/Recipient.jsx
@@ -167,7 +167,7 @@ export class Status extends Component {
   }
 }
 
-const StatusWithBreakpoints = translate()(withBreakpoints()(withClient(Status)))
+const StatusWithBreakpoints = withBreakpoints()(translate()(withClient(Status)))
 
 const Recipient = (props, { client, t }) => {
   const { instance, isOwner, status, ...rest } = props
--- a/packages/cozy-sharing/src/index.jsx
+++ b/packages/cozy-sharing/src/index.jsx
@@ -58,7 +58,7 @@ const track = (document, action) => {
 const trackSharingByLink = document => track(document, 'shareByLink')
 const isFile = ({ _type }) => _type === 'io.cozy.files'
 
-class SharingProvider extends Component {
+export class SharingProvider extends Component {
   constructor(props, context) {
     super(props, context)
     const instanceUri = props.client.getStackClient().uri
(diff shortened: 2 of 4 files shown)


packages/cozy-sharing/src/components/Recipient.jsx
```diff
@@ -167,7 +167,7 @@ export class Status extends Component {
   }
 }
 
-const StatusWithBreakpoints = translate()(withBreakpoints()(withClient(Status)))
+const StatusWithBreakpoints = withBreakpoints()(translate()(withClient(Status)))
```
drazik: Nit: a lodash's `flowRight` can make it easier to read (and/or to add another HOC)

packages/cozy-sharing/src/index.spec.jsx
```diff
@@ -0,0 +1,35 @@
...
+import { RefreshableSharings, SharingProvider } from './'
+import AppLike from '../test/AppLike'
+
+const DumbComponent = () => {
+  return <div>test </div>
+}
+describe('RefreshableSharings', () => {
+  it('should test', () => {
+    const client = createMockClient({})
+    //client.getStackClient.mockResolvedValue({ uri: 'test' })
```
drazik:
```suggestion
```

packages/cozy-sharing/src/components/Recipient.jsx
y-lohse: Also nit but the change is… useless, no? 😄

packages/cozy-sharing/src/index.jsx
```diff
@@ -380,3 +380,16 @@ export const ShareModal = ({ document, ...rest }) => (
...
   </SharingContext.Consumer>
 )
+/**
+ * Expose a refresh method
+ */
+export const RefreshableSharings = ({ children, ...rest }) => (
+  <SharingContext.Consumer>
+    {({ refresh }) =>
+      children({
+        rest,
```
y-lohse: Did you mean `...rest`?
Also, I've never seen this, calling `children` as a function. Is there any doc about it that I missed? I only found [this](https://reactjs.org/docs/jsx-in-depth.html#functions-as-children) and there seems to be no guarantee that `children` is actually a function..?

Crash--: We already have `children()` in https://github.com/cozy/cozy-libs/blob/master/packages/cozy-sharing/src/components/SharedDocuments.jsx#L10 

It's just a `Function as Child Component` and it works well also with class component https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/drive/SharingsContainer.jsx#L171 

https://medium.com/merrickchristensen/function-as-child-components-5f3920a9ace9

In fact, I've no need of `rest`thanks :)

PR #929, opened 2020-01-28
feat: Add custom konnector policy

ptbrowne:
For the needs of the banking konnectors, a konnector must be able to insert itself
inside the connection process. A konnectorPolicy is introduced that can :

- insert custom behavior before account saving
- control whether credentials are saved into the vault

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -28,40 +28,82 @@ import VaultCiphersList from './VaultCiphersList'
 import manifest from '../helpers/manifest'
 import HarvestVaultProvider from './HarvestVaultProvider'
 import clone from 'lodash/clone'
+import flag from 'cozy-flags'
 
 import { createOrUpdateCipher } from '../models/cipherUtils'
+
+const defaultKonnectorPolicy = {
+  accountContainsAuth: true,
+  saveInVault: true,
+  onAccountCreation: null,
+  match: () => true
+}
+
+const policies = [defaultKonnectorPolicy]
 
 const IDLE = 'IDLE'
 const RUNNING = 'RUNNING'
 
 const MODAL_PLACE_ID = 'coz-harvest-modal-place'
 
+const findKonnectorPolicy = konnector =>
+  policies.find(policy => policy.match(konnector))
+
+/**
+ * Creates or updates an io.cozy.accounts
+ * Used as a form submit handler
+ *
+ * @param  {io.cozy.account} options.account - Existing io.cozy.account or object
+ * @param  {[type]} options.cipher - Vault cipher if vault has been unlocked
+ * @param  {CozyClient} options.client - A cozy client
+ * @param  {io.cozy.konnector} options.konnector - Konnector to which the account is linked
+ * @param  {KonnectorPolicy} options.konnectorPolicy - Controls if auth is saved in io.cozy.accounts
+ * and if auth is saved into the vault
+ * @param  {function} options.saveAccount
+ * @param  {object} options.userData
+ * @param  {function} options.ensureTrigger
+ */
 const createOrUpdateAccount = async ({
   account,
   cipher,
+  client,
   konnector,
+  konnectorPolicy,
   saveAccount,
-  userData
+  userCredentials
 }) => {
   const isUpdate = !!account
+  const { onAccountCreation, saveInVault } = konnectorPolicy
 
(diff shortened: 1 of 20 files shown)


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -28,40 +28,82 @@ import VaultCiphersList from './VaultCiphersList'
...
 
   accountToSave = isUpdate
-    ? accounts.mergeAuth(accountToSave, userData)
-    : accounts.build(konnector, userData)
+    ? accounts.mergeAuth(accountToSave, userCredentials)
+    : accounts.build(konnector, userCredentials)
+
+  if (onAccountCreation) {
+    accountToSave = await onAccountCreation({
+      client,
```
y-lohse: It seems strange to pass a cozy client here, shouldn't it be the responsability of the policy hook to get it if it needs it?

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -0,0 +1,129 @@
...
+ * @param  {io.cozy.konnectors} options.konnector
+ * @return {BIConnection}
+ */
+export const createOrUpdateBIConnection = async ({
+  account,
+  client,
+  konnector
+}) => {
+  try {
+    const mode = getBIModeFromCurrentLocation()
```
y-lohse: This will not work on mobile apps since the location on a mobile app is not the URL of the cozy. Is that a problem?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -27,41 +27,112 @@ import TriggerLauncher from './TriggerLauncher'
...
+    ? accounts.mergeAuth(accountToSave, userCredentials)
+    : accounts.build(konnector, userCredentials)
 
-  if (cipher) {
+  if (onAccountCreation) {
+    accountToSave = await onAccountCreation({
+      client,
+      account: accountToSave,
+      konnector,
+      saveAccount
```
y-lohse: I'm not sure I understand why we pass `saveAccount`, it's not really the policie's job to actually save the account?

packages/cozy-harvest-lib/src/services/biUtils.js
```diff
@@ -0,0 +1,102 @@
...
+      publicKey: biPublicKeyProd
+    }
+  } else {
+    return {
+      url: biURLDev
+    }
+  }
+}
+
+const biRequest = async (method, path, config, rawForm, bearer) => {
```
y-lohse: Not a big deal, but I'm surprised this isn't in a separate package. Do we not need these functions elsewhere?

ptbrowne: I extracted only a small part of the bi API from the banking konnectors. I only mutualized the encryption and translation of credential fields since otherwise we would have needed to copy/paste all the field definitions and this is subject to change. The bi request part is very stable and the code right now in the banking connectors is not suitable for being pruned (see https://gitlab.cozycloud.cc/labs/banking-konnectors/blob/master/src/bi/bi-api.js for all the supported APIs inside the banking konnector), this is why I recoded it here.

I agree with the premise here though, that a bi-sdk package could be of use. I a going to leave a comment.

packages/cozy-harvest-lib/src/services/budget-insight.js
ptbrowne: You're right, I will do that from the client uri !

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
ptbrowne: Get it from where ? As far as I know there is not global client object and we need it from a React component ?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
ptbrowne: For the budget-insight policy, we have to save the account prior to launching the temporary token job, since we have to have an account to launch a konnector job.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
y-lohse: Oh sorry, forgot to reply to this one. I don't see a way to get the client instance either past this point, so ok for me. Let's watch out for other cases with similar needs though.

PR #930, opened 2020-01-29
WIP: Upgrade UI and cozy-client

Crash--:
Seems I have a broken snapshots, I need to work on it.. Later.


ptbrowne:
You change from pinned versions to caret versions for dev dependencies. I am not sure we want to do that (it would be OK though for peer dependencies since we are dependent on the host environment) since change of version could mean change of snapshots for example.

Crash--:
Thanks. it is not intentional I didn't read the diff yet. But yep I have used `yarn add react@latest` and this shortcut add "^" version and not a pin like react@16.12.0

ptbrowne:
yarn add react@latest -E is what you want I guess ;)

PR #931, opened 2020-01-29
fix: Go directly to AccountForm if no need for Vault

ptbrowne:
The connection modal was blank otherwise

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -37,7 +37,8 @@ const defaultKonnectorPolicy = {
   accountContainsAuth: true,
   saveInVault: true,
   onAccountCreation: null,
-  match: () => true
+  match: () => true,
+  name: 'default'
 }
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -109,7 +115,7 @@ const createOrUpdateAccount = async ({
     )
   } else {
     // eslint-disable-next-line no-console
```
drazik: This can be removed since console has been replaced by logger

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -279,7 +290,7 @@ export class DumbTriggerManager extends Component {
         })
       } else {
         // eslint-disable-next-line no-console
```
drazik: This can also be removed

ptbrowne:
Thanks drazik I solved the issues and opened another PR with those two commits.

PR #933, opened 2020-01-30
[Harvest] feat: Links to other apps

y-lohse:
This PR allows harvest to offer links to Cozy Banks and Cozy Contacts when it detects that a konnector imports data for these apps (using the `data_types` field of the manifest.). I made a few other changes along the way, but the commits are explicit.

Capture d'écran 2020-01-30 06 54 49

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
@@ -4,7 +4,6 @@ import { withClient } from 'cozy-client'
 
 import Stack from 'cozy-ui/transpiled/react/Stack'
 import get from 'lodash/get'
-import has from 'lodash/has'
 
 import * as konnectorsModel from '../../helpers/konnectors'
 import KonnectorUpdateInfos from '../infos/KonnectorUpdateInfos'
@@ -14,7 +13,7 @@ import DocumentsLinkCard from '../cards/DocumentsLinkCard'
 import TriggerErrorInfo from '../infos/TriggerErrorInfo'
 import useMaintenanceStatus from '../hooks/useMaintenanceStatus'
 
-const DataTab = ({
+export const DataTab = ({
   konnector,
   trigger,
   error,
@@ -32,6 +31,8 @@ const DataTab = ({
     hasError &&
     !hasLoginError &&
     !isTermsVersionMismatchErrorWithVersionAvailable
+
+  const documentSaveFolderId = get(trigger, 'message.folder_to_save')
 
   const {
     data: { isInMaintenance, messages: maintenanceMessages }
@@ -54,8 +55,8 @@ const DataTab = ({
         <TriggerErrorInfo error={error} konnector={konnector} />
       )}
       <LaunchTriggerCard initialTrigger={trigger} disabled={isInMaintenance} />
-      {has(trigger, 'message.folder_to_save') && (
-        <DocumentsLinkCard folderId={get(trigger, 'message.folder_to_save')} />
+      {documentSaveFolderId && (
+        <DocumentsLinkCard folderId={documentSaveFolderId} />
       )}
     </Stack>
   )
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
@@ -9,7 +9,7 @@ import * as konnectorsModel from '../../helpers/konnectors'
 import KonnectorUpdateInfos from '../infos/KonnectorUpdateInfos'
 import LaunchTriggerCard from '../cards/LaunchTriggerCard'
 import KonnectorMaintenance from '../Maintenance'
-import DocumentsLinkCard from '../cards/DocumentsLinkCard'
+import AppLinkCard from '../cards/AppLinkCard'
 import TriggerErrorInfo from '../infos/TriggerErrorInfo'
 import useMaintenanceStatus from '../hooks/useMaintenanceStatus'
 
@@ -33,6 +33,10 @@ export const DataTab = ({
     !isTermsVersionMismatchErrorWithVersionAvailable
 
   const documentSaveFolderId = get(trigger, 'message.folder_to_save')
+  const showBanksLink = get(konnector, 'data_types', []).includes(
+    'bankAccounts'
+  )
+  const showContactsLink = get(konnector, 'data_types', []).includes('contact')
 
   const {
     data: { isInMaintenance, messages: maintenanceMessages }
@@ -56,7 +60,18 @@ export const DataTab = ({
       )}
       <LaunchTriggerCard initialTrigger={trigger} disabled={isInMaintenance} />
       {documentSaveFolderId && (
-        <DocumentsLinkCard folderId={documentSaveFolderId} />
+        <AppLinkCard
+          slug="drive"
+          path={`/files/${documentSaveFolderId}`}
+          icon="file"
+          iconColor="puertoRico"
+        />
+      )}
+      {showBanksLink && (
+        <AppLinkCard slug="banks" icon="bank" iconColor="weirdGreen" />
+      )}
+      {showContactsLink && (
+        <AppLinkCard slug="contacts" icon="team" iconColor="brightSun" />
       )}
     </Stack>
   )
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
@@ -12,6 +12,26 @@ import KonnectorMaintenance from '../Maintenance'
 import AppLinkCard from '../cards/AppLinkCard'
 import TriggerErrorInfo from '../infos/TriggerErrorInfo'
 import useMaintenanceStatus from '../hooks/useMaintenanceStatus'
+import relatedAppsConfiguration from '../../models/relatedAppsConfiguration'
+
+const appLinksProps = {
+  drive: ({ trigger }) => ({
+    slug: 'drive',
+    path: `/files/${get(trigger, 'message.folder_to_save')}`,
+    icon: 'file',
+    iconColor: 'puertoRico'
+  }),
+  banks: () => ({
+    slug: 'banks',
+    icon: 'bank',
+    iconColor: 'weirdGreen'
+  }),
+  contacts: () => ({
+    slug: 'contacts',
+    icon: 'team',
+    iconColor: 'brightSun'
+  })
+}
 
 export const DataTab = ({
   konnector,
@@ -32,11 +52,15 @@ export const DataTab = ({
     !hasLoginError &&
     !isTermsVersionMismatchErrorWithVersionAvailable
 
-  const folderToSaveId = get(trigger, 'message.folder_to_save')
-  const showBanksLink = get(konnector, 'data_types', []).includes(
-    'bankAccounts'
-  )
-  const showContactsLink = get(konnector, 'data_types', []).includes('contact')
+  const appLinks = relatedAppsConfiguration
+    .filter(app =>
+      app.predicate({
+        konnectorManifest: konnector,
+        trigger
+      })
+    )
+    .map(({ slug }) => appLinksProps[slug] && appLinksProps[slug]({ trigger }))
+    .filter(Boolean)
 
   const {
     data: { isInMaintenance, messages: maintenanceMessages }
@@ -59,20 +83,9 @@ export const DataTab = ({
         <TriggerErrorInfo error={error} konnector={konnector} />
       )}
       <LaunchTriggerCard initialTrigger={trigger} disabled={isInMaintenance} />
-      {folderToSaveId && (
-        <AppLinkCard
-          slug="drive"
-          path={`/files/${folderToSaveId}`}
-          icon="file"
-          iconColor="puertoRico"
-        />
-      )}
-      {showBanksLink && (
-        <AppLinkCard slug="banks" icon="bank" iconColor="weirdGreen" />
-      )}
-      {showContactsLink && (
-        <AppLinkCard slug="contacts" icon="team" iconColor="brightSun" />
-      )}
+      {appLinks.map(({ slug, ...otherProps }) => (
+        <AppLinkCard key={slug} slug={slug} {...otherProps} />
+      ))}
     </Stack>
   )
 }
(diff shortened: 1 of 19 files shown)


packages/cozy-harvest-lib/test/components/__snapshots__/AccountModal.spec.jsx.snap
```diff
@@ -57,7 +57,7 @@ exports[`AccountModal with an account should display the AccountSelect & Content
     />
   </KonnectorModalHeader>
   <ModalContent>
-    <withClient(withI18n(KonnectorAccountTabs))
+    <withI18n(KonnectorAccountTabs)
```
Crash--: I don't understand this change. You didn't edit AccoutModal.jsx, it's weird, no?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
```diff
@@ -56,7 +60,18 @@ export const DataTab = ({
...
+          path={`/files/${documentSaveFolderId}`}
+          icon="file"
+          iconColor="puertoRico"
+        />
+      )}
+      {showBanksLink && (
+        <AppLinkCard slug="banks" icon="bank" iconColor="weirdGreen" />
+      )}
+      {showContactsLink && (
+        <AppLinkCard slug="contacts" icon="team" iconColor="brightSun" />
```
Crash--: Should we create a `ShowLinks` component that handle the several get('data_types') and returns the right components by itself? 

```
<>
 <LaunchTriggerCard initialTrigger={trigger} disabled={isInMaintenance} />
<ShowLinks trigger={trigger}  konnector={konnector} />
</>
```

it'll be easier to maintain, no?

packages/cozy-harvest-lib/src/components/cards/AppLinkCard.jsx
```diff
@@ -44,7 +45,8 @@ const AppLinkCard = ({
           href={generateWebLink({
             cozyUrl: cozyURL.origin,
             slug,
-            nativePath: path
+            nativePath: path,
+            subDomainType
           })}
         >
           {({ onClick, href }) => (
```
Crash--: Should we add a test for this component ?

packages/cozy-harvest-lib/test/components/__snapshots__/AccountModal.spec.jsx.snap
y-lohse: No but I changed `KonnectorAccountTabs` and removed the `withClient` there. Same for the other one :)

packages/cozy-harvest-lib/test/components/KonnectorConfiguration/DataTab.spec.jsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import { DataTab } from 'components/KonnectorConfiguration/DataTab'
+import LaunchTriggerCard from 'components/cards/LaunchTriggerCard'
+import DocumentsLinkCard from 'components/cards/DocumentsLinkCard'
+
+describe('DataTab', () => {
+  beforeEach(() => {
+    jest.spyOn(console, 'error').mockImplementation(() => {})
```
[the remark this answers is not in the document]
y-lohse: Failing PropTypes because of the testing context, I'll add a comment.

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
y-lohse: I considered it, but since some use `konnector` and some use `trigger`, I thought this abstraction would be too much and actually makes things less maintainable.

packages/cozy-harvest-lib/test/components/KonnectorConfiguration/DataTab.spec.jsx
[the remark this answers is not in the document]
Crash--: Can't we fix the propType issue instead of hiding it?

packages/cozy-harvest-lib/src/components/cards/AppLinkCard.jsx
y-lohse: i don't really see what could be tested here except a snapshot of the component. And since it's a very straightforward component with no logic I don't think it's useful, I thought the important test to add was checking that the links appear as we want.

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
```diff
@@ -33,6 +32,8 @@ const DataTab = ({
     !hasLoginError &&
     !isTermsVersionMismatchErrorWithVersionAvailable
 
+  const documentSaveFolderId = get(trigger, 'message.folder_to_save')
```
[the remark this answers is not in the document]
ptbrowne: foldertoSaveId is even better IMHO, I am always wondering if the values I am treating are id or full objects.

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
ptbrowne: I am not sure the component approach is best but I think it would be interesting to extract some of the logic from the render, and have it more declarative.

```
const appConfig = {
  banks: {
    shouldBeShown: ({ konnector }) => blabla,
    icon: 'banks'
    iconColor: 'weirdGreen'
  },
  ...
}

const appOrder = ['drive', 'banks', 'contacts']
```

I think it makes the intention clearer and the config can be used elsewhere easily.

Crash--: I'm pretty sure we will add links to linked apps on other pages / components or even apps. 

Having at least a hook that handle the logic to check "bankAccount" or "folder_to_save" or "contacts" can be very useful. 

```
const linkedApps = useLinkedApps({konnector, trigger})

linkedApp.banks && <AppLinkCard slug="banks">
```

Also, I'm wondering if `AppLinkCard` should not be moved to UI with that in mind

ptbrowne: A mon avis ca devrait surement etre mutualisé avec ce qu il y a ici : https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/components/KonnectorSuccess.jsx

ptbrowne: IMHO, no need for a hook, `useLinkedApps` could be `getLinkedApps` (not really linked ? `getUsefulApps` ?), and this function could use a declarative config.

Crash--: > AMHA that should be mutualized with what already exists for the success message : https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/components/KonnectorSuccess.jsx 

I knew it we already have something similar! Thank you!

packages/cozy-harvest-lib/test/components/KonnectorConfiguration/DataTab.spec.jsx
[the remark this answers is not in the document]
y-lohse: The issue is that a child component (not even the component we're testing here) needs a translation context, and that would require using `mount` instead of `shallow`, so needing to mock a lot more things.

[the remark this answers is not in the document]
ptbrowne: In Banks, it is common to wrap our components into a I18n component, it also helps since snapshots have the right translations inside them.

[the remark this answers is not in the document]
y-lohse: But then you have to use `mount` and not `shallow`, right?

(no snapshots here 😃)

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
y-lohse: Ah I forgot we had this. Yeah that calls for some form of mutualization then, I'll have a go at this tomorrow.

y-lohse: I created a `getRelatedAppsSlugs` that holds only the logic and is used by both `KonnectorSuccess` and `DataTab`. Let me know what you think!

packages/cozy-harvest-lib/src/components/KonnectorSuccess.jsx
```diff
@@ -40,12 +39,12 @@ export class KonnectorSuccess extends Component {
...
-        app.predicate(this.state, this.props)
-      ),
-      app => -app.priority
-    )
+    const relatedApps = getRelatedAppsSlugs({
+      konnectorManifest: this.props.konnector,
+      trigger: this.state.trigger
+    })
+      .map(slug => KonnectorSuccess.apps[slug])
+      .filter(Boolean)
```
Crash--: We lose the notion of priority here. It was needed for our AF project. We should certainly keep it, no?

In fact, I think we have it also on the `DataTabs` but it is implicit right? We should keep it explicit

packages/cozy-harvest-lib/test/components/KonnectorConfiguration/DataTab.spec.jsx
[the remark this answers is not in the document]
Crash--: Personally I prefer the mount versus the hide of errors

packages/cozy-harvest-lib/src/components/KonnectorSuccess.jsx
y-lohse: The priority is now determined by the order in the configuration array rather than a `priority` number. I figured we could swap the array order around in our AF projects if we want a different order there. I don't mind putting the number system back in if you think this is not explicit enough.

Crash--: I would prefer yep. Explicit is way better :D.

Also can we extract the `appLinksProps` in a new file since we already now we have to override it? :D

packages/cozy-harvest-lib/src/models/getRelatedAppsSlugs.js
```diff
@@ -21,15 +24,21 @@ const relatedAppsConfiguration = [
...
-    .filter(app => app.predicate({ konnectorManifest, trigger }))
-    .map(({ slug }) => slug)
+const getRelatedAppsSlugs = ({ konnectorManifest, trigger }) => {
+  const matchingApps = relatedAppsConfiguration.filter(app =>
+    app.predicate({ konnectorManifest, trigger })
+  )
+
+  return sortBy(matchingApps, ({ priority }) => -priority).map(
+    ({ slug }) => slug
+  )
```
ptbrowne: Nit: I wonder if this function could let the caller extract the slugs ? I think it would be a bit mor flexible. Anyway this would easily be refactorable.

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx
```diff
@@ -87,83 +52,25 @@ const KonnectorAccountTabs = ({
...
-                    initialTrigger={initialTrigger}
-                    disabled={isInMaintenance}
-                  />
-                  {has(initialTrigger, 'message.folder_to_save') && (
-                    <DocumentsLinkCard
-                      folderId={get(initialTrigger, 'message.folder_to_save')}
-                    />
-                  )}
-                </Stack>
+                <DataTab
```
ptbrowne: Nit: The `Tab` components represents the buttons to change tab, I think we should have either DataTabPanel or DataTabContent.

packages/cozy-harvest-lib/src/components/infos/KonnectorUpdateInfos.jsx
```diff
@@ -15,21 +16,27 @@ export const KonnectorUpdateInfos = props => {
...
+            {t('infos.konnectorUpdate.title')}
+          </SubTitle>
+          <Text>
+            {isBlocking
+              ? t('infos.konnectorUpdate.body.blocking')
+              : t('infos.konnectorUpdate.body.regular')}
+          </Text>
+        </>
+      }
+      action={
```
ptbrowne: :+1:

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab.jsx
```diff
@@ -59,10 +59,10 @@ export const DataTab = ({
         <TriggerErrorInfo error={error} konnector={konnector} />
       )}
       <LaunchTriggerCard initialTrigger={trigger} disabled={isInMaintenance} />
-      {documentSaveFolderId && (
+      {folderToSaveId && (
         <AppLinkCard
           slug="drive"
-          path={`/files/${documentSaveFolderId}`}
+          path={`/files/${folderToSaveId}`}
```
ptbrowne: Nit: I would prefer this commit to be fixed up with the commit introducing the wrong variable name.

packages/cozy-harvest-lib/test/components/KonnectorConfiguration/DataTab.spec.jsx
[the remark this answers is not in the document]
ptbrowne: > But then you have to use mount and not shallow, right?

Why ?

I agree with @Crash-- here.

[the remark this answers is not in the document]
y-lohse: If I do `shallow(<I18n><MyComponent /></I18n>)` I need to `dive`, and thenthe proptype errors appear again. Same goes if I provide a `wrappingComponent` to `shallow`.

> Personally I prefer the mount versus the hide of errors

Mounting on such a high level component is a bit of a nightmare, really. If the errors came from the component we're testing I would be on board with this, but they aren't.

packages/cozy-harvest-lib/src/models/getRelatedAppsSlugs.js
y-lohse: Yeah I changed it because the `predicate` and `priority` don't have a reason to leak outside of this file, but if there are other things we need to expose it would be easy to change :)

packages/cozy-harvest-lib/test/components/KonnectorConfiguration/DataTab.spec.jsx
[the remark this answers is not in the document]
ptbrowne: Is it possible to do

```
jest.mock('ChildComponent', () => () => null) non ?
```


The issue is that a child component (not even the component we're testing here) needs a translation context, and that would require using mount instead of shallow, so needing to mock a lot more things.

PR #936, opened 2020-01-30
Merge pull request #936 from cozy/convert-bi-errors

ptbrowne:
Convert BI errors to user understandable errors through KonnectorJobError.


packages/cozy-harvest-lib/package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@babel/runtime": "^7.5.2",
     "@material-ui/core": "3",
-    "cozy-bi-auth": "0.0.2",
+    "cozy-bi-auth": "0.0.3",
```
ptbrowne: Now cozy-bi-auth shares the mapping between cozy errors and BI errors.

packages/cozy-harvest-lib/src/models/cipherUtils.js
```diff
@@ -113,7 +113,9 @@ export const createOrUpdateCipher = async (
     })
   }
 
-  await shareCipherWithCozy(vaultClient, cipher.id)
+  if (cipher) {
+    await shareCipherWithCozy(vaultClient, cipher.id)
+  }
```
Crash--: Is it possible to have no `cipher` here? How since we create it or update it. Is it to prevent an error during those 2 methods? 

```


  if (cipherId) {
    cipher = await updateCipher(vaultClient, cipherId, { login, password })
  } else {
    cipher = await createCipher(vaultClient, {
      konnector,
      login,
      password
    })
  }

if(cipher)... 
```

ptbrowne: If the vault is unlocked, no cipher can be created. I have a problem here though since updateCipher does check if the vault is locked or not (contrary to createCipher). I am moving the check in this function and adding tests. Thanks for the catch.

packages/cozy-harvest-lib/src/models/cipherUtils.js
```diff
@@ -83,13 +78,35 @@ export const updateCipher = async (vaultClient, cipherId, data) => {
...
+ * @param  {io.cozy.konnector} options.konnector
+ * @return {Cipher|null} - Returns null if vault is locked
+ */
 export const createOrUpdateCipher = async (
   vaultClient,
   cipherId,
   { userCredentials, account, konnector }
 ) => {
+  if (vaultClient.isLocked()) {
+    // eslint-disable-next-line no-console
```
drazik:
```suggestion
```

PR #940, opened 2020-01-31
Small fixes

ptbrowne:
- Consider only "result" event when waiting for temporary token.

Since we will receive other events from the konnector in the future,
for example "progress" events, we filter on the "type" attribute 
received in the "data" part of the event.

- Send the bankId as part of the temporary token job options. This is
necessary for multi branch banks, where the bankId is not baked in
the manifest.konnector.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
@@ -4,15 +4,15 @@ import { withClient } from 'cozy-client'
 
 import Stack from 'cozy-ui/transpiled/react/Stack'
 
-import * as konnectorsModel from 'helpers/konnectors'
-import KonnectorUpdateInfos from 'components/infos/KonnectorUpdateInfos'
-import LaunchTriggerCard from 'components/cards/LaunchTriggerCard'
-import KonnectorMaintenance from 'components/Maintenance'
-import AppLinkCard from 'components/cards/AppLinkCard'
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
```diff
@@ -4,15 +4,15 @@ import { withClient } from 'cozy-client'
...
-import appLinksProps from 'components/KonnectorConfiguration/DataTab/appLinksProps'
+import * as konnectorsModel from '../../../helpers/konnectors'
+import KonnectorUpdateInfos from '../../../components/infos/KonnectorUpdateInfos'
+import LaunchTriggerCard from '../../../components/cards/LaunchTriggerCard'
+import KonnectorMaintenance from '../../../components/Maintenance'
+import AppLinkCard from '../../../components/cards/AppLinkCard'
+import TriggerErrorInfo from '../../../components/infos/TriggerErrorInfo'
+import useMaintenanceStatus from '../../../components/hooks/useMaintenanceStatus'
+import getRelatedAppsSlugs from '../../../models/getRelatedAppsSlugs'
+import appLinksProps from '../../../components/KonnectorConfiguration/DataTab/appLinksProps'
```
drazik: I don't understand why we need to use relative paths here :thinking:

ptbrowne: Because otherwise we would try to resolve node_modules/models, node_modules/components etc..

PR #941, opened 2020-01-31
fix: Submitting button returns to idle state after error

ptbrowne:
The submit button for account creation  would not return to idle state
if there was an error.

### DIFF
--- a/packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
+++ b/packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
@@ -432,6 +432,22 @@ describe('TriggerManager', () => {
       await submitPromise
     })
 
+    it('should stop being rendered as submitting on error', async () => {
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
```diff
@@ -432,6 +432,20 @@ describe('TriggerManager', () => {
...
 
+    it('should stop being rendered as submitting on error', async () => {
+      const wrapper = shallowWithAccount()
+      mockVaultClient.isLocked.mockReset().mockImplementationOnce(() => {
+        throw new Error('fakeerror')
+      })
+
+      try {
+        await wrapper.instance().handleSubmit(fixtures.data)
+        expect(mockVaultClient).toHaveBeenCalledTimes(1)
```
drazik: I don't understand why you check this? Did you want to check that `mockVaultClient.isLocked` has been called?

PR #942, opened 2020-01-31
feat(cozy-harvest-lib): Add deleteAccounts service

drazik:
This finds the accounts that have a relationship with a given cipher,
and delete it. It will be used in a service to delete accounts when a
cipher has been deleted by the user.

### DIFF
--- a/packages/cozy-harvest-lib/src/services/deleteAccounts.js
+++ b/packages/cozy-harvest-lib/src/services/deleteAccounts.js
@@ -1,0 +1,22 @@
+import get from 'lodash/get'
+import { fetchAccountsForCipherId } from './utils'
+import logger from '../logger'
+import { deleteAccount } from '../connections/accounts.js'
+
+const deleteAccounts = async (cozyClient, bitwardenCipherDocument) => {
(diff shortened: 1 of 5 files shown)


packages/cozy-harvest-lib/src/services/deleteAccounts.js
```diff
@@ -0,0 +1,13 @@
+import get from 'lodash/get'
+
+import { fetchAccountsForCipherId } from './utils'
+
+const deleteAccounts = async (cozyClient, bitwardenCipherDocument) => {
+  const bitwardenCipherId = get(bitwardenCipherDocument, '_id')
+
+  const accounts = await fetchAccountsForCipherId(cozyClient, bitwardenCipherId)
+
+  await Promise.all(accounts.map(account => cozyClient.destroy(account)))
```
ptbrowne: As a service I think it would be interesting to add logging ? The minilog instance in src/logger could help ?

drazik: Yep good idea, thanks. See https://github.com/cozy/cozy-libs/pull/942/commits/c89c753f76477ef0240a28decaf716096b1053e1

PR #949, opened 2020-02-06
fix(cozy-harvest-lib): Always show account form when editing an account

drazik:
When the user tries to edit an account, we show the form directly. But
ciphers are being fetched in background, and we were redirecting to the
ciphers list step as soon as we detect that some cipher exists.
Actually, when the user is on the account form step, we shouldn't
automatically redirect. If the user wants to see the ciphers list, he
can by clicking the "back" arrow.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -419,13 +419,12 @@ export class DumbTriggerManager extends Component {
   }
 
   showCiphersList(ciphers) {
-    const newState = { step: 'ciphersList' }
-
-    if (ciphers) {
-      newState.ciphers = ciphers
-    }
-
-    this.setState(newState)
+    this.setState(prevState => {
+      return {
+        step: prevState.step === 'accountForm' ? 'accountForm' : 'ciphersList',
+        ciphers: ciphers ? ciphers : prevState.ciphers
+      }
+    })
   }
 
   async handleVaultUnlock() {
--- a/packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
+++ b/packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
@@ -364,6 +364,48 @@ describe('TriggerManager', () => {
       const component = shallowWithAccount().getElement()
       expect(component).toMatchSnapshot()
     })
+
+    describe('when the vault contains ciphers', () => {
+      it('should show the account form', async () => {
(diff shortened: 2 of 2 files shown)


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -446,10 +446,15 @@ export class DumbTriggerManager extends Component {
...
-        this.showCiphersList(ciphers)
+      // When we fetched the ciphers while the user is already on the account
+      // form step, we don't want to automatically redirect him on the ciphers
+      // list
+      if (this.state.step !== 'accountForm') {
+        if (ciphers.length === 0) {
+          this.showAccountForm()
+        } else {
+          this.showCiphersList(ciphers)
+        }
```
Crash--: This is a good use case for a more declarative code ? :D

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -446,10 +446,15 @@ export class DumbTriggerManager extends Component {
...
       })
 
-      if (ciphers.length === 0) {
-        this.showAccountForm()
-      } else {
-        this.showCiphersList(ciphers)
+      // When we fetched the ciphers while the user is already on the account
+      // form step, we don't want to automatically redirect him on the ciphers
+      // list
+      if (this.state.step !== 'accountForm') {
```
ptbrowne: Instead of doing a negative boolean check, which will fail as soon as a step is added, could we reverse the condition to be positive ? Which step are we assuming here ?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
drazik: I'm not sure I understand your point. Do you mean it should be handled via components rendered?

drazik: Just discussed it with @ptbrowne. He is going to refactored all this in a more declarative way while adding 2FA feature for banking konnectors. Since here I added missing test cases, we merge this PR so he has the tests to allow him to refactor easily.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
drazik: See https://github.com/cozy/cozy-libs/commit/ba0469679de1962bd9d2505d53479d0a985ae81d

PR #950, opened 2020-02-06
[cozy-harvest-lib] Don't duplicate ciphers when selecting an existing one

drazik:
When creating a new account with an existing cipher, it was duplicated. This PR fixes this. Now, the cipher is linked to the created account, but not duplicated.

Also added tests.

### DIFF
--- a/packages/cozy-harvest-lib/src/models/cipherUtils.spec.js
+++ b/packages/cozy-harvest-lib/src/models/cipherUtils.spec.js
@@ -1,28 +1,15 @@
-import { createOrUpdateCipher } from './cipherUtils'
+import * as cipherUtils from './cipherUtils'
 
 describe('createOrUpdateCipher', () => {
   const setup = ({
-    account: accountAttrs,
     konnector: konnectorAttrs,
     userCredentials: userCredentialsAttrs,
     vaultClient: vaultClientAttrs
   } = {}) => {
-    const account = {
-      relationships: {
-        vaultCipher: {
-          data: [
-            {
-              _id: 'cipher-relationship-id'
-            }
-          ]
-        }
-      },
-      ...accountAttrs
-    }
-
     const konnector = {
       vendor_link: 'konnector-vendor-link',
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/models/cipherUtils.spec.js
```diff
@@ -32,79 +19,137 @@ describe('createOrUpdateCipher', () => {
...
+              match: 'Domain',
+              uri: 'konnector-vendor-link'
+            }
+          ],
+          username: 'login-to-be-updated'
+        },
+        name: 'konnector-name',
+        organizationId: 'cozy-org-id',
+        type: 'Login'
+      })
```
ptbrowne: Seems this is the same object as in above tests. This could be extracted and reused here. This would make it easier to understand what changes from test to test (I had to compare key by key this object with the other one above otherwise).

packages/cozy-harvest-lib/src/models/cipherUtils.spec.js
```diff
@@ -32,79 +19,137 @@ describe('createOrUpdateCipher', () => {
...
+
+        const cipherId = null
+
+        const cipher = await cipherUtils.createOrUpdateCipher(
+          vaultClient,
+          cipherId,
+          { userCredentials, account, konnector }
+        )
+
+        expect(cipher).toMatchSnapshot()
```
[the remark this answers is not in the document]
drazik: I used expect without snapshots in https://github.com/cozy/cozy-libs/commit/3d2117605f016a81e9ae79e58937c4b91abb6516, it's clearer :+1:

packages/cozy-harvest-lib/src/models/cipherUtils.spec.js
drazik: I put it in a `existingCipher` variable, so it's clearer :+1: See https://github.com/cozy/cozy-libs/commit/3d2117605f016a81e9ae79e58937c4b91abb6516

PR #955, opened 2020-02-07
feat(cozy-harvest-lib): Add logs to updateAccountsPassword service

drazik:
This service lacked some debug logs. This PR adds some.

### DIFF
--- a/packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
+++ b/packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
@@ -9,6 +9,8 @@ import {
   launchTriggers
 } from './utils'
 
+import logger from './logger'
+
 const updateAccountsPassword = async (
   cozyClient,
   vaultClient,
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/services/updateAccountsPassword.js
```diff
@@ -9,6 +9,10 @@ import {
   launchTriggers
 } from './utils'
 
+import logger from 'cozy-logger'
+
+logger.namespace('cozy-harvest-lib')
```
ptbrowne: There is a src/logger file that exports a logger is the right namepsace.

drazik: Yes but here since we are in a service, we need to handle the stack's log format. That's why I used cozy-logger. Does minilog do it?

ptbrowne: Ah yes sorry, I think we could tell minilog to do it yes. At first, taking it from src/services/logger and explaining this could do the trick.

drazik: See https://github.com/cozy/cozy-libs/commit/c368afa992f70bce622112dbd9897c2bd39c7693

PR #956, opened 2020-02-11
Support for 2FA and loginSuccess via custom konnector policy

ptbrowne:
A custom konnector policy should be able to trigger a loginSuccess
to tell the UI that logins has been successful and that we should
not let a loading spinner continue.

Previously this was difficult to achieve since the business logic
for a connection was scattered between two components:

- TriggerManager would start the account creation (also handling
  cipher and trigger creation)
- TriggerLauncher would launch the job and listen to the KonnectorJob
  object

Since the loginSuccess and 2FA events were previously sent in the
job and could now be sent at account creation, logic should be shared
between TriggerManager and TriggerLauncher. This pushed the opportunity
to gather the whole connection creation (and update) logic into a
single object. KonnectorJob already listened to the job, it now takes
care of listening to the account, refetches the trigger when needed,
essentially managing the whole connection creation state from end to end. It is renamed to ConnectionFlow in the end.

On any update, it signals to components that they should update, and it
gives them a mean to access its state (+ derivated values).

While developing, I found it handy to have a CLI to execute the form
submitting, instead of using the web UI. It forced a cleaner separation
of concerns between UI and business logic, and also made sure that the
KonnectorJob/ConnectionFlow object was flexible enough to accomodate both
web and CLI usages.

In the future, it will be more straightforward to add UI components
listening to the ConnectionFlow object to show the progress for example
and react to realtime "notified" messages from the konnector.


packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -204,6 +204,7 @@ export class DumbTriggerManager extends Component {
   async handleSubmit(data = {}) {
     const { client, konnector, saveAccount, vaultClient } = this.props
     const { account } = this.state
+    const cipherId = this.getSelectedCipherId()
```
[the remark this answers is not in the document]
ptbrowne: >  Il y a-t-il un risque de faire une sélection à un moment

Je ne comprends pas "faire une sélection" ? S'il n'y a pas de cipher selectionné, `this.getSelectedCipherId()` renverra undefined et c'est supporté dans le KonnectorJob.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -319,10 +252,12 @@ export class DumbTriggerManager extends Component {
    * @return {Object}          io.cozy.jobs document, runned with account data
    */
   async handleNewAccount(account) {
-    const trigger = await this.ensureTrigger(account)
+    const { client, trigger, konnector } = this.props
+    const trigger = await ensureTrigger(client, { trigger, account, konnector })
```
drazik: I don't see `ensureTrigger` import. Is it missing?
(I bet it's fixed in a later commit, I'm currently reviewing https://github.com/cozy/cozy-libs/pull/956/commits/7cfcfa5e969172cceb2a17ac77458e7119cd380d)

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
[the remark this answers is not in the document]
drazik: I think it doesn't change anything. `this.getSelectedCipherId()` will return the same thing, no matter if it's called in its previous place or here. So since the usage is still the same, it's OK.

packages/cli-tree/src/index.js
```diff
@@ -0,0 +1,73 @@
...
+      }
+    }
+  })
+
+  return [parser, subparsers]
+}
+
+module.exports = {
+  build
+}
```
drazik: Just to be sure: that's the one from banking-konnector that you extracted?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -114,7 +112,11 @@ export class DumbTriggerManager extends Component {
     try {
       this.setState({ error: null, status: RUNNING })
       const oAuthAccount = await findAccount(client, accountId)
-      return await this.handleNewAccount(oAuthAccount)
+      return await flow.ensureTriggerAndLaunch(client, {
```
drazik: Were does `flow` come from? Shouldn't it be extracted from `this.props`?
(Maybe fixed later, currently reviewing https://github.com/cozy/cozy-libs/pull/956/commits/f35bfa35574945ea2412d663165b13cc0fea077f)

packages/cozy-harvest-lib/src/models/KonnectorJob.js
```diff
@@ -34,6 +34,21 @@ export const RUNNING_TWOFA = 'RUNNING_TWOFA'
...
+  SUCCESS_EVENT
+]
+
+const eventToStatus = {
+  [ERROR_EVENT]: ERRORED,
+  [LOGIN_SUCCESS_EVENT]: LOGIN_SUCCESS,
+  [SUCCESS_EVENT]: SUCCESS
+}
+const stepEvents = [LOGIN_SUCCESS_EVENT]
+const isStatusEvent = eventName => Boolean(jobEventToStatus[eventName])
```
drazik: Shouldn't it be `eventToStatus[eventName]`?
(maybe fixed later, I'm reviewing https://github.com/cozy/cozy-libs/pull/956/commits/3a1af74c27c412f4df11b31327bbee221c074c4d)

packages/cozy-harvest-lib/src/models/KonnectorJob.js
```diff
@@ -49,15 +49,6 @@ const stepEvents = [LOGIN_SUCCESS_EVENT]
 const isStatusEvent = eventName => Boolean(jobEventToStatus[eventName])
 const isStepEvent = eventName => stepEvents.includes(eventName)
 
-export const watchKonnectorJob = (client, job) => {
```
drazik: It's used nowhere else? I'm wondering as it was exported, and it's not anymore.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -147,10 +181,9 @@ export class DumbTriggerManager extends Component {
       const cipherId = this.getSelectedCipherId()
       await flow.handleFormSubmit({
         client,
-        account,
+        account: account || {},
```
drazik: if `flow.handleFormSubmit` need account to not be undefined, shouldn't it have a default parameter instead of forcing its caller to take care of passing it an empty object?

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -68,6 +69,19 @@ export const getTwoFACodeProvider = account => {
...
+export const updateTwoFAState = (account_, { retry, type }) => {
+  const account = clone(account_)
+  let state = retry ? 'TWOFA_NEEDED_RETRY' : 'TWOFA_NEEDED'
+  if (type === 'email') {
+    state += '.EMAIL'
+  } else if (type === 'sms') {
+    state += '.SMS'
+  } else if (type === 'app') {
+    state += '.APP'
+  }
```
drazik: Nit: if the type is not handled, we have an implicit default case here. Maybe we can handle it explicitly:

```js
const typesToStateFragment = {
  email: 'EMAIL',
  sms: 'SMS',
  app: 'APP'
}

const stateFragment = typesToStateFragment[type]

state += stateFragment ? `.${stateFragment}` : ''
```

packages/cozy-harvest-lib/src/models/KonnectorJob.js
```diff
@@ -134,24 +141,113 @@ export class KonnectorJob {
...
+    this.triggerEvent(TWO_FA_REQUEST_EVENT, twoFARequestOptions)
+    try {
+      const account = accounts.updateTwoFAState(
+        this.account,
+        twoFARequestOptions
+      )
+      await this.saveAccount(account)
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error)
```
drazik: Don't you want to use the logger here?

packages/cozy-harvest-lib/src/models/KonnectorJob.js
```diff
@@ -134,24 +141,113 @@ export class KonnectorJob {
...
+      updatedAccount
+    )
+    logger.info('Saved account')
+    return this.account
+  }
+
+  flushTwoFAWaiters() {
+    logger.debug(`Flushing ${this.twoFAWaiters.length} two fa waiters`)
+    for (const w of this.twoFAWaiters) {
+      w()
```
drazik: Nit: I think renaming `w` to `waiter` would make reading easier

packages/cozy-harvest-lib/src/models/KonnectorJob.js
```diff
@@ -134,24 +141,113 @@ export class KonnectorJob {
...
+      this.flushTwoFAWaiters()
+    } catch (error) {
+      console.error(error)
+      this.setState({ status: ERRORED, error })
+    }
+  }
+
+  /**
+   * "Sends" 2FA Code
+>>>>>>> theirs
```
drazik:
```suggestion

```

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -172,34 +197,142 @@ const removeSensibleDataFromAccount = fullAccount => {
   return account
 }
 
+export const setBIConnectionId = (account_, biConnectionId) => {
+  const account = clone(account_)
```
drazik: Nit: renaming `account_` to `originalAccount` would be clearer

packages/cli-tree/src/index.js
ptbrowne: Yes.

packages/cozy-harvest-lib/src/models/KonnectorJob.js
ptbrowne: Renamed to callback since it is called.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -0,0 +1,529 @@
...
+
+    this.state = {
+      status: IDLE
+    }
+
+    // Stores data necessary for custom connection flow, for example
+    // a budget-insight temporary token
+    this.data = {}
+
+    this.twoFAWaiters = this.twoFAWaiters || []
```
y-lohse: How could `this.twoFAWaiters` be already defined?

packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
```diff
@@ -55,7 +54,7 @@ export class DeleteAccountButton extends Component {
         busy={deleting}
         onClick={this.handleDelete}
         label={t('accountForm.disconnect.button')}
-        {...omit(this.props, Object.keys(excludedButtonProptypes))}
+        {...omit(this.props, Object.keys(excludedButtonProptypes), 'client')}
```
y-lohse: I don't think `omit` supports a  third argument? https://lodash.com/docs/4.17.15#omit

packages/cozy-harvest-lib/src/components/KonnectorAccounts.jsx
```diff
@@ -8,17 +8,18 @@ import { translate } from 'cozy-ui/transpiled/react/I18n'
 import Infos from 'cozy-ui/transpiled/react/Infos'
 import Button from 'cozy-ui/transpiled/react/Button'
 import get from 'lodash/get'
+import compose from 'lodash/flowRight'
```
y-lohse: Nit but I find it strange to rename `flowRight` to `compose`

packages/cozy-harvest-lib/src/components/KonnectorAccounts.jsx
```diff
@@ -79,19 +84,14 @@ export class KonnectorAccounts extends React.Component {
...
-            trigger
-          }
-        })
-      )).filter(({ account }) => !!account)
-      this.setState({ accounts, error: null })
+      const triggerAccounts = await fetchAccountsFromTriggers(client, triggers)
+      this.setState({ accounts: triggerAccounts, error: null })
     } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error)
```
y-lohse: Why not use `logger` here?

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
```diff
@@ -226,8 +190,4 @@ TriggerLauncher.propTypes = {
   initialTrigger: PropTypes.object
 }
 
-export default translate()(
-  withClient(
-    withMutations(accountsMutations, triggersMutations)(TriggerLauncher)
-  )
-)
+export default translate()(withClient(TriggerLauncher))
```
y-lohse: Should `TriggerLauncher ` be renamed, as it's not really launching the trigger anymore?

packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
ptbrowne: It does 

```
require('lodash/omit')({ a: 1, b: 2}, ['a'], 'b')
```

The docs are not very clear but it is there ;)

> [paths] (...(string|string[])): The property paths to omit.

packages/cozy-harvest-lib/src/components/KonnectorAccounts.jsx
ptbrowne: Compose is the original name of this function. It was renamed into flowRight for parity with the flowLeft function in later versions of lodash (in lodashFP, it is named compose BTW). Personally I find it better named as compose as it makes reference to [function composition](https://en.wikipedia.org/wiki/Function_composition).

packages/cozy-harvest-lib/src/components/DeleteAccountButton.jsx
y-lohse: TIL!

packages/cli-tree/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@cozy/cli-tree",
```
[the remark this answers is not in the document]
ptbrowne: Fixed

packages/cozy-harvest-lib/src/components/TriggerLauncher.jsx
ptbrowne: Yes. FlowProvider ?

y-lohse: Sounds good to me!

packages/cozy-harvest-lib/src/components/KonnectorAccounts.jsx
y-lohse: > I use this name since the beginning I use compose at Cozy. I thought it was understood as a convention.

I know and it's understood, no worries. I just hadn't come across it in a while so I thought I mention it in passing.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
ptbrowne: If the function has already been called once.

packages/cozy-harvest-lib/test/__mocks__/cozy-keys-lib/index.js
```diff
@@ -18,6 +18,7 @@ const UriMatchType = {
   Domain: 'Domain'
 }
 
+
```
drazik: Looks like this "wip" commit can be removed (https://github.com/cozy/cozy-libs/pull/956/commits/2103878394043bf5ce661167bdfa08391c37d32f)

PR #959, opened 2020-02-17
Publication channel can depend on registry channel

ptbrowne:
## Need

Following a discussion on mattermost where it was reminded that publication messages should be published on the Publication channel.

For beta and stable channels, it is useful to have them published
on the publication channel. It remains useful to also have them
in the app channel (for dev and PO following the channel of the app).

Dev versions on the contrary should not be published on the publication channel as it would be too noisy.

## Implementation

The MATTERMOST_CHANNEL env variable now supports a JSON objects (additionally to a simple string), to configure different channels based on registry channel.

```
export MATTERMOST_CHANNEL='{"beta":"publication,team", "dev": "team", "stable":"publication,team"}'
```

### DIFF
--- a/packages/cozy-app-publish/lib/hooks/post/mattermost.js
+++ b/packages/cozy-app-publish/lib/hooks/post/mattermost.js
@@ -94,20 +94,21 @@ const sendMattermostReleaseMessage = async options => {
   console.log('↳ ℹ️  Sending message to Mattermost')
 
   const hookURL = process.env.MATTERMOST_HOOK_URL
-  const channel = process.env.MATTERMOST_CHANNEL
+  const channels = process.env.MATTERMOST_CHANNEL.split(',')
   const iconURL = 'https://travis-ci.com/images/logos/TravisCI-Mascot-1.png'
   const username = 'Travis'
   const message = getMessage(options)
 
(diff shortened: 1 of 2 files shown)


packages/cozy-app-publish/lib/hooks/post/mattermost.js
```diff
@@ -93,20 +81,69 @@ const sendMattermostReleaseMessage = (
...
+    })
+  }
 }
 
+module.exports = sendMattermostReleaseMessage
 module.exports.getMessage = getMessage
+module.exports.sendMattermostMessage = sendMattermostMessage
 module.exports.sendMattermostReleaseMessage = sendMattermostReleaseMessage
+
+if (require.main === module) {
```
y-lohse: Could you explain what this does?

ptbrowne: > When a file is run directly from Node.js, require.main is set to its module. That means that it is possible to determine whether a file has been run directly by testing require.main === module.

https://nodejs.org/api/modules.html#modules_accessing_the_main_module

This enables this file to be imported but the code inside the if is not execute when the file is imported, only when the file is directly executed with noed.

PR #960, opened 2020-02-17
[Harvest] Feat: Fallback for app links

y-lohse:
In #933 I added links to contacts and banks from the harvest modal. This PR makes sure that when the apps are not installed, a link to the store is displayed instead.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/Success/DriveLink.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/Success/DriveLink.jsx
@@ -1,28 +1,43 @@
 import React, { memo } from 'react'
 import PropTypes from 'prop-types'
+import AppLinker from 'cozy-ui/transpiled/react/AppLinker'
 import { ButtonLink } from 'cozy-ui/transpiled/react/Button'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
 import { withClient } from 'cozy-client'
 import useAppLinkWithStoreFallback from '../../hooks/useAppLinkWithStoreFallback'
 
 const DriveLink = memo(({ folderId, client, t }) => {
-  const { fetchStatus, url, isInstalled } = useAppLinkWithStoreFallback(
-    'drive',
-    client,
-    `#/files/${folderId}`
-  )
-  const href = fetchStatus === 'loaded' && isInstalled ? url : undefined
+  const slug = 'drive'
+  const path = `#/files/${folderId}`
+  const { fetchStatus, url } = useAppLinkWithStoreFallback(slug, client, path)
 
-  return (
-    <ButtonLink
-      href={href}
-      target="_parent"
-      disabled={!isInstalled}
-      subtle
-      icon="openwith"
-      label={t('account.success.driveLinkText')}
-    />
-  )
+  if (fetchStatus === 'loaded') {
+    return (
+      <AppLinker slug={slug} href={url} nativePath={path}>
+        {({ href, name }) => (
+          <ButtonLink
+            icon="openwith"
+            href={href}
+            label={t('account.success.driveLinkText', {
+              appName: name
+            })}
+            subtle
+          />
+        )}
+      </AppLinker>
+    )
+  } else {
+    return (
+      <ButtonLink
+        icon="openwith"
+        label={t('account.success.banksLinkText', {
+          appName: name
+        })}
+        busy
+        subtle
+      />
+    )
+  }
 })
 
 DriveLink.displayName = 'DriveLink'
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/Success/DriveLink.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/Success/DriveLink.jsx
@@ -1,35 +1,33 @@
-import React, { PureComponent } from 'react'
+import React, { memo } from 'react'
 import PropTypes from 'prop-types'
 import { ButtonLink } from 'cozy-ui/transpiled/react/Button'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
 import { queryConnect, Q } from 'cozy-client'
 
-export class DriveLink extends PureComponent {
-  render() {
-    const { folderId, driveQuery, t } = this.props
-    let hasDrive = false
-    if (driveQuery.fetchStatus === 'loaded') {
-      if (driveQuery.data.length > 0) {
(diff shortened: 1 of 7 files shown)


packages/cozy-harvest-lib/src/components/KonnectorConfiguration/Success/DriveLink.jsx
```diff
@@ -1,43 +1,34 @@
...
+    <ButtonLink
+      href={href}
+      target="_parent"
+      disabled={!isInstalled}
+      subtle
+      icon="openwith"
+      label={t('account.success.driveLinkText')}
+    />
+  )
+})
```
Crash--: Why no AppLinker here?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/Success/DriveLink.jsx
```diff
@@ -1,43 +1,34 @@
...
-        icon="openwith"
-        label={t('account.success.driveLinkText')}
-      />
-    )
-  }
-}
+  return (
+    <ButtonLink
+      href={href}
+      target="_parent"
```
Crash--: BanksLinks doesn't have this attribute. Is it wanted?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/Success/DriveLink.jsx
```diff
@@ -1,43 +1,34 @@
...
+      disabled={!isInstalled}
```
Crash--: Why not a busy case like what you have done on the BanksLinks?

packages/cozy-harvest-lib/src/components/cards/AppLinkCard.jsx
```diff
@@ -23,7 +24,13 @@ const AppLinkCard = ({
...
 }) => {
   const cozyURL = new URL(client.getStackClient().uri)
-  const { cozySubdomainType: subDomainType } = client.getInstanceOptions()
+  const { fetchStatus, url, isInstalled } = useAppLinkWithStoreFallback(
+    slug,
+    client,
+    path
+  )
+
+  if (fetchStatus !== 'loaded') return null
```
Crash--: Oh again a new case. 

Should we display a busy button when we wait the query's return?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/Success/DriveLink.jsx
y-lohse: I'll ask but it should probably be added.

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/Success/DriveLink.jsx
y-lohse: Same, let me check but my guess is it's not intentional.

packages/cozy-harvest-lib/src/components/hooks/useAppLinkWithStoreFallback.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const useAppLinkWithStoreFallback = (slug, client, path = '') => {
+  const [fetchStatus, setFetchStatus] = useState('loading')
+  const [isInstalled, setIsInstalled] = useState(true)
+  const [url, setURL] = useState()
+
+  useEffect(() => {
+    const load = async () => {
+      try {
+        const apps = await client.query(Q('io.cozy.apps'))
```
ptbrowne: Why do we load all the apps when we need only one ?
Could we use a fetch policy here to limit the number of calls and limit the spinning hell ?

packages/cozy-harvest-lib/src/components/cards/AppLinkCard.jsx
y-lohse: Yeah this one is intentional. The other buttons look similar whether or not the app is already installed, but here the label and icons change. I figured this would make the button change too much once the data is loaded, so I thought it would be better to only display it once we know what it should look like.

packages/cozy-harvest-lib/src/components/hooks/useAppLinkWithStoreFallback.jsx
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from 'react'
+import { Q, models } from 'cozy-client'
+
+const { applications: ApplicationsModel } = models
```
ptbrowne: Why do we rename this import to ApplicationsModel ?
It seems like ApplicationsModel is a constructor when it is only a namespace for helper functions.

Not in the scope of this PR: I have the feeling that applications/jobs/triggers/files namespaces for helpers could directly be exported from `cozy-client` and not from a models object. It would enable more precise tree shaking (here we do not need all the models, only the applications).

packages/cozy-harvest-lib/src/components/hooks/useAppLinkWithStoreFallback.jsx
```diff
@@ -0,0 +1,41 @@
...
+        } else {
+          setURL(ApplicationsModel.getStoreURL(apps.data, appDocument))
+        }
+        setFetchStatus('loaded')
+      } catch (error) {
+        setFetchStatus('errored')
+      }
+    }
+    load()
+  }, [client, slug, path])
```
ptbrowne: This is not the first time that we have a `fetching` effect with an internal loading state, I remember we had one for the maintenance route, I wonder if we should have a  useFetcher hook that did that, this way we would not have to test this loading/loaded/errored logic everytime.

ptbrowne: This is very similar to the state that query maintain in the cozy-client state, I wonder if we could piggyback on that, I feel like we are reinventing the wheel here.

packages/cozy-harvest-lib/src/components/hooks/useAppLinkWithStoreFallback.jsx
Crash--: Can we do that? I mean having a fetchPolicy on a query()? I thought we were waiting a useQuery() hook or something to do so

y-lohse: We can't at the moment, see https://github.com/cozy/cozy-client/issues/646

PR #961, opened 2020-02-17
fix: Load current konnector error

y-lohse:
Something that slipped through the big PR — `FlowProvider` no longer directly injects `error` and similar props, instead they need to be retrieved from the `flow` object.


ptbrowne:
Thanks 😊

Crash--:
No updated tests? :-°

PR #963, opened 2020-02-18
fix: Account correctly created with the account_type property

ptbrowne:
Since the TriggerManger would pass an empty object as account, the
ConnectionFlow executed the update path, that would not put the
account_type property inside the io.cozy.accounts

Without an account_type, the stack would not trigger the on_deletion
mechanism which is particularly important for banking connectors.

### DIFF
--- a/packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
+++ b/packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
@@ -120,7 +120,9 @@ describe('TriggerManager', () => {
 
         await expect(findByLabelText('username')).resolves.toBeDefined()
         await expect(findByLabelText('passphrase')).resolves.toBeDefined()
-        await expect(findByTitle('back', null, { timeout: 500 })).rejects.toThrow()
+        await expect(
+          findByTitle('back', null, { timeout: 500 })
+        ).rejects.toThrow()
       })
     })
 
@@ -172,7 +174,9 @@ describe('TriggerManager', () => {
           const passwordField = await findByLabelText('passphrase')
           const backButton = await findByTitle('back')
 
-          await expect(findByLabelText('username', null, { timeout: 500 })).rejects.toThrow()
+          await expect(
+            findByLabelText('username', null, { timeout: 500 })
+          ).rejects.toThrow()
           expect(passwordField).toBeDefined()
           expect(backButton).toBeDefined()
 
@@ -305,8 +309,7 @@ describe('TriggerManager', () => {
           flow,
           identifierInput,
           passphraseInput,
-          submitButton,
-          debug
+          submitButton
         } = await setupForm({ account })
         fireEvent.change(identifierInput, {
           target: { value: 'my-identifier' }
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
```diff
@@ -231,4 +231,95 @@ describe('TriggerManager', () => {
...
+      const flow = new ConnectionFlow(client)
+      jest.spyOn(flow, 'handleFormSubmit').mockImplementation(() => {})
+      const utils = render(
+        <TriggerManager {...props} account={account} flow={flow} />
+      )
+      const identifierLabel = await utils.findByLabelText('username')
+      const passphraseLabel = await utils.findByLabelText('passphrase')
+      const identifierInput = identifierLabel.nextElementSibling
+      const passphraseInput =
+        passphraseLabel.nextElementSibling.nextElementSibling
```
Crash--: nextElementSibling.nextElementSibling seems fragile. Can't we use a better selector? 

Is there a reason to use `findBy*` instead of `getBy`? `getBy` is not async and return the first match (getBy* queries return the first matching node for a query, and throw an error if no elements match or if more than one match is found (use getAllBy instead).)

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
```diff
@@ -231,4 +231,95 @@ describe('TriggerManager', () => {
...
+        identifierInput,
+        passphraseInput,
+        submitButton,
+
+        ...utils
+      }
+    }
+
+    describe('when no account is passed in props', () => {
+      fit('should correctly send the form', async () => {
```
Crash--:
```suggestion
      it('should correctly send the form', async () => {
```

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
```diff
@@ -231,4 +231,95 @@ describe('TriggerManager', () => {
...
+              passphrase: 'my-passphrase',
+              username: 'my-identifier'
+            }
+          })
+        )
+      })
+    })
+
+    describe('when an account is passed in props', () => {
+      fit('should correctly send the form', async () => {
```
Crash--:
```suggestion
      it('should correctly send the form', async () => {
```

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
```diff
@@ -231,4 +231,95 @@ describe('TriggerManager', () => {
...
+        } = await setupForm({ account })
+        fireEvent.change(identifierInput, {
+          target: { value: 'my-identifier' }
+        })
+        fireEvent.change(passphraseInput, {
+          target: { value: 'my-passphrase' }
+        })
+        fireEvent.click(submitButton)
+        expect(flow.handleFormSubmit).toHaveBeenCalledWith(
+          expect.objectContaining({
```
Crash--: Thanks! Last time I was looking for something link that! 😍

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
ptbrowne: > nextElementSibling.nextElementSibling seems fragile. Can't we use a better selector?

I know, I tried to use a data-test-id but the AccountFields would not forward the prop until the dom node.

> Is there a reason to use findBy* instead of getBy

I tried with a getBy first, but it did not work, I think the AccountForm render is a bit async.

PR #964, opened 2020-02-18
test: Check FlowProvider state prop

y-lohse:
This is a follow up on #961… but I found it hard to add a lot of new tests. Actually `FlowProvider` and `ConnectionFlow` have both a decent test coverage already. I ended up reworking the current tests so that they are more sensible and added one that explicitly tests the error case fixed in the previous PR.

### DIFF
--- a/packages/cozy-harvest-lib/test/components/FlowProvider.spec.js
+++ b/packages/cozy-harvest-lib/test/components/FlowProvider.spec.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React from 'react'
 import { shallow } from 'enzyme'
 import { FlowProvider } from 'components/FlowProvider'
 import {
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/test/components/FlowProvider.spec.js
```diff
@@ -57,59 +55,65 @@ describe('FlowProvider', () => {
...
+    flow.triggerEvent(TWO_FA_REQUEST_EVENT)
+    wrapper.update()
+    expect(wrapper.find(TwoFAModal).length).toBe(1)
+
+    flow.triggerEvent(LOGIN_SUCCESS_EVENT)
+    wrapper.update()
+    expect(wrapper.find(TwoFAModal).length).toBe(0)
+    expect(children.mock.calls[2][0].flow.getState().status).toEqual(
+      'LOGIN_SUCCESS'
+    )
```
ptbrowne: Since the flow instance does not change between calls, wouldn't it be simpler to simply do:

```
    expect(flow.getState().status).toEqual(
      'LOGIN_SUCCESS'
    )
```

y-lohse: It would be simpler, but the current approach really tests what the child component of `FlowProvider` receives, without relying on the internal implementation of it :)

PR #966, opened 2020-02-18
fix: Support Budget Insight alternate schema for update connection

ptbrowne:
Budget Insight sometimes "errors" when it cannot force update the connection
to the banking vendor (for example if the connection has been refreshed
not too long ago). In this case, the response schema is different than the
normal Connection ressource that we expected.

### DIFF
--- a/packages/cozy-harvest-lib/src/services/bi-http.js
+++ b/packages/cozy-harvest-lib/src/services/bi-http.js
@@ -97,13 +97,24 @@ export const updateBIConnection = async (
     connId,
     `Must pass connection id to updateBIConnection (${connId} was passed)`
   )
-  return await biRequest(
-    'PUT',
+  const resp = await biRequest(
+    'POST',
     `/users/me/connections/${connId}`,
     config,
     encodeToForm(data),
     biAccessToken
   )
+
+  // The doc indicates that the response should be a connection object.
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/services/bi-http.js
```diff
@@ -97,13 +97,24 @@ export const updateBIConnection = async (
...
+    'POST',
     `/users/me/connections/${connId}`,
     config,
     encodeToForm(data),
     biAccessToken
   )
+
+  // The doc indicates that the response should be a connection object.
+  // But, when the connection cannot be force updated, the response is
+  // in the form { connection, message }.
```
Crash--: Is there any link to BI's doc we can add? 

Is it something you already handle in the connector?

packages/cozy-harvest-lib/src/services/bi-http.js
```diff
@@ -97,13 +97,24 @@ export const updateBIConnection = async (
     connId,
     `Must pass connection id to updateBIConnection (${connId} was passed)`
   )
-  return await biRequest(
-    'PUT',
+  const resp = await biRequest(
+    'POST',
```
Crash--: Even if we already have a connId, we have to POST instead of PUT?

Issue #475, opened 2019-05-31
[realtime] Socket.js may open multiple sockets in parallel


edas:
In socket.js : 
```
async subscribe(type, id) {
    if (!this.isOpen()) {
      await this.connect()
    }
```

When the socket is "CONNECTING", this code will open a new websocket in `this._webSocket`. The old one is not closed, which means:
- The old one may emit an "OPEN" event while the new one in `this._webSocket` is still "CONNECTING"
- We may receive duplicate events later because both websockets will emit message events
- Both sockets may be kept opened

I see we have some defensive code against that like 
```
authenticate() {
    if (this.isOpen()) {
```
which forbids authentication if an old socket is receiving 'onopen' but the new one is still CONNECTING. The underlying problem is however still there

edas:
Closing, as it is related to an old version of realtime, before the rewrite of #967

Issue #474, opened 2019-05-31
[realtime] Use an internal private data from MicroEE


edas:
To resubscribe to events when recreating a new socket, Realtime uses `this._events`, which is an internal variable from MicroEE, not part of the public interface.

edas:
Closing, as it is related to an old version of realtime, before the rewrite of #967

PR #967, opened 2020-02-21
feat: Rewrite of realtime

### LINKED ISSUE #449: [realtime] Initialisation with wrong option
CozyRealtime should throw early if wrong client is passed.

Original call (mistake on the end of the developer)

```
new CozyRealtime(client)
```

Error

```
TypeError: this._cozyClient is undefined index.js:196
    CozyRealtime index.js:196
    initializeRealtime index.js:233
    initApp index.js:185
    createThunkMiddleware Redux
    initApp App.jsx:75
    App App.jsx:31
    React 23
    renderApp index.jsx:14
    LiWt index.jsx:41
```

What I would like to see

`ValueError: You need to pass { cozyClient } when initializing realtime: new CozyRealtime({ cozyClient })`

edas:
BREAKING CHANGE: There should be no breaking changes, but as it is a full rewrite, let's act as if it does.

Among the changes :

- Realtime now correctly reconnect if the computer goes to sleep, whatever the time gone offline.

- We do not try on attempt every 200ms 60 times anymore. There is an exponential backoff mechanism. The reconnection may take up to 5 minutes (worst case scenario).

- It now allows to specify an id when subscribing to the CREATED event. This was needed for cozy-notes.

- It allows to subscribe to all events for a doctype (and optionally an id). The socket with the stack was subscribing to all events anyways.

- The code is longer but should be way simplier. I deleted the extra layer with  Socket.js which was mostly boilerplate and additional complexity for asynchroneous events.

- The subscriptions are now handled by a specific class instead of doing shameful magic with the internal of MicroEE.

- The retry logic has an external class, easying to tweak the values in the future.

- There are Minilog logs on level info or debug that can be activated when needed (same inside the main test file, because sometimes it may help). Only warnings are activated by default.

-  Fix #449

It should fix or make obsolete most  Realtime issues. At least #474,  #475, #466,


edas:
The failing test seems to be in cozy-harvest-lib

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,534 @@
...
+  async sendNotification(doctype, id, data) {
+    return this.client
+      .getStackClient()
+      .fetchJSON('POST', `/realtime/${doctype}/${id}`, {
+        data
+      })
+  }
+  send(...args) {
+    return this.sendNotification(...args)
+  }
```
Crash--: Why this abstraction? Can't we have a direct call to sendNotification? Or renaming sendNotification to send?

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,534 @@
...
+   * @param {string} type
+   * @param {string|undefined|function} id - (or handler is `handler` is undefined)
+   * @param {function|undefined} handler - (not used if `id` is a function)
+   */
+  subscribe(event, type, id, handler) {
+    this.start() //  start the realtime in case it was stopped
+    const sub = this.normalizeSubscription(event, type, id, handler)
+    const has = this.subscriptions.hasSameTypeAndId(sub)
+    this.subscriptions.add(sub)
+    // send to the server if there wasn't any with that type & id before
```
Crash--:
```suggestion
    // send to the server if there wasn't any subscription with that type & id before
```

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,534 @@
...
+    if (!has && this.isWebSocketAuthenticated()) {
```
Crash--: What if the websocket is not authenticated? Should we store the `subscriptions` we want to make and send them when the socket is authenticated?

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,534 @@
...
-   * Re subscribe on server
+   * Normalize a subscription request object
+   *
+   * This allows to support the two different signatures possible
+   * for the subscribe() and the unsubscribe() methods
+   *
+   * @private
+   * @param {EventName} event
+   * @param {string} type
+   * @param {string|undefined|function} id - (or handler is `handler` is undefined)
```
Crash--: ?

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,534 @@
...
+   * @private
    */
-  unsubscribeAll() {
-    this._getEventKeys()
-      .map(key => this._events[key].length > 0 && key)
-      .filter(Boolean)
-      .forEach(key => (this._events[key] = []))
+  onClientTokenRefreshed() {
+    // no need to authenticate again
+  }
```
Crash--: I understand the theory behind that case. But what if we start an Oauth flow, with the token that will expire in a few seconds, and then connect to a realtime? We will send an expired token, the client will refresh the token, and then we will have to wait the retry mechanism?

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,534 @@
...
+  /* * * * * * * * * * * * * * * */
+  /* * Global (browser) EVENTS * */
+  /* * * * * * * * * * * * * * * */
+
+  /**
+   * Subscribe to global events
+   * @private
+   */
+  subscribeGlobalEvents() {
+    if (isInBrowser && window.addEventListener) {
```
Crash--: I'm not confortable with the fact that we only handle the browser use case. Realtime should manage our mobile app correctly

Crash--: Wo, ok. isBrowser only test for window !== undefined. I don't like the naming then.

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,534 @@
...
+      window.addEventListener('online', this.onOnline)
```
Crash--: Since in a cordova context, we need to install the cordova-plugin-network-information plugin to have access to these events, we should check if this plugin is installed and warn the dev if it's not. 

Also, we should add somewhere that we should not call realtime.start() / subscribe() before the `deviceReady` call from Cordova

packages/cozy-realtime/src/RetryManager.js
```diff
@@ -0,0 +1,142 @@
...
+      await new Promise(resolve => {
+        global.setTimeout(resolve, this.wait)
+      })
+    }
+  }
+}
+
+MicroEE.mixin(RetryManager)
+
+export default RetryManager
```
Crash--: If I'm not mistaken we already have a similar mechanism in cozy-note. Should we extract this kind of logic and make them accessible from cozy-libs or something?

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,534 @@
...
+      this.websocket.onmessage = null
+      this.websocket.onerror = null
+      this.websocket.onopen = null
+      this.websocket.onclose = null
+      try {
+        this.websocket.close()
+      } catch (e) {
+        // void
+      }
+      this.websocket = null
```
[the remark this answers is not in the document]
edas: If close returns an error,it will be catched line 109. As the catch blockdoes not raise itself, the process will continue  and line 112 will be executed, setting the attribute  to null. Am I  wrong ?

That said, a finally block is indeed clearer, I will do the change.

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,534 @@
...
+   * Close the Realtime
+   * @private
+   */
+  stop() {
+    if (this.isStarted) {
+      logger.info('stopped')
+      this.unsubscribeGlobalEvents()
+      this.unsubscribeClientEvents()
+      if (this.hasWebSocket()) {
+        this.rejectWebSocket()
```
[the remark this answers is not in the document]
edas: :+1: for changing the name

packages/cozy-realtime/src/CozyRealtime.js
edas: Because the previous class had a direct `send` method. I found it misleading (I would expect something to send a message through the websocket, but it does something else) so I made a new name but kept the old one for compatibility.

packages/cozy-realtime/src/CozyRealtime.js
edas: Ok, I add a link to  normalizeSubscription() which  deals with that.  These methods had a special signature where you can pass (event, type, id, handler) or (event,  type,handler).I kept these possibilities for compatibility

packages/cozy-realtime/src/CozyRealtime.js
ptbrowne: Could we have a deprecation warning here ?

packages/cozy-realtime/src/RetryManager.js
```diff
@@ -0,0 +1,142 @@
+import MicroEE from 'microee'
+
+import {
+  maxWaitBetweenRetries,
+  baseWaitAfterFirstFailure,
+  timeBeforeSuccessfull,
+  raiseErrorAfterAttempts
+} from './config'
+
+class RetryManager {
```
ptbrowne: Can we have a docstring here ?

packages/cozy-realtime/src/RetryManager.js
```diff
@@ -0,0 +1,142 @@
...
+    this.reset()
+    this.onSuccess = this.onSuccess.bind(this)
+    this.onFailure = this.onFailure.bind(this)
+    this.onSuccessAcknowledged = this.onSuccessAcknowledged.bind(this)
+  }
+
+  /**
+   * On a success
+   *
+   * If `timeBeforeSuccessfull` if not null, it will wait this amount of
```
ptbrowne:
```suggestion
   * If `timeBeforeSuccessful` if not null, it will wait this amount of
```

packages/cozy-realtime/src/CozyRealtime.js
edas: We always store the subscriptions, it is the role of `this.subscriptions.add(sub)` two lines above. The subscriptions are always all sent after authentication and future opening are already covered. This is  a special cas for the current already opened websocket.

packages/cozy-realtime/src/CozyRealtime.js
edas: We won't even have a retry mechanism in this case (as we won't be able torecover automatically anyways for a bad token).

And yes, you're right, we may have a very bad timing. I will add a reconnect there.

packages/cozy-realtime/src/CozyRealtime.js
Crash--: In that case, where are we removing the subscription if we send the Subscription successfully?

packages/cozy-realtime/src/RetryManager.js
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  /**
+   * Returns if we reached a level where we need to raise an error
+   *
+   * @returns {boolean}
+   */
+  hasReachedErrorLevel() {
+    return this.retries > 0 && this.retries % raiseErrorAfterAttempts == 0
+  }
```
ptbrowne: Architecture: I wonder her why `raiseErrorAfterAttempts` is not a parameter of the class instead of a global config variable.

Naming: hasReachedErrorLevel seems to imply that the "error level" is a threshold whereas modular arithmetic is applied, meaning that the level can be reached multiple times. Maybe `shouldEmitError` could be used instead ?

packages/cozy-realtime/src/RetryManager.js
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Reset failures counters (do not wait for before trying to connect next time)
```
ptbrowne: I think a word is missing here : do not wait for what ?

packages/cozy-realtime/src/CozyRealtime.js
edas: would you be more confortable with `hasBrowserContext` ?

packages/cozy-realtime/src/RetryManager.js
```diff
@@ -0,0 +1,142 @@
...
+   */
+  reset() {
+    this.clearSuccessWatcher()
+    this.retries = 0
+    this.wait = false
+  }
+
+  /**
+   * Increase the failure counter (wait more before next attempt)
```
ptbrowne:
```suggestion
   * Increases the failure counter (wait more before next attempt)
```

packages/cozy-realtime/src/RetryManager.js
```diff
@@ -0,0 +1,142 @@
...
+   *
+   * @private
+   */
+  increaseFailureCounter() {
+    if (this.wait === false) {
+      // first reconnect is immediate
+      this.wait = 0
```
ptbrowne: What is the difference between `this.wait = 0` and `thiss.wait = false`

packages/cozy-realtime/src/CozyRealtime.js
edas: These checks were not in the previous version. To you know how I can check that we are in a cordova layer, that we have the correct plugin installed, and if deviceReady has fired or not?

packages/cozy-realtime/src/RetryManager.js
```diff
@@ -0,0 +1,142 @@
...
+   * @param {function} ack - will be called after the wait time
+   * @param {number} time - waiting time
+   */
+  startSuccessWatcher(ack, time) {
+    if (!this.successWatcher) {
+      const callback = () => {
+        this.successWatcher = null
+        ack()
+      }
+      this.successWatcher = global.setTimeout(callback, time)
```
ptbrowne: Why not calling setTimeout directly (without global. ) ?

packages/cozy-realtime/src/RetryManager.js
```diff
@@ -0,0 +1,142 @@
...
+   */
+  clearSuccessWatcher() {
+    if (this.successWatcher) {
+      global.clearTimeout(this.successWatcher)
+      this.successWatcher = null
+    }
+  }
+
+  /**
+   * Starts the success watcher - acknowledges the success if not interrupted after some time
```
ptbrowne: I have trouble understanding why we need this ? Could the docstring be expanded with a rationale behind this ?

packages/cozy-realtime/src/RetryManager.spec.js
```diff
@@ -0,0 +1,152 @@
+import times from 'lodash/times'
+
+import RetryManager from './RetryManager'
+import { raiseErrorAfterAttempts, timeBeforeSuccessfull } from './config'
+
+const pause = time => new Promise(resolve => setTimeout(resolve, time))
```
ptbrowne: We usually use `sleep` name for this function.

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,534 @@
...
+      const token = getToken(this.client)
+      logger.debug('authenticate with', token)
+      this.sendWebSocketMessage('AUTH', token)
+      this.websocket.authenticated = true
+    } else {
+      logger.warn(
+        'Trying to authenticate to a non-opened websocket. This should not happen.',
+        this.websocket,
+        this.websocket && this.websocket.readyState
+      )
```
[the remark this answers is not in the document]
edas: Well, warning should not happen either. The semantic in use is "If we can recover automatically or if we end up in a stable situation which is the one the developper wanted anyways, it's a warning, otherwise it's an error"

packages/cozy-realtime/src/CozyRealtime.js
Crash--: Oh I forgot to put the link... we do this kind of check here https://github.com/cozy/cozy-libs/blob/master/packages/cozy-device-helper/src/plugins.js#L7 

We can not really check if `deviceReady` has been already called or not. Adding a note to the readme should be enough IMHO.

packages/cozy-realtime/src/CozyRealtime.js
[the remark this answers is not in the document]
Crash--: So it means that in our future work to connect minilog to sentry, we will listen to not only error but also warn? just to keep in mind

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,549 @@
...
 
-  let id, handler
-  if (!handlerOrUndefined) {
-    id = undefined
-    handler = handlerOrId
-  } else {
-    id = handlerOrId
-    handler = handlerOrUndefined
+  /**
+   * Start all handlers and try to connect if online
```
ptbrowne: What happens if we are not online ? Will the realtime try to connect automtically ? I think this should be expressed in the class docstring.

ptbrowne: Nit: we usually use 3rd person in docstrings : s/Start/Starts

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,549 @@
...
   }
 
-  _getEventKeys() {
-    if (!this._events) return []
+  /**
+   * Unsubscribe to global events
+   * @private
+   */
+  unsubscribeGlobalEvents() {
+    if (isInBrowser && window.removeEventListener) {
```
ptbrowne: I would prefer to repeat `typeof window !== 'undefined'` instead of using isInBrowser here. In the case where the `isInBrowser` logic changes and uses another object, other than window, the code would be more robust (here it is almost a coincidence that isInBrowser uses `window`)

packages/cozy-realtime/src/CozyRealtime.js
edas: Seen together.

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,549 @@
...
    */
-  unsubscribeAll() {
-    this._getEventKeys()
-      .map(key => this._events[key].length > 0 && key)
-      .filter(Boolean)
-      .forEach(key => (this._events[key] = []))
+  onClientTokenRefreshed() {
+    // In theory, we don't have to do anything, and there is no need to authenticate again.
+    // We are reconnecting anyways because if we have a really bad timing, we may have sent an
+    // expired token for authentication, and we need to reconnect if that happen
```
ptbrowne: I don't understand the "really bad timing" part here. I think the "In theory we don't have to do anything" can be removed if we know that we a "really bad timing" is possible.

packages/cozy-realtime/src/CozyRealtime.js
Crash--: Yep 👍

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,549 @@
...
+   *
+   * @private
+   * @returns {WebSocket} the created websocket
+   */
+  createSocket() {
+    if (this.websocket) {
+      logger.warn(
+        'Unattended replacement of an existing socket, this should not happen'
+      )
+      this.revokeWebSocket(this.websocket)
```
ptbrowne: I think we can remove `this.websocket` from the arguments of this.removeWebSocket.

packages/cozy-realtime/src/CozyRealtime.js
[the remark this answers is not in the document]
edas: That's a good question. I didn't have that in mind. @ptbrowne do you have advices on this?

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,549 @@
...
+  /**
+   * Create and attach a WebSocket to the current instance
+   *
+   * @private
+   * @returns {WebSocket} the created websocket
+   */
+  createSocket() {
+    if (this.websocket) {
+      logger.warn(
+        'Unattended replacement of an existing socket, this should not happen'
```
```suggestion
        'Unexpected replacement of an existing socket, this should not happen'
```

ptbrowne: I think the warning could be more useful if it is included a possible solution : "this should not happen. createSocket should not be called twice in a row. You can call stop() to remove the websocket."

ptbrowne: If this should not happen, should we log an error ?

packages/cozy-realtime/src/CozyRealtime.js
edas: done

packages/cozy-realtime/src/RetryManager.js
edas: done

packages/cozy-realtime/src/RetryManager.js
edas: :+1: done

packages/cozy-realtime/src/RetryManager.js
edas: Yep, that's why I made anexternal class instead of an embeddedlogic as in notes. I still have  to check what's needed to use the same class for both. I prefer to make it fully external in a second step

packages/cozy-realtime/src/RetryManager.js
edas: fixed

packages/cozy-realtime/src/RetryManager.js
edas: This is an ugly hack to have the first reconnect to be immediate. 
false (no wait) -> 0 (wait 0ms) -> 128 (wait 128ms) -> …

I am adding comments for  this

packages/cozy-realtime/src/RetryManager.js
edas: I used to always be explicit when getting a function or attribute from the global context.

packages/cozy-realtime/src/RetryManager.js
edas: :+1:

packages/cozy-realtime/src/RetryManager.spec.js
edas: :+1:

packages/cozy-realtime/src/CozyRealtime.js
edas: :+1: for the third person, fixed.

When not online, it will connect once it receive thee online event from the browser. Let me add that in the comment

packages/cozy-realtime/src/CozyRealtime.js
edas: :+1:

packages/cozy-realtime/src/CozyRealtime.js
edas: :+1:

packages/cozy-realtime/src/CozyRealtime.js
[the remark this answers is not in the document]
ptbrowne: I looked on the internet for a mental model of what should be an error and what should be a warning:

> A warning tends to be a notice (either from the compiler or a library) that does not imply a fatal error, but instead something that could go wrong (eg memory leakage, unused variables) or something that might not work in the future (eg deprecation). An error means that something went wrong, either recoverably or unrecoverably. A recoverable error is one after which the program can (probably) still continue whereas an unrecoverable error is one which causes the program to 'hang' or 'crash'.

https://www.quora.com/What-is-the-difference-between-warning-and-errors

I like this mental model. With this mental model, I think here we have an error, even if this is recoverable. As a developer, I look at the error and think that I should fix it, which seems to be the case here.

> So it means that in our future work to connect minilog to sentry, we will listen to not only error but also warn?

I think warnings (notices that something could break) are informative and can help when debugging, so I am OK to send them to Sentry.

packages/cozy-realtime/src/CozyRealtime.js
edas: We ended up with `hasBrowserContext`

packages/cozy-realtime/src/CozyRealtime.js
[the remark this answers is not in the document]
edas: Thanks for the research. Let's put errors is these cases then.

packages/cozy-realtime/src/CozyRealtime.js
edas: changed

packages/cozy-realtime/src/CozyRealtime.js
edas: waiting for https://github.com/cozy/cozy-libs/pull/969

edas: For deviceready, I added an event handler for that in order to re-attach online/offline event handler if ever  the deviceready fires after cozy-realtime init.

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -1,360 +1,600 @@
...
+  /**
+   * Subscribes to global events
+   * @private
+   */
+  subscribeGlobalEvents() {
+    if (hasBrowserContext && window.addEventListener) {
+      window.addEventListener('online', this.onOnline)
+      window.addEventListener('offline', this.onOffline)
+    }
+    if (isCordova) {
```
Crash--:
```suggestion
    if (isCordova()) {
```

packages/cozy-realtime/src/CozyRealtime.js
Crash--: > For deviceready, I added an event handler for that in order to re-attach online/offline event handler if ever the deviceready fires after cozy-realtime init.

Thanks!

PR #968, opened 2020-02-24
fix: Load flags from legacy DOM

y-lohse:
There was a small typo causing a crash when we tried to load flags from the `data-cozy-flags` DOM node.

### DIFF
--- a/packages/cozy-flags/src/tests.js
+++ b/packages/cozy-flags/src/tests.js
@@ -117,7 +117,7 @@ export default function testFlagAPI(flag) {
         }
       })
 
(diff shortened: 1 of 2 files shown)


packages/cozy-flags/src/tests.js
```diff
@@ -116,6 +116,26 @@ export default function testFlagAPI(flag) {
           document.body.removeChild(div)
         }
       })
+
+      fit('should initialize from DOM (legacy)', async () => {
```
ptbrowne:
```suggestion
      it('should initialize from DOM (legacy)', async () => {
```

PR #973, opened 2020-02-24
fix: Init flags when client is already logged in

y-lohse:
The flags plugin for cozy-client waits for the `login` event to load the flags. We often do something like:

```
const client = new CozyClient({ token, url })
client.registerPlugin(flag.plugin)
```

… but when the plugin is registered, the `login` event has already been fired and the flags are never loaded.

This PR solves the problem by loading the flags iff it detects that the client is already logged in. SInce loading the flags is `async`, there is now also an exposed promise on `client.plugins.flags.initializing` to know when they are ready.

### DIFF
--- a/packages/cozy-flags/src/tests.js
+++ b/packages/cozy-flags/src/tests.js
@@ -89,15 +89,29 @@ export default function testFlagAPI(flag) {
       return { client }
     }
 
-    it('should have a cozy-client plugin', async () => {
-      const { client } = setup()
-      client.registerPlugin(flag.plugin)
-      await client.plugins.flags.handleLogin()
-      expect(flag('has_feature1')).toBe(true)
-      expect(flag('has_feature2')).toBe(false)
-      expect(flag('from_remote')).toBe(true)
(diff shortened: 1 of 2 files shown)


packages/cozy-flags/src/tests.js
```diff
@@ -89,15 +89,29 @@ export default function testFlagAPI(flag) {
...
-    it('should have a cozy-client plugin', async () => {
-      const { client } = setup()
-      client.registerPlugin(flag.plugin)
-      await client.plugins.flags.handleLogin()
-      expect(flag('has_feature1')).toBe(true)
-      expect(flag('has_feature2')).toBe(false)
-      expect(flag('from_remote')).toBe(true)
-      expect(flag('number_of_foos')).toBe(10)
-      expect(flag('bar_config')).toEqual({ qux: 'quux' })
+    fdescribe('cozy-client plugin', () => {
```
ptbrowne:
```suggestion
    describe('cozy-client plugin', () => {
```

PR #976, opened 2020-02-26
Fix two FA flows

ptbrowne:
- Tell the right thing to the user in case of app two fa
- Update cozy-bi-auth to have the correct mapping for the
  SCARequired error from BI
- Pull locales from transifex and store them in the repository

### DIFF
--- a/packages/cozy-harvest-lib/src/locales/en.json
+++ b/packages/cozy-harvest-lib/src/locales/en.json
@@ -339,7 +339,7 @@
       "default": "Enter the code sent to you by e-mail or SMS",
       "email": "Enter the code sent to you by e-mail",
       "sms": "Enter the code sent to you by SMS",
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/src/locales/en.json
```diff
@@ -339,7 +339,7 @@
       "default": "Enter the code sent to you by e-mail or SMS",
       "email": "Enter the code sent to you by e-mail",
       "sms": "Enter the code sent to you by SMS",
-      "app": "Enter the one-time code sent to you"
+      "app": "Use your provider' app to continue authentication"
```
y-lohse:
```suggestion
      "app": "Use your provider's app to continue authentication"
```

ptbrowne: Thanks.

PR #977, opened 2020-02-26
fix(realtime): Connect concurrency

edas:
Previously, an event could call for a reconnect
while the previous one was waiting before
creating  a new socket. This would create
a two socket in a row.

This change adds a protection against concurrency
when creating a socket.

### DIFF
--- a/packages/cozy-realtime/src/CozyRealtime.js
+++ b/packages/cozy-realtime/src/CozyRealtime.js
@@ -100,8 +100,19 @@ class CozyRealtime {
    */
   async connect() {
     logger.info('connecting…')
-    await this.retryManager.waitBeforeNextAttempt()
-    this.createSocket()
+    // avoid multiple concurrent calls to connect, keeps the first one
+    if (this.waitingForConnect) {
+      logger.debug('Pending reconnection, skipping reconnect')
+      return
+    }
+    logger.debug('No pending reconnection, will reconnect')
+    try {
+      this.waitingForConnect = true
+      await this.retryManager.waitBeforeNextAttempt()
+      this.createSocket()
+    } finally {
+      this.waitingForConnect = false
+    }
   }
 
   /**
@@ -618,6 +629,7 @@ class CozyRealtime {
    * @private
    */
   onOnline() {
+    logger.info('reconnect because receiving an online event')
     this.reconnect()
   }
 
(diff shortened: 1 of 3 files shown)


packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -100,8 +100,19 @@ class CozyRealtime {
    */
   async connect() {
     logger.info('connecting…')
-    await this.retryManager.waitBeforeNextAttempt()
-    this.createSocket()
+    // avoid multiple concurrent calls to connect, keeps the first one
+    if (!this.waitingForConnect) {
```
ptbrowne: Nit: I find it clearer when conditions are checked positively, and early bail out is used.

```
if (this.waitingForConnect) {
  logger.debug('Skipping reconnect because one is already waiting')
  return
}


... normal flow of function ...

File: packages/cozy-realtime/src/CozyRealtime.js [hunk 2]
```diff
@@ -100,8 +100,19 @@ class CozyRealtime {
...
+      logger.debug('no current reconnect is already waiting')
```
```suggestion
      logger.debug('No pending reconnection, will reconnect')
```

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -100,8 +100,19 @@ class CozyRealtime {
...
+      try {
+        this.waitingForConnect = true
+        await this.retryManager.waitBeforeNextAttempt()
+        this.createSocket()
+      } finally {
+        this.waitingForConnect = false
+      }
+    } else {
+      logger.debug('skipping reconnect because one is already waiting')
```
```suggestion
      logger.debug('Pending reconnection, skipping reconnect')
```

packages/cozy-realtime/src/CozyRealtime.spec.js
```diff
@@ -608,5 +608,31 @@ describe('CozyRealtime', () => {
...
+        await realtime.waitForSocketReady()
+
+        // makes the reconnect wait
+        realtime.retryManager.onFailure()
+        realtime.retryManager.onFailure()
+        realtime.retryManager.onFailure()
+        realtime.retryManager.onFailure()
+        realtime.retryManager.onFailure()
+        realtime.onOnline()
+        await sleep(25)
```
ptbrowne: What is this magic value ?

edas: It's a "wait a little for async events to take place"

PR #978, opened 2020-02-26
Realtime various fixes

edas:
These are very quick and little patches. Some were seen Tuesday on the frontend meeting. Others are fixes and cleaning

### DIFF
--- a/packages/cozy-realtime/src/CozyRealtime.js
+++ b/packages/cozy-realtime/src/CozyRealtime.js
@@ -448,7 +448,13 @@ class CozyRealtime {
     logger.info('receive message from server', { event, payload })
     const handlers = this.subscriptions.getAllHandlersForEvent(event, payload)
     for (const handler of handlers) {
-      handler(payload.doc)
+      try {
+        handler(payload.doc)
+      } catch (e) {
+        logger.error(
+          `handler did throw for ${event}, ${payload.type}, ${payload.id}`
+        )
+      }
     }
     if (event === 'error') {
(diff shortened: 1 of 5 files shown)


packages/cozy-realtime/src/SubscriptionList.js
```diff
@@ -59,10 +59,14 @@ export default class SubscriptionList {
...
-      if (found >= 0) pullAt(this.subscriptions, found)
+      if (found >= 0) {
+        pullAt(this.subscriptions, found)
+        return
+      }
     } else {
       // remove all matching
-      remove(this.subscriptions, s => isEqual(s, subscription))
+      const removed = remove(this.subscriptions, s => isEqual(s, subscription))
+      if (removed) return
```
y-lohse: Haven't run the code, but I  presume `if (removed)` is always true since `remove()` returns an array?

packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -448,7 +448,11 @@ class CozyRealtime {
     logger.info('receive message from server', { event, payload })
     const handlers = this.subscriptions.getAllHandlersForEvent(event, payload)
     for (const handler of handlers) {
-      handler(payload.doc)
+      try {
+        handler(payload.doc)
+      } catch (e) {
+        logger.error(`handler did throw for event ${event}`)
```
y-lohse: Nit, but we could also log the doctype and id to help with debuging?

PR #980, opened 2020-02-27
fix: Drive link from harvest

y-lohse:
The link to drive was missing a `#` at the start of the path.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/appLinksProps.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/appLinksProps.jsx
@@ -3,7 +3,7 @@ import get from 'lodash/get'
 const appLinksProps = {
   drive: ({ trigger }) => ({
     slug: 'drive',
-    path: `/files/${get(trigger, 'message.folder_to_save')}`,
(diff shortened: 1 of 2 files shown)


ptbrowne:
Tu pourras mettre à jour la branche beta de la home avec ctte nouvelle version ?  https://github.com/cozy/cozy-home/pull/1460. Merci

y-lohse:
> I think we should use generateWebLink from cozy-client

The thing is, we don't want to store a fully built URL at this point, because we will pass this data (slug, path, url…) to `useAppLinkWithStoreFallback` to get a store link back if needed. Storing a full URL here would prevent that.

I'm not sure we can change this, and if we do I think it falls outside the scope of this PR which is supposed to be a fast fix for an issue we have in production :)

Crash--:
> The thing is, we don't want to store a fully built URL at this point, because we will pass this data (slug, path, url…) to useAppLinkWithStoreFallback to get a store link back if needed. Storing a full URL here would prevent that.

You can still pass those info to useAppLinkWithStoreFallback. And do the work in this hook.

I'm not fully agreed on the fact that this is not the scope of this PR. In fact we have this bug because we create the URL manually without using our internal method  🤷‍♂

PR #984, opened 2020-03-02
Two small bugs in harvest

ptbrowne:
- Error info in configuration tab would not disappear after successful connection.

This was due to different `ConnectionFlow`s used between two components. To remedy
this problem, LaunchTriggerCard gains the ability to use a connectionFlow passed
from above

- Undefined in success URL for budget insight konnectors

Since at the time of login_success event, no trigger has been made, we have to
create our own trigger like object.


packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
```diff
@@ -10,83 +10,76 @@ import Uppercase, { Text } from 'cozy-ui/transpiled/react/Text'
...
-                </div>
-              </div>
-            )
-          }}
-        </FlowProvider>
-      </Card>
-    )
+const LaunchTriggerCard = props => {
+  if (props.flow) {
+    return <DumbLaunchTriggerCard {...props} flow={props.flow} />
```
y-lohse: Why add the `flow` prop separately ?

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
```diff
@@ -10,83 +10,76 @@ import Uppercase, { Text } from 'cozy-ui/transpiled/react/Text'
...
   }
+  return (
+    <FlowProvider {...pick(props, Object.keys(DumbFlowProvider.propTypes))}>
```
y-lohse: How do you feel about changing this `pick()` call? I believe it's the one that caused problems when we removed the proptypes from the built packages, and it also makes it hard to understand what props are being passed to `FlowProvider` without looking at it's proptypes.

packages/cozy-harvest-lib/src/locales/en.json
```diff
@@ -51,7 +51,7 @@
       },
       "banks": {
         "title": "Your banking data",
-        "description": "This service retrieves your latest baking operations and keeps a complete record for you.",
+        "description": "This service retrieves and keeps a complete record your latest banking operations for you.",
```
y-lohse:
```suggestion
        "description": "This service retrieves and keeps a complete record of your latest banking operations for you.",
```

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
ptbrowne: I'd be happy to remove it, it's a smell.

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
ptbrowne: No good reason ;)

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
ptbrowne: I [changed the API to use flowProps](https://github.com/cozy/cozy-libs/pull/984/commits/cef9dba413d0f99346a9835a08e13e77df835779) instead of directly spreading with pick + FlowProvider.propTypes. Backward compatibility is conserved with props.initialTrigger.

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.spec.jsx
```diff
@@ -0,0 +1,56 @@
...
+      <CozyProvider client={client}>
+        <I18n dictRequire={() => enLocale} lang="en">
+          <LaunchTriggerCard {...props} />
+        </I18n>
+      </CozyProvider>
+    )
+    return { root }
+  }
+
+  it('should creates a flow automatically if props.flow is not passed from above', () => {
```
y-lohse:
```suggestion
  it('should create a flow automatically if props.flow is not passed from above', () => {
```

PR #988, opened 2020-03-06
fix: Last execution time is correctly refreshed

ptbrowne:
LaunchTriggerCard bases its render on ConnectionFlow so it must
be explicit about that otherwise updaes from above could be 
"bailed out".


ptbrowne:
Sorry I should have put that in the description: I am not sure how to test for this specifig bug since the bug was due to how React handles update. Nonetheless, I saw that withConnectionFlow does not have any test, I will add one, thanks.

PR #989, opened 2020-03-09
fix: Call revokeSelf if not the owner of the sharing

Crash--:
Currently we're not doing the right thing when we try to remove a sharing from the ActionMenu. 

We display the "Remove action" with the right conditions aka "I'm the owner" or "I've accepted the cozy to cozy sharing", but after that, when we call the revoke() we don't make any difference between if I'm the owner or not. 

If I'm the owner, I can revoke members I want for the sharing and so, I need to call /revoke/:id. 
If I'm the "accepted sharing", I can only remove myself and then I need to call /revoke/self

### DIFF
--- a/packages/cozy-sharing/src/SharingDetailsModal.jsx
+++ b/packages/cozy-sharing/src/SharingDetailsModal.jsx
@@ -22,7 +22,8 @@ export class SharingDetailsModal extends Component {
       recipients,
       document,
       documentType = 'Document',
-      onRevoke
+      onRevoke,
+      onRevokeSelf
     } = this.props
     return (
       <Modal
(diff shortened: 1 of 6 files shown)


Crash--:
We miss it... :D

edas:
>  I need to call /revoke/:id.
> If I'm the "accepted sharing", I can only remove myself and then I need to call /revoke/self

But calling /revoke/:id on your own id should also work, doesn't it?

Crash--:
>.But calling /revoke/:id on your own id should also work, doesn't it?

I don't think so. It's not clear in the doc if it should work or not (https://docs.cozy.io/en/cozy-stack/sharing/#delete-sharingssharing-idrecipientsindex ping @nono ), but it seems easier to call https://docs.cozy.io/en/cozy-stack/sharing/#delete-sharingssharing-idrecipientsself to revoke myself from a sharing. Having to deal with the user's index in the member arrays is not straightforward.

nono:
Well, in theory, it should work, but currently, it doesn't: this route can be only be called on the cozy instance of the sharer.

Crash--:
added a note here https://github.com/cozy/cozy-stack/pull/2428

PR #991, opened 2020-03-09
fix: Bad password matching would result in null fields

ptbrowne:
Since the sort does not specify a number, false or true are used and falsy
values come first, this mean the search can find a cipher with a different
   password that what is searched. Another check is necessary if no cipher
   with the same password have ever been added.

   For a reason I still ignore, this resulted in ciphers with null values 
   after both ciphers with different password were merged because the 
   deciphering would not work (mac failed error)

### DIFF
--- a/packages/cozy-harvest-lib/src/models/cipherUtils.js
+++ b/packages/cozy-harvest-lib/src/models/cipherUtils.js
@@ -169,16 +169,21 @@ const searchForCipher = async (vaultClient, searchOptions) => {
 
   const cipher = await vaultClient.getByIdOrSearch(id, search, sort)
 
-  // If there is no id, we do strict password matching
-  if (!id) {
-    // Since getByIdOrSearch does not support password matching,
-    // we have to do the password matching here
+  if (id) {
+    // We know the cipher we are targetting, so we do not need to check
+    // for password equality. Checking for password inequality would
+    // prevent password update through Harvest.
+    return cipher
+  } else {
+    // If there is no id, we do strict password matching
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/models/cipherUtils.js
```diff
@@ -161,7 +161,16 @@ const searchForCipher = async (vaultClient, searchOptions) => {
...
-  const sort = [view => view.login.password === password, 'revisionDate']
+  const sort = [
+    view => (view.login.password === password ? 0 : 1),
+    'revisionDate'
+  ]
   const cipher = await vaultClient.getByIdOrSearch(id, search, sort)
-  return cipher || null
+  if (cipher) {
+    const decrypted = await vaultClient.decrypt(cipher)
+    if (decrypted.login.password == password) {
```
Crash--:
```suggestion
    if (decrypted.login.password === password) {
```

y-lohse: i think this part might warrant a comment, even with the PR description it took me a while to understand why we seem to do the check twice.
But I still find `getByIdOrSearch `… strange. We could make the code a lot more explicit if it returned more than one result for example. Anyway, not blocking for me but something to keep in mind.

ptbrowne: > But I still find getByIdOrSearch … strange. 

Yes I completely agree, I am not at ease with the goal and responsibilities of WebVaultClient.

PR #995, opened 2020-03-19
fix: Grid spacing

Crash--:
Drive was using MUI V4. We downgraded it to use V3 as UI. V3 and V4 don't have the same implementation.

In v4 spacing = 8 means 8*8, in v3 spacing 8 means what we want

### DIFF
--- a/packages/cozy-scanner/src/DocumentQualification.jsx
+++ b/packages/cozy-scanner/src/DocumentQualification.jsx
@@ -156,7 +156,7 @@ export class DocumentQualification extends Component {
             </Title>
           </div>
         )}
-        <Grid container spacing={1}>
+        <Grid container spacing={8}>
           <GridItem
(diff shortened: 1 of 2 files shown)


edas:
As I understand, spacing=8 means 8 x 8 px, so 64px. It seems way too much. https://material-ui.com/components/grid/

You could however change the conversion fonction in the theme, or add other values : https://material-ui.com/customization/spacing/

Crash--:
In fact I was wrong on the V4 implementation.

But we currently use V3 and spacing 8 means what we want https://v3.material-ui.com/layout/grid/

Crash--:
We have a peerDep on Cozy UI that already have a peerDep on Mui3. I think we already have this warning in fact... 

I'm not sure that adding a peerDep of a peerDep is a good solution for our lib

y-lohse:
Oh right, Grid is imported from cozy-ui, not MUI. No problem then.

PR #996, opened 2020-03-23
Fix/email adresses

Crash--:
We can now copy paste string containing email addresses. We'll extract the addresses and create contact if needed 

We check the email validity not only when writing but also when submitting. 

added: 
- PropTypes
- Few tests


packages/cozy-sharing/src/components/ShareByEmail.jsx
```diff
@@ -201,81 +190,83 @@ class ShareByEmail extends Component {
...
-      .then(recipients =>
-        onShare(document, recipients, sharingType, sharingDesc)
+    try {
+      const allCreatedRecipients = await Promise.all(
+        verifiedContacts.map(recipient =>
+          recipient.id
+            ? recipient
+            : createContact({
+                email: [{ address: recipient.email, primary: true }]
+              }).then(resp => resp.data)
```
y-lohse: Could we use async/await here?

packages/cozy-sharing/src/helpers/email.js
```diff
@@ -0,0 +1,11 @@
+// Extract emails address from a string
+export const extractEmails = str => {
+  if (typeof str === 'string')
+    return str.match(/([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+\.[a-zA-Z0-9_-]+)/gi)
```
y-lohse: `a-zA-Z0-9._-` can be replaced by `\w.-`

packages/cozy-sharing/src/helpers/email.js
```diff
@@ -0,0 +1,11 @@
...
+  return null
+}
+// Validate if a string is an email address or not
+export const validateEmail = str => {
+  const reg = /^(([^<>()[\].,;:\s@"]+(\.[^<>()[\].,;:\s@"]+)*)|(".+"))@(([^<>()[\].,;:\s@"]+\.)+[^<>()[\].,;:\s@"]{2,})$/i
```
y-lohse: Can we have a short description of what this baby is looking for? Not gonna lie, it's a bit hard to decipher 😁

packages/cozy-sharing/src/components/ShareByEmail.jsx
Crash--: old commit. already fixed in a more recent commit ;)

PR #997, opened 2020-03-23
Sharing hooks

y-lohse:
At the moment cozy-sharings exposes a variety of components that in turn expose a subset of the context. That's cool but doesn't work with hooks at all.

This PR exposes the `SharingContext` so that it can be called with `useContext`.  
It also exposes the `hasWriteAccess` function in the context so that it can be used outside of `<SharedDocument />`.

### DIFF
--- a/packages/cozy-sharing/README.md
+++ b/packages/cozy-sharing/README.md
@@ -2,12 +2,61 @@
 
 Open the playgrounds in cozy-libs and run `yarn start`
 
-# How to use the lib 
+# How to use the lib
 
-In your app, you have to : 
+In your app, you have to :
 
-- import the Provider: ```import SharingProvider, { ShareButton, ShareModal } from 'cozy-sharing'```
-- import the sylesheet.css: ```import 'cozy-sharing/dist/stylesheet.css'```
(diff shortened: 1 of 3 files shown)


packages/cozy-sharing/README.md
```diff
@@ -2,12 +2,26 @@
...
-# How to use the lib 
+# How to use the lib
 
-In your app, you have to : 
+In your app, you have to :
 
-- import the Provider: ```import SharingProvider, { ShareButton, ShareModal } from 'cozy-sharing'```
-- import the sylesheet.css: ```import 'cozy-sharing/dist/stylesheet.css'```
+- import the Provider: `import SharingProvider, { ShareButton, ShareModal } from 'cozy-sharing'`
+- import the sylesheet.css: `import 'cozy-sharing/dist/stylesheet.css'`
```
ptbrowne:
```suggestion
- import the stylesheet.css: `import 'cozy-sharing/dist/stylesheet.css'`
```

In the example, it is not clear why we need CSS, could there be an example showing
a component needing CSS ?

PR #1006, opened 2020-04-08
fix(Harvest): Some regressions which affected the 2FA flow

doubleface:
Keep watching connection flow when connector execution is finished
Give connector information to the flow in all the contexts to allow it to get the associated policies

### DIFF
--- a/packages/cozy-harvest-lib/src/components/FlowProvider.jsx
+++ b/packages/cozy-harvest-lib/src/components/FlowProvider.jsx
@@ -94,7 +94,6 @@ export class FlowProvider extends Component {
     if (this.state.showTwoFAModal) {
       this.dismissTwoFAModal()
     }
-    this.stopWatchingConnectionFlow()
 
     const { onError } = this.props
     if (typeof onError === 'function') onError(error)
@@ -113,7 +112,6 @@ export class FlowProvider extends Component {
     if (this.state.showTwoFAModal) {
       this.dismissTwoFAModal()
     }
-    this.stopWatchingConnectionFlow()
     const { onSuccess } = this.props
     const flow = this.flow
     if (typeof onSuccess === 'function') onSuccess(flow.trigger)
@@ -171,7 +169,7 @@ FlowProvider.propTypes = {
    */
   client: PropTypes.object.isRequired,
   /**
-   * konnector manifest
+   * Konnector manifest
    */
   konnector: PropTypes.object.isRequired,
   /**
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -135,9 +135,10 @@ export const createOrUpdateAccount = async ({
  * This should be the go to source of truth for the state of a Konnector Job.
  */
 export class ConnectionFlow {
-  constructor(client, trigger) {
+  constructor(client, trigger, konnector) {
     this.client = client
     this.trigger = trigger
+    this.konnector = konnector
```
y-lohse, line 141: Where do we use `this.konnector`?

doubleface: https://github.com/cozy/cozy-libs/blob/686036095b80a4204e6ab0bd50cfe13e9bcafbb4/packages/cozy-harvest-lib/src/models/ConnectionFlow.js#L533

packages/cozy-harvest-lib/src/components/FlowProvider.jsx
```diff
@@ -170,6 +170,10 @@ FlowProvider.propTypes = {
    * CozyClient instance
    */
   client: PropTypes.object.isRequired,
+  /**
+   * konnector manifest
```
ptbrowne, line 172:
```suggestion
   * Konnector manifest
```

packages/cozy-harvest-lib/test/components/__snapshots__/TriggerManager.spec.js.snap
```diff
@@ -123,6 +123,7 @@ exports[`TriggerManager when given an oauth konnector should redirect to OAuthFo
       "handleAccountUpdated": [Function],
       "handleJobUpdated": [Function],
       "handleTwoFA": [Function],
+      "konnector": undefined,
```
ptbrowne, line 126: Is it normal ?

ptbrowne: I see no tests testing the fix, could you add one please ?

doubleface: It was not normal, I fixed the unit tests and added one unit test to check that flow knows the konnector.

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
```diff
@@ -66,7 +66,7 @@ const tMock = jest.fn()
 const client = new CozyClient({})
 const props = {
   konnector: fixtures.konnector,
-  flow: new ConnectionFlow(client),
+  flow: new ConnectionFlow(client, undefined, fixtures.konnector),
```
ptbrowne, line 69: Nit: At this point I am wondering if an option object would not be better, I had trouble remembering that the second argument is a trigger.

PR #1007, opened 2020-04-15
Sharing link with write permissions

y-lohse:
This PR updates the link sharing UI. Instead of a simple toggle, it's now a 2 step process:

- A button to activate the sharing
- An ActionMenu to change the permissions

Capture d'écran 2020-04-15 10 00 22
Capture d'écran 2020-04-15 10 01 06

### DIFF
--- a/packages/cozy-sharing/locales/en.json
+++ b/packages/cozy-sharing/locales/en.json
@@ -64,11 +64,19 @@
       "sharedWithMe": "Shared with me",
       "sharedBy": "Shared by %{name}",
       "shareByLink": {
-        "subtitle": "By public link",
-        "creating": "Creating your link...",
+        "subtitle": "Or share by link",
+        "create": "Generate a link",
+        "activated": "Link activated",
         "copy": "Copy link",
         "copied": "Link has been copied to clipboard",
-        "failed": "Unable to copy to clipboard"
+        "failed": "Unable to copy to clipboard",
+        "ro": "Can read",
+        "rw": "Can change",
+        "desc": {
+          "ro": "Link holders can only read or download content.",
+          "rw": "Link holders can alter and delete content."
+        },
+        "deactivate": "Deactivate link"
       },
       "shareByEmail": {
         "subtitle": "By email",
@@ -139,12 +147,20 @@
       "sharedWithMe": "Shared with me",
       "sharedBy": "Shared by %{name}",
       "shareByLink": {
-        "subtitle": "By public link",
+        "subtitle": "Share by link",
         "desc": "Anyone with the provided link can see and edit your note.",
-        "creating": "Creating your link...",
+        "create": "Generate a link",
+        "activated": "Link activated",
         "copy": "Copy link",
         "copied": "Link has been copied to clipboard",
-        "failed": "Unable to copy to clipboard"
+        "failed": "Unable to copy to clipboard",
+        "ro": "Can read",
+        "rw": "Can change",
+        "desc": {
+          "ro": "Link holders can only read or download content.",
+          "rw": "Link holders can alter and delete content."
+        },
+        "deactivate": "Deactivate link"
       },
       "shareByEmail": {
         "subtitle": "By email",
@@ -211,13 +227,21 @@
       "sharedByMe": "Shared",
       "sharedWithMe": "Shared with me",
       "shareByLink": {
-        "subtitle": "By public link",
+        "subtitle": "Or share by link",
         "desc": "Anyone with the provided link can see and download your photos.",
         "fetchFailed": "Woops! It seems your connectivity is limited, try again later when it gets better.",
-        "creating": "Creating your link...",
+        "create": "Generate a link",
+        "activated": "Link activated",
         "copy": "Copy link",
         "copied": "Link has been copied to clipboard",
-        "failed": "Unable to copy to clipboard"
+        "failed": "Unable to copy to clipboard",
+        "ro": "Can read",
+        "rw": "Can change",
+        "desc": {
+          "ro": "Link holders can only read or download content.",
+          "rw": "Link holders can alter and delete content."
+        },
+        "deactivate": "Deactivate link"
       },
       "shareByEmail": {
         "title": "By email",
--- a/packages/cozy-sharing/locales/en.json
+++ b/packages/cozy-sharing/locales/en.json
@@ -64,7 +64,6 @@
       "sharedBy": "Shared by %{name}",
       "shareByLink": {
         "subtitle": "By public link",
-        "desc": "Anyone with the provided link can see and download your files.",
         "creating": "Creating your link...",
         "copy": "Copy link",
         "copied": "Link has been copied to clipboard",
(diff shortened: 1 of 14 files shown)


packages/cozy-sharing/locales/fr.json
```diff
@@ -211,13 +227,21 @@
...
+        "activated": "Lien activé",
         "copy": "Copier le lien",
         "copied": "Lien copié dans le presse-papier",
-        "failed": "Impossible de copier dans le presse-papier"
+        "failed": "Impossible de copier dans le presse-papier",
+        "ro": "Peut consulter",
+        "rw": "Peut modifier",
+        "desc": {
+          "ro": "Les détenteurs du lien peuvent uniquement consulter ou télécharger le contenu.",
+          "rw": "Les détenteurs du lien peuvent modifier et supprimer le contenu. "
```
Crash--, line 242:
```suggestion
          "rw": "Les détenteurs du lien peuvent modifier et supprimer le contenu."
```

packages/cozy-sharing/src/components/ShareByLink.jsx
```diff
@@ -66,47 +77,163 @@ class ShareByLink extends React.Component {
...
+    } = this.props
+    const permissionCategories = get(
+      permissions,
+      '[0].attributes.permissions',
+      {}
+    )
+    const hasWritePermissions =
+      Object.values(permissionCategories).filter(permissionCategory =>
+        get(permissionCategory, 'verbs', []).includes('POST')
+      ).length > 0
```
Crash--, line 102: can we extract this logic from the render? And having a better place then the render to read it?

packages/cozy-sharing/src/components/ShareByLink.jsx
```diff
@@ -66,47 +77,163 @@ class ShareByLink extends React.Component {
     }
   }
 
+  toggleMenu = () => {
+    this.setState(state => ({ ...state, menuIsOpen: !state.menuIsOpen }))
+  }
+
   render() {
     const t = this.context.t
```
Crash--, line 85: I know this is not the scope of the PR but removing the context access can be a great improvement when working on an old file

packages/cozy-sharing/src/index.jsx
```diff
@@ -213,6 +218,27 @@ export class SharingProvider extends Component {
...
+        const resp = await this.props.client
+          .collection('io.cozy.permissions')
+          .add(permissionDocument, updatedPermissions)
+        this.dispatch(updateSharingLink(resp))
+        return resp
+      })
+    )
+
+    return responses[0]
+  }
```
Crash--, line 240: We handle several permissions, but we only return the result of the first one?

Crash--: Also this method is more generic than just "updateSharingLinkPermissions". It's more "updateDocumentPermissions", no ?

packages/cozy-sharing/src/index.jsx
```diff
@@ -213,6 +218,27 @@ export class SharingProvider extends Component {
     return resp
   }
 
+  updateSharingLinkPermissions = async (document, newVerbs) => {
```
Crash--, line 221: can we have a doctring for this method plz

packages/cozy-sharing/src/index.jsx
Crash--: No catch ?

y-lohse: > We handle several permissions, but we only return the result of the first one?

Since having one document with multiple permissions is not a "normal" state, I didn't want to propagate it too much down the chain. But tthe other methods do it as well, so I'll update this one to stay consistent.

> Also this method is more generic than just "updateSharingLinkPermissions". It's more "updateDocumentPermissions", no ?

Yes, you're right.

> No catch ?

Not here… but actually there should be one in `ShareByLink`, i'll add it.

packages/cozy-sharing/src/components/ShareByLink.jsx
y-lohse: I extracted the logic but kept the reading in the render, as I think it's the right place because It's an information derived from a prop. Do you think it would be better elsewhere?

ptbrowne:
In the screenshot, "Peut consulter" seems a bit too close to the arrow, and it seems darker than the text to its left.

y-lohse:
> In the screenshot, "Peut consulter" seems a bit too close to the arrow, and it seems darker than the text to its left.

The arrow spacing has been checked with Joel and is ok (and I don't feel like it's too close?). But yes you're right the color is off, I'll look into it.

ptbrowne:
AMHA



feels more balanced than

| current | tweaked |
|----|----|
image | image |

I tweaked a bit the spacing on the styleguide:

```
.c-DropdownButton-Icon--ohvD2 {
    margin-top: 0.125rem;
    margin-left: 0.5rem;
}
```

The problem is that those margins should depend on the size of the text next to it (which is not possible right now).

Also, on your screenshots it seems we are not using Lato.

y-lohse:
I corrected the color problem. 

Capture d'écran 2020-04-16 09 49 04

The font issue is a problem on my local machine that I hadn't noticed before, apparently I'm using an empty theme. I'll fix it one of these days :)

> The problem is that those margins should depend on the size of the text next to it (which is not possible right now).

Yes, this is why Joel found a value that worked well enough with all sizes, even if it's not perfect. It's the sort of thing we will likely be able to fix with a cozy-ui theme.

ptbrowne:
OK, I opened [an issue](https://github.com/cozy/cozy-ui/issues/1433) on the matter. I do not see how this could be solved via a theme 🤔.

packages/cozy-sharing/src/components/ShareByLink.jsx
```diff
@@ -1,22 +1,43 @@
...
+    '[0].attributes.permissions',
+    {}
+  )
+  return (
+    Object.values(permissionCategories).filter(permissionCategory =>
+      get(permissionCategory, 'verbs', []).includes('POST')
+    ).length > 0
+  )
+}
+
```
Crash--, line 31: Since this API is not straightforward from a DX POV, we should add this kind of method to cozy-client directly. Either in the PermissionCollection or in a PermissionModel, no? 

I know we already have this kind of code in Notes, and we should mutualize.

y-lohse: I updated the code to use the existing permission model, which did pretty much the same. I did keep the multi-permission check in here because I think this is specific to cozy-sharings.

packages/cozy-sharing/locales/en.json
```diff
@@ -211,13 +228,21 @@
...
+        "activated": "Link activated",
         "copy": "Copy link",
         "copied": "Link has been copied to clipboard",
-        "failed": "Unable to copy to clipboard"
+        "failed": "Unable to copy to clipboard",
+        "ro": "Can read",
+        "rw": "Can change",
+        "desc": {
+          "ro": "Link holders can only read or download content.",
+          "rw": "Link holders can alter and delete content."
```
Crash--, line 243:
```suggestion
          "rw": "Link holders can change and delete content."
```

Are we sure about `alter`? We use `change` or `edit` in other places.

PR #1008, opened 2020-04-28
Fix/sharing

Crash--:
- Too much warnings in the console, let's fix that


nono:
Do we know why the travis build is red?

PR #1013, opened 2020-05-27
Sharing/realtime

Crash--:
Mainly two things in this PR : 

- First, realtime on the sharing doctype 
- Second: fix a bug when a sharing is added twice (can happen with the realtime now, but also if you close and reopen the sharing modal)

### DIFF
--- a/packages/cozy-sharing/src/ducks/sharings.js
+++ b/packages/cozy-sharing/src/ducks/sharings.js
@@ -1,0 +1,38 @@
+import {
+  mergeObjectFromRealTime,
+  createFakeInternalObjectFromRealtime
+} from '../helpers/realtime'
+
+import { addSharing, updateSharing } from '../state'
+import { getFilesPaths } from '../helpers/files'
+
+export const getSharingObject = (internalSharing, sharing) => {
+  if (internalSharing) {
+    return mergeObjectFromRealTime(internalSharing, sharing)
+  } else {
+    return createFakeInternalObjectFromRealtime(sharing, 'io.cozy.sharings')
+  }
+}
+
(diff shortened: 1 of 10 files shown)


packages/cozy-sharing/src/state.js
```diff
@@ -26,11 +26,13 @@ export const receiveSharings = ({
     apps
   }
 })
-export const addSharing = (data, path) => ({
-  type: ADD_SHARING,
-  data,
-  path
-})
+export const addSharing = (data, path) => {
+  return {
```
y-lohse, line 30: Nit, but that seems like an unnecessary change?

packages/cozy-sharing/src/state.js
```diff
@@ -87,15 +89,19 @@ const updateByIdItem = (state, id, updater) => {
...
-  )
+const indexSharing = (state = {}, sharing) => {
+  const sharedDocs = getSharedDocIds(sharing)
+  return sharedDocs.reduce((byId, id) => {
+    const updatedByIdItem = updateByIdItem(byId, id, state => ({
+      ...state,
+      //todo index by id instead of having an array
+      sharings: Object.values(state.sharings).includes(sharing.id)
+        ? [...state.sharings]
+        : [...state.sharings, sharing.id]
```
y-lohse, line 100: Maybe we could run the result through `uniq` to make things more explicit?

packages/cozy-sharing/src/ducks/sharings.js
```diff
@@ -0,0 +1,38 @@
+import {
```
y-lohse, line 1: Nit: I think this is more a helper file than a duck

PR #1014, opened 2020-05-28
Support additional fields

ptbrowne:
- Additional fields are saved into the vault's cipher from Harvest
- Additional fields are updated into the io.cozy.accounts when the cipher is updated


packages/cozy-harvest-lib/src/models/cipherUtils.js
```diff
@@ -79,6 +91,23 @@ export const updateCipher = async (vaultClient, cipherId, data) => {
   }
 }
 
+const isAdditionalField = (fieldName, { identifierProperty }) => {
```
y-lohse, line 94: Nit: why do we use one named parameter, and one that isn't here?

packages/cozy-harvest-lib/src/models/cipherUtils.js
```diff
@@ -6,6 +6,12 @@ import logger from '../logger'
 
 import { CipherType, UriMatchType } from 'cozy-keys-lib'
 
+const fieldTypes = {
```
y-lohse, line 9: Nit: these field types are not exposed through `cozy-keys-lib`?

packages/cozy-harvest-lib/src/models/cipherUtils.js
ptbrowne: Because to me it is expected that isAdditionalField would run on a field name, and then identifierProperty is part of options.

packages/cozy-harvest-lib/src/models/cipherUtils.js
ptbrowne: Not yet FieldType

PR #1017, opened 2020-06-01
feat: add command completion feature with cli-tree with tabtab

doubleface:
This adds completion to cli-tree using [tabtab](https://www.npmjs.com/package/tabtab).

Completion installation commands are automatically present.
Need to add the `completionHandler` in the program.

I would have liked to make it transparent and not need any change in the program using cli-tree but the `completionHandler` needs to be async (because of tabtab implementation) and `build` is sync.

Other possibility I saw is to add a new `completeAndBuild` function at the same level as build, which would be async.
This PR may be the occasion for other ideas.

### DIFF
--- a/packages/cli-tree/src/completion.js
+++ b/packages/cli-tree/src/completion.js
@@ -6,27 +6,7 @@ const _ = require('lodash')
 
 module.exports = {
   getCompletionSetupCommands,
-  completionHandler: async (commands = {}) => {
-    const env = tabtab.parseEnv(process.env)
-
-    if (!env.complete) {
-      return false
-    }
-
-    const command = findCommand(
-      { _main: commands },
-      '_main ' +
-        env.line
-          .split(' ')
-          .slice(1)
-          .join(' ')
-    )
-    if (!command) process.exit(1)
-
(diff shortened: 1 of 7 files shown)


packages/cli-tree/src/completion.js
```diff
@@ -0,0 +1,84 @@
+/* eslint no-console: off */
+const tabtab = require('tabtab')
+const get = require('lodash/get')
+const toPairs = require('lodash/toPairs')
+const _ = require('lodash')
+
+module.exports = {
+  getCompletionSetupCommands,
+  completionHandler: async (commands = {}) => {
```
ptbrowne, line 9: Can we have the completion handler at module level, like the other functions that are exported ?

doubleface: I am not sure to understand. the completionHandler is available at the module level. You would prefer to have the completionHandler code in index.js ?

ptbrowne: I meant transforming into:

```

module.exports = {
  getCompletionSetupCommands,
  completionHandler,
  ...
}

const completionHandler = async (commands = {}) => {
  ...
}
```

PR #1021, opened 2020-06-04
Correctly save fields after logging in from a cipher

ptbrowne:
When adding an account from a cipher, and changing an additional
field, the additional field would not be changed since login
and password were the same.

Here, we check also to see if fields have change, to know if we
need to update the cipher.


packages/cozy-harvest-lib/src/components/KonnectorModal.jsx
```diff
@@ -319,7 +318,6 @@ KonnectorModal.propTypes = {
     })
   }).isRequired,
   findAccount: PropTypes.func.isRequired,
```
y-lohse, line 320: I think `findAccount ` was injected via `accountsMutation` but I don't see it imported anywhere, did I miss something?

ptbrowne: 😱 Well spotted, I edited https://github.com/cozy/cozy-libs/pull/1021/commits/440c0b1b3e635adfd872df2537e5acbd7368c2da to account for the removal of withMutations.

PR #1023, opened 2020-06-08
fix: Remove initialization of flag did not work

ptbrowne:
GET was not passed as first argument of fetchJSON :(

### DIFF
--- a/packages/cozy-flags/src/flag.js
+++ b/packages/cozy-flags/src/flag.js
@@ -63,7 +63,7 @@ export const enable = flagsToEnable => {
 export const initializeFromRemote = async client => {
   const {
     data: { attributes }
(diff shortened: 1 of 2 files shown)


packages/cozy-flags/src/flag.js
```diff
@@ -63,7 +63,7 @@ export const enable = flagsToEnable => {
 export const initializeFromRemote = async client => {
   const {
     data: { attributes }
-  } = await client.stackClient.fetchJSON('/settings/flags')
+  } = await client.stackClient.fetchJSON('GET', '/settings/flags')
```
Crash--, line 66: Can't we use client.get('io.cozy.settings', 'flags')?

ptbrowne: We could, is it blocking for you ?

PR #1019, opened 2020-06-03
Docs: Add doc about CozyClient Plugin

Crash--:
Let's talk about the Realtime Plugin for CozyClient

PR #1027, opened 2020-06-17
docs: Usage with cozy-client

y-lohse:
Documents usage with cozy-client as a plugin (+typo corrections).

### DIFF
--- a/packages/cozy-realtime/README.md
+++ b/packages/cozy-realtime/README.md
@@ -59,16 +59,29 @@ const handleUpdate = accounts => {
 
 // To subscribe
 await realtime.subscribe('created', type, handleCreate)
-await realtime.subscribe('updated', type, handleCreate)
-await realtime.subscribe('updated', type, id, handleUpdate)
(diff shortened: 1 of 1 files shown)


packages/cozy-realtime/README.md
```diff
@@ -59,18 +59,31 @@ const handleUpdate = accounts => {
...
+await realtime.unsubscribe('updated', type, handleUpdate)
+await realtime.unsubscribe('updated', type, id, handleSingleUpdate)
 
 // To unsubscribe all
 await realtime.unsubscribeAll()
 ```
 
+## Usage with cozy-client
+
+Instead of passing an instance of cozy-client to cozy-realtime, it is possible to provide cozy-realtime as a plugin to cozy-client.
```
ptbrowne, line 76: This should be the first paragraph as it is the recommended way to do things.

Crash--:
See https://github.com/cozy/cozy-libs/pull/1019  ...

PR #1031, opened 2020-07-07
fix: Don't load data if the client isn't ready

y-lohse:
The SharingProvider loads data on `componentDidMount` — this may lead to errors when cozy-client is not connected, for example on a mobile app.


y-lohse:
I think we're now handling things in a better way :). Depends on #1034.

y-lohse:
Should be all good now!

PR #1035, opened 2020-07-09
Repository doctor

ptbrowne:
I think it'd be useful to have a tool that performs automated
checks on our repositories to guarantee the coherence. Examples:

- Check that travis.yml follows our rules
- Check that our dependencies (only cozy packages) are up-to-date
- Check that the file structure is the same everywhere
- Check that our best practices concerning eslint / jest and other libs are respected

In this PR I add a small dependency checker. It fetches the
last version for dependencies listed in 'dependencies.json'
and fetches dependencies from package.json for repositories
listed in 'repositories.json'. It then outputs the diff
between the last version of the package and the version
from the repository. See the output below:

This could be run automatically by a bot and the output could be sent to mattermost.

Example run:

Capture d’écran 2020-07-10 à 17 59 15

## Features

- Ability to report results to console or mattermost
- Rules can be configured through JSON file and through CLI args
- Files are checked through the GitHub API, instead of cloning repositories (faster)


packages/repo-doctor/dependencies.json
```diff
@@ -0,0 +1,6 @@
+[
+  "cozy-ui",
+  "cozy-realtime",
+  "cozy-harvest-lib",
+  "cozy-flags"
```
Crash--, line 5: cozy-client ?

ptbrowne: Of course :) Fixed :)

packages/repo-doctor/src/fetch.js
```diff
@@ -0,0 +1,67 @@
...
+  version: depEntry[1],
+  type
+})
+
+const fetchRepositoryInfo = async repository => {
+  const slug = repository.slug
+  const packageJsonPath = repository.packageJsonPath || 'package.json'
+  const packageJsonURL = `https://raw.githubusercontent.com/${slug}/master/${packageJsonPath}`
+  const packageJsonText = await fetch(packageJsonURL).then(resp => resp.text())
+  const packageJsonData = JSON.parse(packageJsonText)
```
y-lohse, line 14: Could you do `resp.json()` instead of going text->JSON or does it not work?

packages/repo-doctor/src/reporters/mattermost.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  write(msg) {
+    this.messages.push(msg)
+  }
+
+  async sendAllMessages() {
+    const payload = {
+      channel: '@patrick',
```
y-lohse, line 30: I guess ideally this would be configurable through a command line arg?

packages/repo-doctor/src/fetch.js
```diff
@@ -0,0 +1,67 @@
+const fetch = require('node-fetch')
```
y-lohse, line 1: I think it's a bit weird that there is a root file called `fetch`, that has *some* fetching in it, but not all of it (ie. the locales fetching). Not sure what the solution is here, i'm still wrapping my head around the codebase, but just pointing it out fwiw.

packages/repo-doctor/src/rules/locales.js
```diff
@@ -0,0 +1,31 @@
...
+  return await fetch(
+    `https://api.github.com/repos/${repoSlug}/contents/${pathName}`
+  ).then(resp => resp.json())
+}
+
+const localesInRepo = () =>
+  async function*(repositoryInfo) {
+    const localeDirContent = await fetchRepositoryDirectoryContent(
+      repositoryInfo.slug,
+      'src/locales'
```
y-lohse, line 13: I don't think this will work out, the locales for drive and photos for example are in the same repo but not the same folder, and packages such as cozy-sharing won't work either I think?

packages/repo-doctor/src/toolbelt.js
```diff
@@ -0,0 +1,16 @@
...
+  return res
+}
+
+const keyBy = (values, fn) => {
+  return fromEntries(values.map(v => [fn(v), v]))
+}
+
+module.exports = {
+  fromEntries,
+  keyBy
```
y-lohse, line 15: memoize could probably go in here as well, through i'm surprised we can't use lodash for any of these.

packages/repo-doctor/src/rules/locales.js
ptbrowne: This is why I think rules could be configured per repo.

packages/repo-doctor/src/reporters/mattermost.js
ptbrowne: In the config I think it would be better : { reporters: { mattermost: { channel: '' }}

packages/repo-doctor/src/fetch.js
ptbrowne: I moved the utility fetch function from locales.js to fetch.js.

packages/repo-doctor/src/toolbelt.js
ptbrowne: Yep, I don't know why I recoded these functions instead of using lodash.

packages/repo-doctor/src/reporters/mattermost.js
ptbrowne: I added this ability in https://github.com/cozy/cozy-libs/pull/1035/commits/5d9f32a5285534623326029c878f813fd12a9524

ptbrowne:
> It's not easy to understand the architecture just by reading the code

Do you have any advice on what has been difficult to understand ?

> it happens in 2 stages but it's not quite clear what gets configured when. Is it necessary to have this separation, or could the options be consolidated before being fed to the rules? 

I think each rule should know how to handle CLI args. Otherwise it's the responsibility of the rule "orchestrator" (index.js) to know how to pass CLI args to each rule. I think it would introduce too much coupling between the two. Here I think what I would like to do is: "here are some known CLI args like --dep and --repo : if your rule has any business with one of this concept, please implement the necessary filtering in the rule".

> I also think that some docstrings are necessary so we don't have to dive into the code to find out which config options are supported by which rules etc.

I would prefer to use config validation for this. This way we express the schema of the config in the code. In a second time we could take this schema and render it into documentation. WDYT ?

ptbrowne:
I think function based rule were cute but will not scale and can be difficult to understand. What do you think of using class based rules : https://github.com/cozy/cozy-libs/pull/1035/commits/46e6cabfd7f2664dec6ff5130059c8776eaed4a3

y-lohse:
> > It's not easy to understand the architecture just by reading the code
> 
> Do you have any advice on what has been difficult to understand ?

The problem was mostly that I didn't know in what order to read the files, so I kept going back and forth. Maybe a small readme outlining the key concepts (cli entry point, rules, reporters) would be a good addition?

> > it happens in 2 stages but it's not quite clear what gets configured when. Is it necessary to have this separation, or could the options be consolidated before being fed to the rules?
> 
> I think each rule should know how to handle CLI args. Otherwise it's the responsibility of the rule "orchestrator" (index.js) to know how to pass CLI args to each rule. I think it would introduce too much coupling between the two. Here I think what I would like to do is: "here are some known CLI args like --dep and --repo : if your rule has any business with one of this concept, please implement the necessary filtering in the rule".

The coupling is a good point. My concern with the approach described above is with maintenance, as it could quickly become difficult to know which rules rely on what args, and so renaming/deprecating a rule becomes difficult as the deprecation needs to be kept around not only until all CIs have been updated, but also until all rules have been updated. It's sort of relatedto the next point though…

> > I also think that some docstrings are necessary so we don't have to dive into the code to find out which config options are supported by which rules etc.
> 
> I would prefer to use config validation for this. This way we express the schema of the config in the code. In a second time we could take this schema and render it into documentation. WDYT ?

Yep, totally ok with this. I was saying docstring but what I really meant was "some easy way for maintainers to figure out who's using what". Hopefully the config validation can also help with the deprecation issues mentioned above.

ptbrowne:
Good point, I've added a README.

As for the flags, I understand your concern, I think the --dep could be removed. It would simplify the code as rules would not have any relationship with the args. I think it can be useful to configure rules via the command line sometimes. It could be done as eslint does with its --rule flag. Interestingly, it uses a special format to configure a rule via the command line : https://eslint.org/docs/user-guide/command-line-interface#rule

```
repo-doctor --rule "dep-update: { dependencies: ["cozy-ui"] }"
```

packages/repo-doctor/src/rules/index.js
```diff
@@ -9,16 +9,19 @@ const ruleFns = { depUpToDate, noForbiddenDep, localesInRepo }
...
     } else {
       ruleName = rule
-      options = {}
+      ruleConfig = {}
     }
-    const ruleFn = ruleFns[ruleName]
-    return ruleFn(options, args)
+    let ruleFn = ruleFns[ruleName]
+
+    return ruleFn.prototype
```
y-lohse: Maybe we could use `isFunction` instead? I think it would make it clearer what we're checking for.

packages/repo-doctor/src/rules/index.js
```diff
@@ -19,6 +20,10 @@ const setupRules = (config, args) => {
     }
     let ruleFn = ruleFns[ruleName]
 
+    if (ruleFn.configSchema) {
+      validate(ruleFn.configSchema, {}, ruleConfig)
```
y-lohse: Not knowing schema-utils, I had to look up the docs to see that

- it throws an error if the config doesn't match
- the middle arg is an empty option

I think we could use a named var, and idk if we should do something with the exception, like wrapping it in a custom error message?

packages/repo-doctor/src/index.js
```diff
@@ -44,7 +47,11 @@ const main = async () => {
     repositories = repositories.filter(repo => repo.slug === args.repo)
   }
 
-  const rules = setupRules(config, args)
+  let rules = setupRules(config, args)
+  if (args.rule) {
+    rules = rules.filter(r => r.constructor.name === args.rule)
```
y-lohse, line 63: Could we use the name from the schema instead? Constructor names may differ with imports, exports, memoization or whatnot.

packages/repo-doctor/repo-doctor.json
```diff
@@ -0,0 +1,56 @@
...
+          "name": "eslint-loader",
+          "reason": "Used through cozy-scripts"
+        }
+      ]
+    }],
+    "LocalesInRepo"
+  ],
+  "reporters": {
+    "mattermost": {
+      "channel": "@patrick"
```
y-lohse, line 53: Is this file still used? If so I suggest we change this channel.

packages/repo-doctor/src/reporters/mattermost.js
```diff
@@ -0,0 +1,55 @@
...
+
+class MattermostReporter {
+  constructor(config) {
+    this.config = config
+    this.checkEnv()
+    this.messages = []
+  }
+
+  checkEnv() {
+    if (!process.env.MATTERMOST_HOOK) {
```
y-lohse, line 18: Wouldn't it make sense to have this variable as part of the configuration rather than an env variable?

packages/repo-doctor/src/rules/index.js
ptbrowne: - I wrapped the error message into a ConfigError and only print the message if the error from validation is a ConfigError
- The second argument to validate was unnecessary, so I removed it

packages/repo-doctor/repo-doctor.json
ptbrowne: I renamed the file to be an example. I don't know where we'll keep the default config though. I think it can be dealt with a later PR.

packages/repo-doctor/src/reporters/mattermost.js
ptbrowne: I'm not sure about that since the hook is supposed to be secret. I don't think we want this to be written in our config file.

packages/repo-doctor/src/index.js
ptbrowne: Mmh the constructor is the name of the class and it will not change with imports and exports. Classes are not subject to memoization. So I don't think we have an issue here.

ptbrowne:
> How to use it? Some rules have no configuration schema but rely on config options, and the reporters have no schemas, so setting one up requires digging into the code

I think the example config is sufficient right now.

> How to extend it? Mostly some docs about how to write rules (they now need to be a class with a run method that is a generator, I think that should be pointed out).

I've added a section in the README on rules implementation.

y-lohse:
> I think the example config is sufficient right now.

I have to say i'm not super convinced 😁 setting up the mattermost reporter without looking at the code for example seems hard to do. But the rest LGTM and I think it's better if we land this PR now and improve on it later 💪

ptbrowne:
I hadn't understood that you meant the configuration of the env variable, I've mentioned it in the README : https://github.com/cozy/cozy-libs/pull/1035/commits/d667d8b22aa377a7660f3c768593e74afa5a4ad6

PR #1039, opened 2020-07-17
Extract Wizard components from cozy-authentication

ptbrowne:
See the companion PR here where Wizard is added :

https://github.com/cozy/cozy-ui/pull/1505

Cozy-authentication will no longer add CSS utility classes from cozy-ui. 
This is important as before, it would import utility classes that could 
clash with utility classes from the cozy-ui in the app.


ptbrowne:
I always do a merge commit :) Why do you say it was all over the place ? I tried to make it understandable by not having too much change in each commit ?

PR #1040, opened 2020-07-20
Fix/procedure

Crash--:
I still had several issues : 
- We had several cozy-client version, so MobileRouter was not happy 
- Upgraded cozy-ui to get the latest Wizard Fixes
- Had to import the .js file instead of the jsx. I don't know why I've the jsx locally. It seems that it's not available on npm (https://unpkg.com/cozy-sharing@1.13.0/dist/withLocales.jsx) 
- Import UI's components from their respective path, not from the index


packages/playgrounds/src/sharing.jsx
```diff
@@ -4,7 +4,7 @@ import { Provider } from 'react-redux'
 import ReactDOM from 'react-dom'
 
 import SharingProvider, { ShareButton, ShareModal } from 'cozy-sharing'
-import withLocales from 'cozy-sharing/dist/withLocales'
```
ptbrowne, line 7: From a fresh repo, the withLocales.jsx is not there, I am not sure we need the fix.

```
ls node_modules/cozy-sharing/dist/ | grep withLocale
withLocales.js
```

packages/playgrounds/src/procedures.jsx
```diff
@@ -9,7 +9,11 @@ import 'cozy-ui/transpiled/react/stylesheet.css'
 import 'cozy-ui/dist/cozy-ui.min.css'
 // import Comp from 'cozy-procedures'
 import injectProcedureRoutes from '../../cozy-procedures/dist'
-import { PageLayout, PageFooter, PageContent } from 'cozy-ui/transpiled/react'
+import {
+  PageLayout,
+  PageFooter,
+  PageContent
+} from 'cozy-ui/transpiled/react/Page'
```
ptbrowne, line 16: I think we could implement a linting rule for this.

packages/playgrounds/src/sharing.jsx
Crash--: I tried to reproduce the issue but with no success. Certainly a previous debug 🤷‍♂️ . Reverted the commit

PR #1043, opened 2020-07-22
fix: Manually add window.opener in OAuth popup

doubleface:
This avoids a blank screen in OAuth steps on some version of safari on IPhone and desktop.
At least Safari 13 on IPhone XS (browserstack).


packages/cozy-harvest-lib/src/components/Popup.jsx
```diff
@@ -111,6 +111,11 @@ export class Popup extends PureComponent {
       title,
       `scrollbars=yes, width=${w}, height=${h}, top=${top}, left=${left}`
     )
+
+    // fixes no window.openen on some new version of safari
```
Crash--, line 115:
```suggestion
    // fixes no window.opener on some new version of safari
```

PR #1044, opened 2020-07-23
feat: Check if deps are at the same version

ptbrowne:
Across our repositories, we have dependency that should be at
the same version no matter what. It can be the case for cozy-client
dependencies, jest dependencies or babel dependencies.

Here, a rule is added that checks if a list of dependencies are
at the same version.

Result:

```
Repository: cozy/cozy-drive
  DepSameVersion: Different versions for cozy-client@13.14.2, cozy-stack-client@13.15.1, cozy-pouch-link@13.13.1
Repository: cozy/cozy-banks
  DepSameVersion: Different versions for cozy-client@13.15.1, cozy-stack-client@13.8.3, cozy-pouch-link@13.15.1
```

### DIFF
--- a/packages/repo-doctor/src/rules/depSameVersion.js
+++ b/packages/repo-doctor/src/rules/depSameVersion.js
@@ -22,7 +22,7 @@ class DepSameVersion {
         versions.slice(1).filter(x => x !== versions[0]).length > 0
       if (notSameVersion) {
         yield {
-          severity: 'warning',
+          severity: 'warn',
           type: 'DepSameVersion',
           message: `Different versions for ${repoDeps
             .map(r => `${r.name}@${r.version}`)


packages/repo-doctor/src/rules/index.js
```diff
@@ -1,8 +1,9 @@
 const { DepUpToDate, NoForbiddenDep } = require('./dependencies')
+const DepSameVersion = require('./depSameVersion')
```
Crash--, line 2: I think you forgot to add the file

ptbrowne: Yes.. fixed, ;)

ptbrowne:
You could have dependency that are not on latest, but still at the same version. Or dependencies that you do not want to check for latest version but that you want to check that they are at the same version.

To me, they do not check the same things.

ptbrowne:
I think we are still in exploratory phase regarding repo-doctor and it is easy to deactivate these rules via the config. So I think there is no problem adding this rule.

Crash-- (approved):

ptbrowne:
Actually, the rule was already merged. There was only a slight problem with the severity.

PR #1046, opened 2020-07-27
feat: Use VError to chain stack traces

ptbrowne:
`verror` is a library from joyent to facilitate error handling. It makes it
easy to chain errors (keeping stack traces along the way), or to deal with
MultiErrors. This makes it easier to understand where an error comes from.

Additionally, when handling the .json() decoding error, we would reuse the
body from the response to do .text(); this is not possible, we need to
clone the Response.


ptbrowne:
You had a publish failed error but the stack trace was coming from publisher.js instead of coming from the file that did the failing HTTP call, so it was hard to understand from where was coming the problem.

PR #1054, opened 2020-08-04
fix: Shared Element when the document has no path

Crash--:
We assumed that a document has a path or at least that
we can fetch it. It works well with an io.cozy.files but
not with an io.cozy.albums since an album doesn't have a
path.

Instead of displaying nothing if there is no path, let's just
not make the call to the methods that need the path to work.

- [x] I need to work on the latest test

### DIFF
--- a/packages/cozy-sharing/src/components/EditableSharingModal.jsx
+++ b/packages/cozy-sharing/src/components/EditableSharingModal.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 import PropTypes from 'prop-types'
 import flow from 'lodash/flow'
 import { queryConnect, withClient, Q } from 'cozy-client'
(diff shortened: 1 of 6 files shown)


packages/cozy-sharing/src/components/EditableSharingModal.jsx
```diff
@@ -58,8 +59,12 @@ const EditableSharingModal = ({
               link={getSharingLink(document.id)}
               permissions={getDocumentPermissions(document.id)}
               isOwner={isOwner(document.id)}
-              hasSharedParent={hasSharedParent(documentPath)}
-              hasSharedChild={hasSharedChild(documentPath)}
+              hasSharedParent={
+                documentPath ? hasSharedParent(documentPath) : false
```
y-lohse, line 63: I think `documentPath && hasSharedParent(documentPath)` would work as well and may be easier to read?

packages/cozy-sharing/src/components/useFetchDocumentPath.jsx
```diff
@@ -0,0 +1,18 @@
+import { useState, useEffect } from 'react'
+import { getFilesPaths } from '../helpers/files'
+
+export const useFetchDocumentPath = (client, document) => {
```
y-lohse, line 4: imo `useFetch` is a bit weird for a name, as the hook user you don't really need to know it's doing a fetch behind the scenes.

packages/cozy-sharing/src/components/useFetchDocumentPath.jsx
```diff
@@ -0,0 +1,18 @@
...
+  const [documentPath, setDocumentPath] = useState(
+    document.path ? document.path : null
+  )
+  useEffect(() => {
+    ;(async () => {
+      try {
+        const path = await getFilesPaths(client, document._type, [document])
```
y-lohse, line 11: Do we still need to do the fetch if `document.path` already exists?

packages/cozy-sharing/src/components/useFetchDocumentPath.jsx
Crash--: I think it's important to know that we're doing a fetch request when using this hook. It tells us if we want to pay the price of a request or not.

packages/cozy-sharing/src/components/useFetchDocumentPath.jsx
Crash--: I don't know there is no previous test telling us what is the expected behavior.

Since getFilesPath (which should be renamed to fetchFilesPaths...) makes a request, I think that the fetch can be useful to have the latest information.

PR #1055, opened 2020-08-05
WIP: fix: Add deprecated message on Contact methods

JF-Cozy:
Add deprecated message on Contact methods : they are copied in cozy-client https://github.com/cozy/cozy-client/pull/749

...waiting to validate cozy-client PR...

### DIFF
--- a/packages/cozy-doctypes/src/contacts/Contact.js
+++ b/packages/cozy-doctypes/src/contacts/Contact.js
@@ -9,6 +9,12 @@ const getPrimaryOrFirst = property => obj => {
   return obj[property].find(property => property.primary) || obj[property][0]
 }
 
+const logDeprecated = methodName =>
+  log(
+    'warn',
+    `${methodName} from cozy-doctypes contact is deprecated, use cozy-client/models/contacts/${methodName} instead`
+  )
+
 /**
  * Class representing the contact model.
  * @extends Document
@@ -31,6 +37,7 @@ class Contact extends Document {
    * @return {string} - the contact's initials
    */
   static getInitials(contact) {
+    logDeprecated('getInitials')
     if (typeof contact === 'string') {
       log(
         'warn',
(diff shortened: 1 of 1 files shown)


packages/cozy-doctypes/src/contacts/Contact.js
```diff
@@ -9,6 +9,8 @@ const getPrimaryOrFirst = property => obj => {
   return obj[property].find(property => property.primary) || obj[property][0]
 }
 
+const isDeprecated = () => log('warn', 'Deprecated, use contact model in cozy-client instead')
```
Crash--, line 12:
```suggestion
const isDeprecated = () => log('warn', 'cozy-doctypes contact's model is deprecated, use contact model in cozy-client instead')
```

Crash--: Also, since you're not deprecated all the methods, I'm wondering if we should not be more precise like : 

```
getInitials from cozy-doctypes contact is deprecated. Use cozy-client/models/contacts/getInitials instead
```

More the warning is explicit, more chance we have to take the migration path

JF-Cozy: Done, what do you think ?

packages/cozy-doctypes/src/contacts/Contact.js
```diff
@@ -9,6 +9,12 @@ const getPrimaryOrFirst = property => obj => {
   return obj[property].find(property => property.primary) || obj[property][0]
 }
 
+const isDeprecated = methodName =>
```
y-lohse, line 12: Super minor nit, but generally `isSomething` is a good way to identify that a variable is a boolean. I think for a function, `logDeprecated` or some name that implies and action makes more sense.

JF-Cozy: I had the same though but no better idea, thanks

PR #1061, opened 2020-08-19
Keep relationships field when matching accounts

ptbrowne:
When matching incoming accounts from a vendor to existing accounts,
we need to merge relationships to existing relationships.


Crash-- (approved):
Is it planned to move this code to cozy-client?

ptbrowne:
It is not planned yet as this code moves very rarely. It would be good though, but I do not think it is a priority.

PR #1064, opened 2020-08-25
No more logs/warnings during tests

ptbrowne:
Logs and warnings are distracting and can hide errors that should
be fixed. Here, all warnings and logs are removed from the
cozy-harvest-lib tests.

- Component needing to be wrapped in I18n
- cozy-logger needing to be deactivated
- minilog needing to be deactivated
- etc...


packages/cozy-harvest-lib/src/models/konnector/KonnectorJobWatcher.spec.js
```diff
@@ -1,3 +1,4 @@
+import CozyClient from 'cozy-client'
```
y-lohse, line 1: SHouldn't we use `createMockClient` in a place like this?

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
```diff
@@ -118,7 +126,9 @@ describe('TriggerManager', () => {
 
       it('should show the new account form', async () => {
         const { findByLabelText, findByTitle } = render(
-          <TriggerManager {...props} />
+          <I18n lang="en" dictRequire={() => enLocale}>
```
y-lohse, line 129: Nit: we could use a setup function to avoid repeating the I18n wrapper :)

y-lohse (approved):
Cool, LGTM!

packages/cozy-harvest-lib/test/components/TriggerManager.spec.js
ptbrowne: Yes, good idea, thanks.

packages/cozy-harvest-lib/src/models/konnector/KonnectorJobWatcher.spec.js
ptbrowne: We could but this is not necessary here. The mock client as of now is mostly used to work with query data (so as to have query data already in the store).

For coherence I think yes, it would be better. Not doing this now though.

PR #1066, opened 2020-08-27
feat(Harvest): Add website link in connector data

doubleface:
This will make it easier to navigate from cozy-home

Screenshot_2020-08-27 Cozy Home

### DIFF
--- a/packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
+++ b/packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
@@ -1,0 +1,48 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import Card from 'cozy-ui/transpiled/react/Card'
+import { SubTitle, Caption } from 'cozy-ui/transpiled/react/Text'
+import { ButtonLink } from 'cozy-ui/transpiled/react/Button'
+import { Media, Img, Bd } from 'cozy-ui/transpiled/react/Media'
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import palette from 'cozy-ui/transpiled/react/palette'
+
+import { translate } from 'cozy-ui/transpiled/react/I18n'
+
+const WebsiteLinkCard = ({ link, t }) => {
+  const imgStyle = { paddingRight: '1rem' }
+  const linkStyle = { textTransform: 'lowercase', margin: 0 }
+  const hrStyle = { color: palette['silver'] }
+  const label = new URL(link).host
+
+  return (
+    <Card>
+      <SubTitle className="u-mb-1">{t('card.websiteLink.title')}</SubTitle>
+      <hr style={hrStyle} />
+      <Media align="top">
+        <Img style={imgStyle}>
+          <Icon icon="globe" color={palette['coolGrey']} />
+        </Img>
+        <Bd>
+          <ButtonLink
+            subtle
+            target="_blank"
+            href={link}
+            label={label}
+            theme="text"
+            className="u-dodgerBlue"
+            style={linkStyle}
+          />
+          <Caption>{t('card.websiteLink.description')}</Caption>
+        </Bd>
+      </Media>
+    </Card>
+  )
+}
+
+WebsiteLinkCard.propTypes = {
+  link: PropTypes.string.isRequired
+}
+
+export default translate()(WebsiteLinkCard)
--- a/packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
+++ b/packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
@@ -11,17 +11,15 @@ import palette from 'cozy-ui/transpiled/react/palette'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
 
 const WebsiteLinkCard = ({ link, t }) => {
-  const imgStyle = { paddingRight: '1rem' }
-  const linkStyle = { textTransform: 'lowercase', margin: 0 }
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
```diff
@@ -0,0 +1,48 @@
...
+  const linkStyle = { textTransform: 'lowercase', margin: 0 }
+  const hrStyle = { color: palette['silver'] }
+  const label = new URL(link).host
+
+  return (
+    <Card>
+      <SubTitle className="u-mb-1">{t('card.websiteLink.title')}</SubTitle>
+      <hr style={hrStyle} />
+      <Media align="top">
+        <Img style={imgStyle}>
```
Crash--, line 24:
```suggestion
        <Img className={'u-pr-1"}>
```

packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
```diff
@@ -0,0 +1,48 @@
...
+const WebsiteLinkCard = ({ link, t }) => {
+  const imgStyle = { paddingRight: '1rem' }
+  const linkStyle = { textTransform: 'lowercase', margin: 0 }
+  const hrStyle = { color: palette['silver'] }
+  const label = new URL(link).host
+
+  return (
+    <Card>
+      <SubTitle className="u-mb-1">{t('card.websiteLink.title')}</SubTitle>
+      <hr style={hrStyle} />
```
Crash--, line 22:
```suggestion
      <hr className={"u-silver"} />
```

packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
```diff
@@ -0,0 +1,48 @@
...
+          <Icon icon="globe" color={palette['coolGrey']} />
+        </Img>
+        <Bd>
+          <ButtonLink
+            subtle
+            target="_blank"
+            href={link}
+            label={label}
+            theme="text"
+            className="u-dodgerBlue"
```
Crash--, line 34:
```suggestion
            className="u-dodgerBlue u-m-0"
```

packages/cozy-harvest-lib/src/components/cards/__snapshots__/WebsiteLinkCard.spec.jsx.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`WebsiteLinkCard should render 1`] = `
+<withI18n(WebsiteLinkCard)
+  link="https://fc.assure.ameli.fr/FRCO-app/login"
+/>
+`;
```
Crash--, line 7: Not sure this is the withI18n we want to snapshot. Not very useful I think. Snapshoting WebsiteLinkCard will be better

Crash-- (commented):
Cool to see you working on UI :D. 

You should avoid using inline style. UI comes with a lot of utility classes (https://docs.cozy.io/cozy-ui/styleguide/section-utilities.html) and we should use it. 

If there is no utility class for want you want (for instance textTransform lowercase...) you should create a new class

packages/cozy-harvest-lib/src/components/cards/__snapshots__/WebsiteLinkCard.spec.jsx.snap
doubleface: This is a copy from some other harvest unit test. It allow me to force a locale. It seems that passing t as a property in the unit test is not enough.

Crash--: Let's do that in an other way. 

Remove the translate() HOC and use our new useI18n() hook. (If we can use our new method in our new component, it's better ;)) 

Your component will be: 
```
const WebsiteLinkCard = ({ link}) => {
  const linkStyle = { textTransform: 'lowercase' }
  const label = new URL(link).host
  const { t } = useI18n()
  return (
    <Card>
      <SubTitle className="u-mb-1">{t('card.websiteLink.title')}</SubTitle>
      <hr className="u-silver" />
      <Media align="top">
        <Img className="u-pr-1">
          <Icon icon="globe" color={palette['coolGrey']} />
        </Img>
        <Bd>
          <ButtonLink
            subtle
            target="_blank"
            href={link}
            label={label}
            theme="text"
            className="u-dodgerBlue u-m-0"
            style={linkStyle}
          />
          <Caption>{t('card.websiteLink.description')}</Caption>
        </Bd>
      </Media>
    </Card>
  )
}
```

Then I will only test a few things like : 
```
const wrapper = mount(
      <I18n dictRequire={() => enLocale} lang="en">
        <WebsiteLinkCard link="https://fc.assure.ameli.fr/FRCO-app/login" />
      </I18n>
    )
    const component = wrapper.find(WebsiteLinkCard)
    expect(component.find('ButtonLink').props().href).toEqual(
      'https://fc.assure.ameli.fr/FRCO-app/login'
    )
    expect(component.find('ButtonLink').props().label).toEqual(
      'fc.assure.ameli.fr'
    )
```

to be sure that we have the right href & label.


Having a snapshot with `<withI18n(WebsiteLinkCard)
  link="https://fc.assure.ameli.fr/FRCO-app/login"
/>` is not wanted.

If you really want to have the full snapshoted component you can either mount() and then component.debug().match... or use a serializer like https://github.com/adriantoine/enzyme-to-json or something similar

PR #1067, opened 2020-08-27
Improvements

ptbrowne:
Various improvements to the configuration tab of the Home, seen
while developing on Banks.


packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/EditContract.jsx
```diff
@@ -190,6 +191,9 @@ const EditContract = props => {
         Alerter.success(t('contractForm.success'))
       },
       onError: err => {
+        if (onError) {
+          onError()
```
Crash--, line 96: no need to have `onError(err)` ?

packages/cozy-harvest-lib/src/components/AccountSelectBox/AccountSelectBox.jsx
```diff
@@ -12,8 +12,12 @@ const AccountSelectBox = ({
   accountsList,
   selectedAccount,
   onChange,
-  onCreate
+  onCreate,
+  loading
 }) => {
+  if (loading) {
+    return <div style={{ height: 20.8 }}>&nbsp;</div>
```
Crash--, line 19: tiny is about 2rem, no? https://github.com/cozy/cozy-ui/blob/master/react/SelectBox/SelectBox.jsx#L14 Weird to see 20.8px

packages/cozy-harvest-lib/src/components/KonnectorAccounts.jsx
```diff
@@ -92,7 +88,8 @@ export class KonnectorAccounts extends React.Component {
   }
 
   async fetchAccounts() {
-    const triggers = get(this.props, 'konnector.triggers.data', [])
+    const triggers =
+      this.props.triggers || get(this.props, 'konnector.triggers.data', [])
```
Crash--, line 92: this change should not be in the refactor commit. It can changes a lot the behavior 🤔

packages/cozy-harvest-lib/src/components/KonnectorIcon.jsx
```diff
@@ -6,10 +6,10 @@ import AppIcon from 'cozy-ui/transpiled/react/AppIcon'
 // TODO move this to cozy-ui
 class KonnectorIcon extends PureComponent {
   fetchIcon() {
-    const { konnector, client } = this.props
+    const { konnector, konnectorSlug, client } = this.props
     return client.getStackClient().getIconURL({
       type: 'konnector',
-      slug: konnector.slug
+      slug: konnectorSlug || konnector.slug
```
Crash--, line 12: nit: I'm not a big fan of having a component named `KonnectorXXX` which can takes something that is not a `konnector` object. 

Do we get somewhere the Slug without having the full object?

packages/cozy-harvest-lib/src/components/KonnectorAccounts.jsx
ptbrowne: Why, the triggers prop is new here, I do not see the risk here 🤔

packages/cozy-harvest-lib/src/components/AccountSelectBox/AccountSelectBox.jsx
ptbrowne: I've checked the height of the component and the height was 20.8 exactly. It's true that it is weird :/

packages/cozy-harvest-lib/src/components/KonnectorIcon.jsx
ptbrowne: Yes, the refactor here is to be able for the caller not to have a full konnector object. For example if you have io.cozy.accounts and want to show the icon of their konnector, you do not want to create a fake `{ slug: myAccount.account_type}` "konnector" object.

packages/cozy-harvest-lib/src/components/KonnectorAccounts.jsx
Crash--: Didn't check exactly where {...props} is used, but : 
```
cozy-harvest-lib git grep -Ri "{...props}" . | wc -l
      32
```

Here is the risk I see. 

but after a quick check, `KonnectorAccounts` seems to be only used one time with only one prop, so it's fine (if we get this change outside of a refactor one)

ptbrowne: Ouch, I did not see the problem with prop drilling, thanks for reminding me not to use this technique :D 

Anyway, you're right for this not to be a refactor, I think I included it by mistake. I've splitted in two the commit, which was helpful also to put a nice description in the commit. See [`8f637cd` (#1067)](https://github.com/cozy/cozy-libs/pull/1067/commits/8f637cde9da4bd71f888cfdef49c607d007ee183)

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/EditContract.jsx
ptbrowne: Done in 3728ba0e.

Crash-- (approved):

PR #1069, opened 2020-08-31
fix(Harvest): Change hr separator height to 1px

doubleface:
To be more like mockup

### DIFF
--- a/packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
+++ b/packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 
 import Card from 'cozy-ui/transpiled/react/Card'
 import { SubTitle, Caption } from 'cozy-ui/transpiled/react/Text'
+import Divider from 'cozy-ui/transpiled/react/MuiCozyTheme/Divider'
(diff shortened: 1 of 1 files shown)


packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
```diff
@@ -11,13 +11,18 @@ import { useI18n } from 'cozy-ui/transpiled/react/I18n'
...
+    backgroundColor: palette['silver']
+  }
   const label = new URL(link).host
   const { t } = useI18n()
 
   return (
     <Card>
       <SubTitle className="u-mb-1">{t('card.websiteLink.title')}</SubTitle>
-      <hr className="u-silver" />
+      <hr style={hrStyle} />
```
ptbrowne, line 25: You can use `react/MuiCozyTheme/Divider/index.jsx` for this kind of Divider.

For cases like that, it is not a good practice to use adhoc styling. If there is a divider here, there is a good likelihood that there will be other divider in other parts of the UI.

doubleface: I could not find this divider in https://docs.cozy.io/cozy-ui/react/#/Material-UI and when I try it, I get an error.

ptbrowne (approved):

PR #1070, opened 2020-08-31
Improvements (again)

ptbrowne:
Needed to integrate Harvest in Banks.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/Contracts.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/Contracts.jsx
@@ -1,4 +1,5 @@
 import React, { useState } from 'react'
+import PropTypes from 'prop-types'
 import startCase from 'lodash/startCase'
 import compose from 'lodash/flowRight'
 
@@ -39,6 +40,7 @@ const ContractItem = ({ contract }) => {
   const [showingEditModal, setShowingEditModal] = useState(false)
   const getPrimaryText =
     getPrimaryTextPerDoctype[contract._type] || getPrimaryTextDefault
+  const { t } = useI18n()
   return (
     <>
       <ListItem
@@ -52,6 +54,7 @@ const ContractItem = ({ contract }) => {
         </ListItemIcon>
         <ListItemText
           primaryText={startCase(getPrimaryText(contract).toLowerCase())}
+          secondaryText={contract._deleted ? t('contracts.deleted') : null}
         />
         <ListItemSecondaryAction>
           <Icon icon="right" className="u-coolGrey u-mr-1" />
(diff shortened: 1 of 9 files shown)


packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/Contracts.jsx
```diff
@@ -52,6 +53,7 @@ const ContractItem = ({ contract }) => {
         </ListItemIcon>
         <ListItemText
           primaryText={startCase(getPrimaryText(contract).toLowerCase())}
+          secondaryText={contract._deleted ? 'Supprimé' : null}
```
Crash--, line 56: translate

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/Contracts.jsx
```diff
@@ -77,22 +85,32 @@ const customHeaderPerDoctype = {
 }
 
 const DumbContracts = ({ contracts, doctype }) => {
+  contracts = contracts.data ? contracts.data : contracts
```
Crash--, line 88: We don't have an eslint rule telling us that we should not modify the params (even after a destructuring assignment)?

Crash-- (approved):

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/Contracts.jsx
ptbrowne: We are not modifying the params here, we are modifying the reference.

ptbrowne: And for the eslint rule, I searched and found : https://eslint.org/docs/rules/no-param-reassign , it's interesting I did not know that it would change the arguments. After a destructuring, I think it's safe since it boils down to assigning a new variable. I can be more explicit and assign a new variable contractData though.

ptbrowne: I do not think the rule is activated per default

ptbrowne: Done in v

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/Contracts.jsx
ptbrowne: Thanks, fixed.

PR #1074, opened 2020-09-01
feat: Activate the no-param-reassign rule

### LINKED ISSUE #1073: Activate the eslint.org/docs/rules/no-param-reassign rule
Modifying functions params can be unsafe as it modifies
the arguments array.
This rule is not activated by default on eslint default
config.
We could activate this rule in eslint-config-cozy-app.

https://eslint.org/docs/rules/no-param-reassign

ptbrowne:
See the doc here: https://eslint.org/docs/rules/no-param-reassign
Fix #1073

### DIFF
--- a/packages/eslint-config-cozy-app/basics.js
+++ b/packages/eslint-config-cozy-app/basics.js
@@ -18,6 +18,7 @@ module.exports = {
         singleQuote: true,
         semi: false
       }
-    ]
+    ],
+    'no-param-reassign': 'warn'
   }
 }


ptbrowne:
As the rule can detect new problems, it is likely to break existing linting. Do you think I should mark the commit as BREAKING CHANGE ?

y-lohse:
cc @Crash-- @JF-Cozy @doubleface @LucsT since this will presumably require changes in a few places. But the rule makes sense and the migration shouldn't be very long, so 👍 for me.

JF-Cozy:
@y-lohse I agree

Crash--:
Yes this is a breaking change, and we'll also have to create a new major version of cozy-script because of this rule. 

Should we first add a warning before rising an error and creating breaking change everywhere? (not only in our internal apps but also external ones).

ptbrowne:
OK for the warning.

y-lohse (approved):

doubleface:
I am ok with this rule with a warning

JF-Cozy (approved):

PR #1076, opened 2020-09-01
fix(Harvest): Add margin on top of the website link

doubleface:
Screenshot_2020-09-01 Cozy Home


Crash-- (approved):
Next time you can consider the use of `Stack` component

PR #1077, opened 2020-09-01
fix: Ability to copy links

y-lohse:
By default, MUI Dialog keeps the focus on Dialogs for accessibility reasons. Most of the time that's good, but here our copy to clipboard lib needs control over the focus to do it's job.

### DIFF
--- a/packages/cozy-sharing/src/components/ShareByLink.jsx
+++ b/packages/cozy-sharing/src/components/ShareByLink.jsx
@@ -52,12 +52,13 @@ class ShareByLink extends React.Component {
     }
   }
 
-  copyLinkToClipboard = () => {
+  copyLinkToClipboard = ({ isAutomaticCopy }) => {
     if (copy(this.props.link))
       Alerter.success(
         this.props.t(`${this.props.documentType}.share.shareByLink.copied`)
       )
-    else
+    else if (!isAutomaticCopy)
+      // In case of automatic copy, the browser can block the copy request. This is not shown to the user since it is expected and can be circumvented by clicking directly on the copy link
       Alerter.error(
         this.props.t(`${this.props.documentType}.share.shareByLink.failed`)
       )
@@ -122,7 +123,7 @@ class ShareByLink extends React.Component {
               className={styles['aligned-dropdown-button']}
               onClick={async () => {
                 await this.createShareLink()
-                this.copyLinkToClipboard()
+                this.copyLinkToClipboard({ isAutomaticCopy: true })
               }}
               busy={loading}
               label={t(`${documentType}.share.shareByLink.create`)}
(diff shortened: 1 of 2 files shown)


ptbrowne:
I have no problem with the solution, but it'd be good to understand more precisely what's going on.

Does this mean everytime we have a MUI Dialog we cannot have a a copy text button, unless we had this property ?

https://github.com/sindresorhus/copy-text-to-clipboard/blob/master/index.js

y-lohse:
> Does this mean everytime we have a MUI Dialog we cannot have a a copy text button, unless we add this property ?

At least while using this lib, yes. I assume that if we use the clipboard API straght up we don't need the prop, since we wouldn't care about the focus.

packages/cozy-sharing/src/components/ShareByLink.jsx
```diff
@@ -52,12 +52,13 @@ class ShareByLink extends React.Component {
...
 
-  copyLinkToClipboard = () => {
+  copyLinkToClipboard = ({ isAutomaticCopy }) => {
     if (copy(this.props.link))
       Alerter.success(
         this.props.t(`${this.props.documentType}.share.shareByLink.copied`)
       )
-    else
+    else if (!isAutomaticCopy)
+      // Copy after a XHR query fails in safari
```
ptbrowne, line 61: I think the behavior of Safari is totally normal, we should not copy after the XHR, it is a security problem and I would not be surprised if browsers in the future prevent us from doing that. To prevent that we could modify the UI so that the copy to clipboard button is added after the link has been created.

y-lohse: > To prevent that we could modify the UI so that the copy to clipboard button is added after the link has been created.

It is :). We try to copy the link to the clipboard right away for convenience, but we do indeed add a copy to clipboard button in the UI that works as expected.

ptbrowne: Ah ok, good. In this case I think the comment would be clearer if it read "In case of automatic copy, the browser can block the copy request. This is not shown to the user since it is expected and can be circumvented by clicking directly on the copy link".

ptbrowne: From an API point of view, I wonder if it would not be better to have ignoreErrors: true instead of isAutomaticCopy. If you read the call point, you can wonder why is there isAutomaticCopy there.

y-lohse: I updated the comment 👍 .

> If you read the call point, you can wonder why is there isAutomaticCopy there.

I wondered about that as well, but then it makes it less clear why the error is ignored in one case and not the other. I think it's better to make it explicit why the error is suppressed rather than why the function is called with an extra arg.

ptbrowne: I see, thanks for the explanation.

Crash-- (approved):

PR #1078, opened 2020-09-01
fix(Harvest): deliver an harvest 2.2.0 with only WebsiteLink component addition

doubleface:
This branch takes its root from the 2.2.0 version of harvest (used
in last 1.30.5 version of cozy-home) and only adds cherry-picked
commits related to WebSiteLink component.

This will allow to deliver the next 1.30.6 version of cozy-home with
the WebSiteLink component and without more impactful and still to
be validated changes.

I also added a fix to not throw a error when the vendor_link is not
a valid url.

Do not throw error on invalid url in manifest

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
@@ -11,6 +11,7 @@ import KonnectorUpdateInfos from '../../../components/infos/KonnectorUpdateInfos
 import LaunchTriggerCard from '../../../components/cards/LaunchTriggerCard'
 import KonnectorMaintenance from '../../../components/Maintenance'
 import AppLinkCard from '../../../components/cards/AppLinkCard'
+import WebsiteLinkCard from '../../../components/cards/WebsiteLinkCard'
 import TriggerErrorInfo from '../../../components/infos/TriggerErrorInfo'
 import useMaintenanceStatus from '../../../components/hooks/useMaintenanceStatus'
 import getRelatedAppsSlugs from '../../../models/getRelatedAppsSlugs'
@@ -70,6 +71,9 @@ export const DataTab = ({ konnector, trigger, client, flow }) => {
         {appLinks.map(({ slug, ...otherProps }) => (
           <AppLinkCard key={slug} slug={slug} {...otherProps} />
         ))}
+        {konnector.vendor_link && (
+          <WebsiteLinkCard link={konnector.vendor_link} />
+        )}
       </Stack>
     </ModalContent>
   )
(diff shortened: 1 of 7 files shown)


packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
```diff
@@ -70,6 +71,9 @@ export const DataTab = ({ konnector, trigger, client, flow }) => {
         {appLinks.map(({ slug, ...otherProps }) => (
           <AppLinkCard key={slug} slug={slug} {...otherProps} />
         ))}
+        {konnector.vendor_link && (
```
ptbrowne, line 74: Does this handle the trainline case where vendor_link can be an array ?

packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
```diff
@@ -0,0 +1,56 @@
...
+import { useI18n } from 'cozy-ui/transpiled/react/I18n'
+
+const WebsiteLinkCard = ({ link }) => {
+  const { t } = useI18n()
+  let url = null
+  try {
+    url = new URL(link)
+  } catch (err) {
+    logger('warn', err.message)
+    return ''
```
ptbrowne, line 21: Nit: `return null` is better here. No need to fix it here right away since we want a quick publication. But it'd be cool to fix it on master.

packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.spec.jsx
```diff
@@ -0,0 +1,32 @@
...
+  it('should ignore invalid urls', () => {
+    const link = 'bad url'
+    const wrapper = mount(
+      <I18n dictRequire={() => enLocale} lang="en">
+        <WebsiteLinkCard link={link} />)
+      </I18n>
+    )
+    const component = wrapper.find(WebsiteLinkCard)
+    expect(component.exists('ButtonLink')).toEqual(false)
+  })
```
ptbrowne, line 31: We should have a testcase with the trainline case.

ptbrowne (approved):

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
doubleface: Actually, the trainline case is not an array, it is just an invalid url (without, https://)

Crash-- (approved):

ptbrowne:
Mmh je ne comprends pas pourquoi cette branche est mergée ? L'idée n'était pas de publier depuis la branche sans merger ? C'était pour intégrer le fix sur les URLs invalides ?

PR #1079, opened 2020-09-01
fix: Handle case when all contracts have been deleted

ptbrowne:
If all contracts have been deleted, we still need to show
the icon of the konnector.


packages/cozy-harvest-lib/src/components/DisconnectedAccountModal.jsx
```diff
@@ -22,19 +25,28 @@ const createDummyKonnectorFromAccount = account => {
  * has been deleted
  */
 const DisconnectedModal = ({ accounts, onClose }) => {
+  const { t } = useI18n()
+  // We keep the konnector in a ref so that when we remove all accounts,
+  // we still have a konnector to show the icon
+  const konnectorRef = useRef()
+  if (!konnectorRef.current) {
+    konnectorRef.current = createDummyKonnectorFromAccount(accounts[0])
+  }
```
Crash--, line 34: clever

Crash-- (approved):

PR #1081, opened 2020-09-02
Remove lint warnings

ptbrowne:
Started to remove warnings for reassignment of function
params.


packages/cozy-doctypes/src/utils.js
```diff
@@ -8,8 +8,8 @@ const PromisePool = require('es6-promise-pool')
  * @param  {Number}   concurrency  How many promise can be in flight at the same time
  * @return {Promise}               Resolved with the results of the promise, not necessary in order
  */
-const parallelMap = (iterable, fn, concurrency) => {
-  concurrency = concurrency || 30
+const parallelMap = (iterable, fn, concurrencyArg) => {
+  const concurrency = concurrencyArg || 30
```
y-lohse, line 12: I think for this type of simple args we could use default arg values instead of a re-assignment.

y-lohse (approved):
👍

PR #1082, opened 2020-09-03
Add aria-label + test on DisconnectedAccountModal

ptbrowne:
Without aria-label we have a warning and this is problematic
for apps that do no want warnings in their console.

Here I also make it so that console.warn throws during tests.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/DisconnectedAccountModal.spec.jsx
+++ b/packages/cozy-harvest-lib/src/components/DisconnectedAccountModal.spec.jsx
@@ -1,0 +1,29 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+
+import { CozyProvider } from 'cozy-client'
+import { BreakpointsProvider } from 'cozy-ui/transpiled/react/hooks/useBreakpoints'
+import I18n from 'cozy-ui/transpiled/react/I18n'
+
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/DisconnectedAccountModal.spec.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+import DisconnectedAccountModal from './DisconnectedAccountModal'
+import bankAccounts from './KonnectorConfiguration/ConfigurationTab/bank-account-fixture.json'
+import en from '../locales/en.json'
+
+describe('DisconnectedAccountModal', () => {
+  const setup = () => {
+    const mockClient = {}
+    const root = render(
+      <CozyProvider client={mockClient}>
```
y-lohse, line 16: Nit: we could extract this into an AppLike component to help future testing efforts.

y-lohse (approved):

packages/cozy-harvest-lib/src/components/DisconnectedAccountModal.spec.jsx
ptbrowne: Yes please ;)

PR #1088, opened 2020-09-07
fix: Use primaryColor instead of dodgerBlue

ptbrowne:
Use primaryColor instead of dodgerBlue for link to site.


Crash-- (approved):
my bad. Thanks

PR #1090, opened 2020-09-08
[Sharing] Avatars

y-lohse:
A few different things happen in this PR:

- Remove the custom Avatar component in favour of the one from cozy-ui
- The next versions of the stack will have the ability to serve an image avatar for sharing members, so we use primarily that image, with a fallback for normal avatars
- When sharing a file from cozy to cozy, the stack creates a shortcut file on the recipients cozy that points to the sharing's preview page. This PR helps us support this kind of "preshared" file
- A new top level component, `SharingOwnerAvatar` that we will use in Drive in an upcoming PR.


packages/cozy-sharing/src/components/AvatarPlusX.jsx
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from 'prop-types'
+import { SharingTooltip, TooltipRecipientList } from './Tooltip'
+import Avatar from 'cozy-ui/transpiled/react/Avatar'
+
+const AvatarPlusX = ({ size, extraRecipients = [] }) => (
+  <div data-tip data-for="extra-recipients-avatar">
+    <Avatar
+      size={size}
+      text={`+${Math.min(extraRecipients.length, 99)}`}
+      style={{ backgroundColor: 'var(--silver)', color: 'var(--charcoalGrey)' }}
```
Crash--, line 11: we should not use --silver / --charcoalGrey, can't we use primary and so on?

packages/cozy-sharing/src/components/Recipient.jsx
```diff
@@ -43,7 +45,11 @@ export const RecipientsAvatars = ({
       )}
       onClick={onClick}
     >
-      {link && <AvatarLink size={size} />}
+      {link && (
+        <Circle size={size} backgroundColor="var(--silver)">
```
Crash--, line 49: same here

packages/cozy-sharing/src/state.js
```diff
@@ -370,11 +370,18 @@ const getSharedDocIds = doc =>
...
 
-export const getSharingDocIds = sharing =>
-  sharing.attributes.rules
+export const getSharingDocIds = sharing => {
+  const docs = sharing.attributes.rules
     .map(r => r.values)
     .reduce((acc, val) => acc.concat(val), [])
 
+  if (sharing.attributes.shortcut_id) {
+    docs.push(sharing.attributes.shortcut_id)
```
Crash--, line 380: Does that mean that the created shortcut is not shared itself?

packages/cozy-sharing/src/components/Recipient.jsx
```diff
@@ -61,18 +61,29 @@ export const RecipientsAvatars = ({
...
+  const client = useClient()
+  return (
+    <Avatar
+      image={`${client.options.uri}${recipient.avatarPath}`}
+      text={Contact.getInitials(recipient)}
+      textId={Contact.getDisplayName(recipient)}
+      {...rest}
+    />
+  )
+}
```
Crash--, line 85: What does that mean if we're in an offline mobile app? Should we have a fallback?

y-lohse: The  `Avatar` component handles the fallback automatically, since we still pass the text etc.

packages/cozy-sharing/src/state.js
y-lohse: That's correct.

packages/cozy-sharing/src/components/Recipient.jsx
Crash--: Nop. 

https://github.com/cozy/cozy-ui/blob/master/react/Avatar/index.jsx#L53-L56

We pass an image, so there is no way Avatar will use the text?

packages/cozy-sharing/src/state.js
Crash--: OK so we create a fake shared docs. So if we go to the sharing views, we'll see it? 
Since the doc will be "shared", what happens if we open the Sharing Modal on it? Are we displaying "Create a link" or remove the link? And what happen if we click on that action? 
To be more precise, getSharingDocs is used to call `createSharingInStore` and then we add this document in the sharing store. I'm not sure of what happen next 

I'm not a big fan of creating fake shared document since that action can have a few consequences (error when trying to remove the sharing or so )

packages/cozy-sharing/src/components/Recipient.jsx
y-lohse: Shit you're right, I mistested that. The actual result isn't too bad:

Capture d’écran 2020-09-08 à 14 38 17

What do you think of testing with `navigator.onLine` whether we pass the image or not?

Crash--: > What do you think of testing with navigator.onLine whether we pass the image or not?

I don't know... Since we should have a cache one day... I think I'm wondering if the actual result is not enough

packages/cozy-sharing/src/state.js
y-lohse: > So if we go to the sharing views, we'll see it?

Yes, while the shortcut is there we're basically treating this file as if it was the real file and not the preview.

> what happens if we open the Sharing Modal on it?

We show the sharing infos for the actual sharing. Not sure if we have the full list of recipients before accepting the sharing, but at least the sharing owner and current user.

> Are we displaying "Create a link" or remove the link? And what happen if we click on that action?

Sharing by link is a totally separate mechanic. So… it's like a normal file, you can share the shortcut by link if you want.

> and then we add this document in the sharing store. I'm not sure of what happen next

Not much from what I can tell. It's in the `byDocId` slice of the store, which gives us an easy access to the related `io.cozy.sharings`.

______

We could not create "fake" shared docs, since in the short io.cozy.files metadata, we have the id of the sharing. But this would require handling this as a special case in drive in multiple places, as well a major changes in the cozy-sharings component so that they accept a `bySharingId` prop on top of `byDocId`.

packages/cozy-sharing/src/components/AvatarPlusX.jsx
y-lohse: We don't have semantic color names for this use case (it's for example unrelated to the primaryTextColor) — imo it's a combination specific to these cozy-sharing recipients. 

What do you think of adding semantic color names for this in cozy-sharing (eg. `--genericRecipientBackground` and `--genericRecipientText`)? That way the names are semantic and we can override them with a theme.

Crash--: I thinks it's better yeah.

packages/cozy-sharing/src/components/recipient.styl
```diff
@@ -2,6 +2,10 @@
 @require 'settings/breakpoints.styl'
 @require 'settings/palette.styl'
 
+:root
+  --genericRecipientBackground: var(--silver)
+  --genericRecipientColor: var(--charcoalGrey)
```
Crash--, line 7: Not still very convinced by that. but since its better than the first version, lets use that

Crash-- (approved):

PR #1093, opened 2020-09-11
chore(Harvest): Update bi-auth to 0.0.8

doubleface:
To get new banking connectors ids into master

A new 2.2.2 version of harvest is already available with this new version of cozy-bi-auth to avoid unneeded feature collisions in home.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4283,6 +4283,14 @@ buffer@^4.3.0:
     ieee754 "^1.1.4"
     isarray "^1.0.0"
 
+buffer@^5.5.0:
+  version "5.6.0"
+  resolved "https://registry.yarnpkg.com/buffer/-/buffer-5.6.0.tgz#a31749dc7d81d84db08abf937b6b8c4033f62786"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -14150,6 +14168,11 @@ react-transition-group@^2.2.1:
     prop-types "^15.6.2"
     react-lifecycles-compat "^3.0.4"
 
+react-zlib-js@^1.0.4:
+  version "1.0.5"
+  resolved "https://registry.yarnpkg.com/react-zlib-js/-/react-zlib-js-1.0.5.tgz#7bb433e1a4ae53a8e6f361b3d36166baf5bbc60f"
+  integrity sha512-TLcPdmqhIl+ylwOwlfm1WUuI7NVvhAv3L74d1AabhjyaAbmLOROTA/Q4EQ/UMCFCOjIkVim9fT3UZOQSFk/mlA==
+
```
ptbrowne, line 14175: Mmh interesting, why is this added by an update of the bi-auth-package ?

doubleface: This comes from an update of node-jose in bi-auth (even if I do not see any reason to add react in the name)

ptbrowne (approved):

PR #1101, opened 2020-09-15
fix: Flaky builds due to SyntaxError

ptbrowne:
A default options object was modified. This could cause race problems
when building a service and an app, the service configuring the
default options to have { node: true } in the babel-preset-cozy-app
default options, preventing the code from the app to be correctly
parsed.

Example of error:

"Support for the experimental syntax 'jsx' isn't currently enabled"

This error would disappear on subsequent builds, because of caching
(not sure why caching helps though). It was particularly problematic on
the CI where there's no cache and where builds must work the 1st time.

Fixes https://github.com/cozy/create-cozy-app/issues/1342


JF-Cozy (approved):

Crash-- (commented):
NICE <3

PR #1104, opened 2020-09-15
Add softDeleteOrRestoreAccounts service

paultranvan:
This service is useful to deal with soft delete and restore from
bitwarden, used for a trash feature. See [the stack PR](https://github.com/cozy/cozy-stack/pull/2633).

When a cipher is soft deleted, a `deletedDate` field is added to the
cipher. Then, the service removes the encrypted credentials from the
referenced account and delete the associated trigger.

When the cipher is restored, the `deletedDate` is removed: the service
recreates the account's trigger. The account's credentials are then
restored through the `updateAccountsFromCipher` service.

### DIFF
--- a/packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
+++ b/packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
@@ -32,7 +32,7 @@ const makeDecrypt = (vaultClient, orgKey) => encryptedVal =>
 
  * When a cipher is soft deleted, a deletedDate field is added to the
  * cipher. Then, the service removes the encrypted credentials from the
- * referenced account and delete the associated trigger.
+ * referenced account and deletes the associated trigger.
 
  * When the cipher is restored, the deletedDate is removed: the service
  * recreates the account's trigger. The account's credentials are then
--- a/packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
+++ b/packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
@@ -1,0 +1,85 @@
+import logger from './logger'
+import {
+  decryptString,
+  getOrganizationKey,
+  fetchAccountsForCipherId,
+  fetchKonnectorFromAccount,
+  fetchTriggersFromAccount,
+  updateAccountsAuth
+} from './utils'
+import get from 'lodash/get'
+import { ensureTrigger } from '../connections/triggers'
+import { translate as t } from 'cozy-ui/transpiled/react/I18n'
+
+const makeDecrypt = (vaultClient, orgKey) => encryptedVal =>
+  decryptString(encryptedVal, vaultClient, orgKey)
+
+const softDeleteOrRestoreAccounts = async (
+  cozyClient,
+  vaultClient,
+  cipherDoc
+) => {
+  const cipherId = get(cipherDoc, '_id')
+
+  logger.debug(`Fetching accounts for cipher ${cipherId}...`)
+  const accounts = await fetchAccountsForCipherId(cozyClient, cipherId)
+  if (accounts.data.length < 1) {
+    logger.debug('No account found for this cipher.')
+    return
+  }
+  const orgKey = await getOrganizationKey(cozyClient, vaultClient)
+  const decrypt = makeDecrypt(vaultClient, orgKey)
+
+  const encryptedPassword = get(cipherDoc, 'login.password')
+  const encryptedUsername = get(cipherDoc, 'login.username')
+
+  const encryptedFields = get(cipherDoc, 'fields') || []
+  const decryptedFields = {}
+  for (const encryptedField of encryptedFields) {
+    const fieldName = await decrypt(encryptedField.name)
+    const fieldValue = await decrypt(encryptedField.value)
+    decryptedFields[fieldName] = fieldValue
+  }
+
+  const decryptedPassword = await decrypt(encryptedPassword)
+  const decryptedUsername = await decrypt(encryptedUsername)
+  if (decryptedPassword === null || decryptedUsername === null) {
+    throw new Error('DECRYPT_FAILED')
+  }
+  if (cipherDoc.deletedDate) {
+    // Soft delete from bitwarden
+    logger.debug('Remove accounts credentials and triggers...')
+
+    // Delete triggers
+    for (const account of accounts.data) {
+      const triggers = await fetchTriggersFromAccount(cozyClient, account)
+      for (const trigger of triggers) {
+        logger.debug(`Remove trigger ${trigger._id}`)
+        await cozyClient.destroy(trigger)
+      }
+    }
+    // Remove credentials
+    logger.debug('Remove accounts credentials...')
+    await updateAccountsAuth(cozyClient, accounts.data, {
+      login: decryptedUsername,
+      ...decryptedFields
+    })
+  } else {
+    // Restore from bitwarden
+    logger.debug('Restore accounts triggers...')
+
+    // Restore triggers
+    for (const account of accounts.data) {
+      const konnector = await fetchKonnectorFromAccount(cozyClient, account)
+
+      await ensureTrigger(cozyClient, {
+        account,
+        t,
+        konnector
+      })
+    }
+    // Note: there is no need to restore credentials here, because it is already
+    // done by the updateAccountsFromCipher service.
+  }
+}
+export default softDeleteOrRestoreAccounts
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
```diff
@@ -0,0 +1,85 @@
...
+  updateAccountsAuth
+} from './utils'
+import get from 'lodash/get'
+import { ensureTrigger } from '../connections/triggers'
+import { translate as t } from 'cozy-ui/transpiled/react/I18n'
+
+const makeDecrypt = (vaultClient, orgKey) => encryptedVal =>
+  decryptString(encryptedVal, vaultClient, orgKey)
+
+const softDeleteOrRestoreAccounts = async (
```
ptbrowne, line 17: Can you add a small doc explaining the flow (maybe copy/pasting the description of this PR here) ? The "updateAccountsFromCipher" bit, restoring credentials, was a bit unclear to me, and could be a little bit clearer I think : "updateAccountsFromCipher will restore the credentials from the linked cipher" ?

packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
```diff
@@ -0,0 +1,85 @@
...
+  const encryptedFields = get(cipherDoc, 'fields') || []
+  const decryptedFields = {}
+  for (const encryptedField of encryptedFields) {
+    const fieldName = await decrypt(encryptedField.name)
+    const fieldValue = await decrypt(encryptedField.value)
+    decryptedFields[fieldName] = fieldValue
+  }
+
+  const decryptedPassword = await decrypt(encryptedPassword)
+  const decryptedUsername = await decrypt(encryptedUsername)
```
ptbrowne, line 45: Nit: All this decrypting logic could be extracted as a function returning a decrypted object.

packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
```diff
@@ -0,0 +1,85 @@
...
+  if (decryptedPassword === null || decryptedUsername === null) {
+    throw new Error('DECRYPT_FAILED')
```
ptbrowne, line 47: A comment explaining why this could be the case would be helpful here.

packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+  if (cipherDoc.deletedDate) {
+    // Soft delete from bitwarden
```
ptbrowne, line 50:
```suggestion
    // The cipher has been soft deleted
```

packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
```diff
@@ -0,0 +1,85 @@
...
+  decryptString,
+  getOrganizationKey,
+  fetchAccountsForCipherId,
+  fetchKonnectorFromAccount,
+  fetchTriggersFromAccount,
+  updateAccountsAuth
+} from './utils'
+import get from 'lodash/get'
+import { ensureTrigger } from '../connections/triggers'
+import { translate as t } from 'cozy-ui/transpiled/react/I18n'
```
ptbrowne, line 12: 🤔 translate is a higher order function that decorates a React component so that it receives a `t` function in its props. It does not directly "translate" a locale string. It's the t function that should be passed to ensureTrigger, not translate.

To have a t function you can do

```
const Polyglot = require('node-polyglot')
const supportedLocales = ['en', 'fr', 'es']
const getT = () => {
  const locale = supportedLocales.find(l => l === process.env.COZY_LOCALE) || 'en'
  const locales = require(`../src/locales/${locale}.json`)
  const polygot = new Polyglot()
  polyglot.extend(locales)
  return polyglot.t.bind(polyglot)
}
```

packages/cozy-harvest-lib/src/services/utils.js
```diff
@@ -47,6 +47,31 @@ export const fetchAccountsForCipherId = async (cozyClient, cipherId) => {
...
+  const konnector = await cozyClient.query(
+    cozyClient.get(konnectorDoctype, slug)
+  )
+  return konnector.data
+    ? {
+        _id: konnector.data._id,
+        _type: konnector.data.type,
+        ...konnector.data.attributes
+      }
+    : null
```
ptbrowne, line 62: Why is the konnector data transformed here ? If you think that would be useful for other applications, it could be useful to have this logic in cozy-client directly (otherwise every other application will transform the data slightly differently, and it's harder to maintain).

packages/cozy-harvest-lib/src/services/utils.js
```diff
@@ -47,6 +47,31 @@ export const fetchAccountsForCipherId = async (cozyClient, cipherId) => {
...
+}
+
+export const fetchTriggersFromAccount = async (cozyClient, account) => {
+  const triggers = await cozyClient.queryAll(
+    cozyClient.find('io.cozy.triggers').where({
+      worker: 'konnector',
+      'message.account': account._id
+    })
+  )
```
ptbrowne, line 71: I wonder if we should not expose query definition creation utilities in cozy client.

```
import { queryTriggersFromAccount } from 'cozy-client'

const triggers = await cozyClient.queryAll(queryTriggersFromAccount(account))
```

ptbrowne: cc @Crash--

ptbrowne (commented):
The only blocker here is the `translate` one (see https://github.com/cozy/cozy-libs/pull/1104/files#r489193152). Otherwise LGTM, thanks 👍

packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
paultranvan: What I meant is that it's from bitwarden's routes: https://github.com/cozy/cozy-stack/blob/master/docs/bitwarden.md
I'll add a clarification

packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
paultranvan: Ok, thanks

packages/cozy-harvest-lib/src/services/utils.js
paultranvan: I agree, but I'm a bit worried that existing code already rely on this format. Maybe @doubleface has insights on this?

packages/cozy-harvest-lib/src/services/utils.js
ptbrowne: I think having this helper in cozy-client/src/models/triggers would make sense.

packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
```diff
@@ -14,6 +14,19 @@ import { translate as t } from 'cozy-ui/transpiled/react/I18n'
...
   decryptString(encryptedVal, vaultClient, orgKey)
 
+/**
+ * This service is useful to deal with soft delete and restore from
+ * bitwarden routes, used for a trash feature.
+ * See https://docs.cozy.io/en/cozy-stack/bitwarden/#put-bitwardenapiciphersiddelete
+
+ * When a cipher is soft deleted, a deletedDate field is added to the
+ * cipher. Then, the service removes the encrypted credentials from the
+ * referenced account and delete the associated trigger.
```
ptbrowne, line 24:
```suggestion
 * referenced account and deletes the associated trigger.
```

packages/cozy-harvest-lib/src/services/softDeleteOrRestoreAccounts.js
paultranvan: Right, but I prefer not moving it, because it will need some refactoring (decrypt is only available here) and adapt the tests

ptbrowne: I do not see why you need to adapt the tests 🤔  This function takes 70 lines, it starts to be big. Moving the creation of the decrypt function outside function would make it easier to see the bigger picture of this function.

packages/cozy-harvest-lib/src/services/utils.js
doubleface: The io.cozy.konnectors doctype is not used by connectors et connector do not use cozy-client yet

doubleface (approved):

ptbrowne (approved):

packages/cozy-harvest-lib/src/services/utils.js
paultranvan: I added a issue here https://github.com/cozy/cozy-client/issues/788

packages/cozy-harvest-lib/src/services/utils.js
paultranvan: I added an issue https://github.com/cozy/cozy-client/issues/789

PR #1107, opened 2020-09-17
feat: Throw error if comments are not spaced

Crash--:
Since I don't have the habit to put a space before the start of my comment, let's force me to create it ;)

WIP: need to test it. Also, I'm not sure about the error?

### DIFF
--- a/packages/eslint-config-cozy-app/basics.js
+++ b/packages/eslint-config-cozy-app/basics.js
@@ -19,6 +19,7 @@ module.exports = {
         semi: false
       }
     ],
-    'no-param-reassign': 'warn'
+    'no-param-reassign': 'warn',
+    'spaced-comment': ['error','always']
   }
 }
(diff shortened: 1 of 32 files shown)


packages/eslint-config-cozy-app/basics.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
         semi: false
       }
     ],
-    'no-param-reassign': 'warn'
+    'no-param-reassign': 'warn',
+    'spaced-comment': ['error','always']
```
ptbrowne, line 23: It's very easy to do the change if necessary.

Something like `codemod --extensions js,jsx "//([a-zA-Z0-9]O" "// \1"` should be enough.

So I think it's ok to make it an error.

Crash--: lint --fix should fix issues too ;)

ptbrowne:
@Crash-- Friendly ping :)

ptbrowne (approved):

PR #1110, opened 2020-09-28
fix(Harvest): Use cozy-realtime for OAuth popup communication

doubleface:
We now use cozy-realtime to send the accountId from the OAuth popup
to the origin harvest window.
postMessage api is more and more restricted depending on browser or
platform.

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/oauth.js
+++ b/packages/cozy-harvest-lib/src/helpers/oauth.js
@@ -1,6 +1,10 @@
 import uuid from 'uuid/v4'
 
 import * as konnectors from './konnectors'
+import CozyClient from 'cozy-client'
+import CozyRealtime from 'cozy-realtime'
+
+export const OAUTH_REALTIME_CHANNEL = 'oauth-popup'
 
 /**
  * Checks that the given data for the given konnector is consistent with the
@@ -29,6 +33,10 @@ export const checkOAuthData = (konnector, data) => {
 /**
  * Handler to be called at the top level of an application.
  * Aimed to handle an OAuth redirect with data in query string.
+ * @param  {Object} options.client (optional) : cozy-client instance which will be used to create
+ * cozy-realtime instance if not specified
+ * @param  {Object} options.realtime (optional) : cozy-realtime instance used to notify the
+ * resulting accountId to the origin harvest window after OAuth redirect
  *
  * We are supposed here to be in the OAuth popup, at the end of the process.
  *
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/helpers/oauth.js
```diff
@@ -41,15 +45,26 @@ export const checkOAuthData = (konnector, data) => {
...
+export const handleOAuthResponse = (options = {}) => {
+  let realtime = options.realtime
+  if (!realtime) {
+    let client = options.client
+    if (!client) {
+      const root = document.querySelector('[role=application]')
+      client = new CozyClient({
+        uri: `${window.location.protocol}//${root.dataset.cozyDomain}`,
+        token: root.dataset.cozyToken
+      })
```
ptbrowne, line 57: We should not need to create a new cozy-client here. Isn't there already created at app level ? It could be passed to this function.

packages/cozy-harvest-lib/src/helpers/oauth.js
```diff
@@ -41,15 +45,26 @@ export const checkOAuthData = (konnector, data) => {
  * ```
  * @return {boolean} `true` if an OAuth response has been handled, `false` otherwise
```
ptbrowne, line 46: Can you document options.{realtime,client}

ptbrowne:
Looking good, I've got minor thing on the client creation though.

packages/cozy-harvest-lib/src/helpers/oauth.js
doubleface: For example, in the home application, this function is called before any initialization ( https://github.com/cozy/cozy-home/blob/d89e562dacb1cbbe7f743488d6972f990fc24e7f/src/targets/browser/index.jsx#L17 ) to intercept OAuth redirect uri.

The other solution would be that the host application sends the cozy-client or cozy-realtime instance itself (which is possible with the new handleOAuthResponse signature).

But I don't want this change to cause regressions in the different applications using cozy-harvest-lib.

ptbrowne: Je vois [ici la méthode loadInstanceOptionsFromDOM](https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L116), à mon avis ce serait bien de l'utiliser. il me semble par contre qu'elle ne va pas appeler login() même si on arrive à avoir l'URL du cozy depuis le DOM, ca pourrait être qqchose à corriger. On a déjà des méthodes statiques fromEnv, fromOldClient, ca aurait sûrement du sens de faire la même chose. Est ce que tu peux voir si tu peux rajouter une méthode fromDOM dans cozy-client et potentiellement l'utiliser ici ?

ptbrowne: Since this is not really part of this PR and since this PR already provides value, if this is urgent, I'd be OK to merge now if we fix the problem later.

PR #1111, opened 2020-09-28
fix: Set peerDep same as devDep

Crash--:
Currently Harvest have a few old version in peerDep. We should kept them synched with the devDep?


ptbrowne (approved):
Good catch, thanks !

PR #1117, opened 2020-10-07
Tracking inside harvest

ptbrowne:
Adds trackPage and trackEvent calls to several components inside
harvest.

Those functions are taken from the context. This is the responsibility
of the app to provide the correct context, connecting the app tracker
to harvest.

If the context is not present, track calls will still work but won't
do nothing.


Crash-- (approved):
Can we add a note about `TrackingContext` in the Harvest's readme? 👍

ptbrowne:
Good idea for the doc, thanks : https://github.com/cozy/cozy-libs/pull/1117/commits/3c43f58f899790f0b4e19b5d31f4a0d49cc4054c.

PR #1118, opened 2020-10-12
feat(Harvest): Handle webauth connectors

doubleface:
Webauth is a protocol similar to OAuth but specific Budget Insight.

This PR needs https://github.com/cozy/cozy-stack/pull/2667 to be deployed in the stack to work properly + some specific secrets to be available.

For us, what is specific to webauth is that OAuth url must contain a temporary token generated by the connector and the resulting callback url, returned by BI will contain a new connection id which must be saved in the account.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/OAuthForm.jsx
+++ b/packages/cozy-harvest-lib/src/components/OAuthForm.jsx
@@ -5,7 +5,9 @@ import Button from 'cozy-ui/transpiled/react/Button'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
 import OAuthWindow from './OAuthWindow'
 import compose from 'lodash/flowRight'
+import get from 'lodash/get'
 import withConnectionFlow from '../models/withConnectionFlow'
+import { findKonnectorPolicy } from '../konnector-policies'
 
 /**
  * The OAuth Form is responsible for displaying a form for OAuth konnectors. It
@@ -17,6 +19,7 @@ export class OAuthForm extends PureComponent {
     this.handleAccountId = this.handleAccountId.bind(this)
     this.handleConnect = this.handleConnect.bind(this)
     this.handleOAuthCancel = this.handleOAuthCancel.bind(this)
+    this.handleMoreParams = this.handleMoreParams.bind(this)
     this.state = {
       initialValues: null,
       showingOAuthModal: false
@@ -24,14 +27,31 @@ export class OAuthForm extends PureComponent {
   }
 
   componentDidMount() {
-    const { account } = this.props
+    const { account, konnector, flow, client } = this.props
     this.setState({ initialValues: account ? account.oauth : null })
+
+    const konnectorPolicy = findKonnectorPolicy(konnector)
+
+    if (konnectorPolicy.fetchMoreOAuthUrlParams) {
+      this.setState({ needMoreParams: true })
+      konnectorPolicy.fetchMoreOAuthUrlParams({
+        flow,
+        konnector,
+        client,
+        onSuccess: this.handleMoreParams
+      })
+    }
+  }
+
+  handleMoreParams(moreParams) {
+    this.setState({ moreParams: moreParams })
   }
 
   handleAccountId(accountId) {
     const { onSuccess } = this.props
     this.hideOAuthWindow()
-    if (typeof onSuccess === 'function') onSuccess(accountId)
+    if (typeof onSuccess === 'function')
+      onSuccess(accountId, get(this.state, 'moreParams.id_connector'))
   }
 
   handleConnect() {
@@ -52,7 +72,12 @@ export class OAuthForm extends PureComponent {
 
   render() {
     const { konnector, t, flowState } = this.props
-    const { initialValues, showOAuthWindow } = this.state
+    const {
+      initialValues,
+      showOAuthWindow,
+      needMoreParams,
+      moreParams
+    } = this.state
     const isBusy = showOAuthWindow === true || flowState.running
 
     return initialValues ? null : (
@@ -65,8 +90,9 @@ export class OAuthForm extends PureComponent {
           label={t('oauth.connect.label')}
           onClick={this.handleConnect}
         />
-        {showOAuthWindow && (
+        {showOAuthWindow && (!needMoreParams || moreParams) && (
           <OAuthWindow
+            moreParams={moreParams}
             konnector={konnector}
             onSuccess={this.handleAccountId}
             onCancel={this.handleOAuthCancel}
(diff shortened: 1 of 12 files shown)


packages/cozy-harvest-lib/src/services/budget-insight.spec.js
```diff
@@ -60,7 +60,8 @@ const account = {
   _id: '1337',
   auth: {
     login: '80546578',
-    secret: 'secretsecret'
+    secret: 'secretsecret',
+    bankId: TEST_BANK_COZY_ID
```
ptbrowne, line 66: This does not seem to be tested, should we check that it is correctly passed to createTemporaryToken ?

packages/cozy-harvest-lib/src/components/OAuthForm.jsx
```diff
@@ -17,21 +19,39 @@ export class OAuthForm extends PureComponent {
...
   }
 
   componentDidMount() {
-    const { account } = this.props
+    const { account, konnector, flow, client } = this.props
     this.setState({ initialValues: account ? account.oauth : null })
+
+    const konnectorPolicy = findKonnectorPolicy(konnector)
+
+    if (konnectorPolicy.fetchMoreOAuthUrlParams) {
```
ptbrowne, line 35: s/More/Extra everywhere is more correct IMHO.

packages/cozy-harvest-lib/src/components/OAuthForm.jsx
```diff
@@ -17,21 +19,39 @@ export class OAuthForm extends PureComponent {
...
+      this.setState({ needMoreParams: true })
+      konnectorPolicy.fetchMoreOAuthUrlParams({
+        flow,
+        konnector,
+        client,
+        onSuccess: this.handleMoreParams
```
ptbrowne, line 41: Is there a particular reason here to not use Promises ?

const extraParams = await konnectorPolicy.fetchExtraOAuthUrlParams({...})
this.setState({ extraParams })

packages/cozy-harvest-lib/src/helpers/oauth.js
```diff
@@ -100,7 +100,8 @@ export const getOAuthUrl = ({
   oAuthStateKey,
   oAuthConf,
   nonce,
-  redirectSlug
+  redirectSlug,
+  moreParams
```
ptbrowne, line 104: This can be named extraParams. This could be documented in the docstring.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -97,16 +97,35 @@ export class DumbTriggerManager extends Component {
    * OAuth Form success handler. OAuthForm retrieves an account id created by the
    * cozy stack
    * @param  {string}  accountId
+   * @param  {integer} bankId
    */
-  async handleOAuthAccountId(accountId) {
+  async handleOAuthAccountId(accountId, bankId) {
```
ptbrowne, line 102: I am a bit surprised to see the banking notion leak here. I cannot see where this is called though. Would it be possible to have only generic oauth handling in TriggerManager and keep banking specific behaviors in the policy ?

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -200,6 +200,51 @@ export const getBIConnectionId = account => {
   return get(account, 'data.auth.bi.connId')
 }
 
+export const handleWebauthAccount = async ({
```
ptbrowne, line 203: Can you add a docstring ?

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -200,6 +200,51 @@ export const getBIConnectionId = account => {
...
+  return connectionId
+}
+
+/**
+ * Gets BI webauth connection id which is returned in the account by the stack
+ * via oauth callback url
+ *
+ * @param {io.cozy.accounts} The account content created by the stack
+ *
+ * @return {Integer} Connection Id
```
ptbrowne, line 242: It does not seem to be an Integer here https://github.com/cozy/cozy-libs/pull/1118/commits/9f71f8cdf51eee7fc4954b24296f444cfcf93e97#diff-8f55641483795f28fabc9c030803dd6aR355, can you change it to String ?

packages/cozy-harvest-lib/src/services/budget-insight.spec.js
```diff
@@ -298,3 +300,78 @@ describe('createOrUpdateBIConnection', () => {
...
+    })
+    const flow = new ConnectionFlow(client, { konnector, account })
+    flow.handleFormSubmit = jest.fn()
+    flow.saveAccount = async account => account
+    const account = { oauth: { query: { id_connection: ['12'] } } }
+    const konnector = {}
+    const t = jest.fn()
+    await handleWebauthAccount({
+      account,
+      bankId: 12,
```
ptbrowne, line 360: Can you use a different id that the connection_id, I find it a bit confusing.

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -97,16 +97,35 @@ export class DumbTriggerManager extends Component {
...
-      trigger: flow.trigger,
-      t: t
-    })
+    let oAuthAccount = await fetchAccount(client, accountId)
+
+    const konnectorPolicy = findKonnectorPolicy(konnector)
+
+    let foundWebauthConnection = false
+    if (konnectorPolicy.handleWebauthAccount) {
+      foundWebauthConnection = await konnectorPolicy.handleWebauthAccount({
```
ptbrowne, line 110: Webauth is very specific to BI. I think TriggerManager should not know about it. Could we find a generic name for the konnector policy method ?

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -192,7 +192,7 @@ export const resetState = account => ({
  * @param  {object} account   io.cozy.accounts document
  * @return {object}           io.cozy.accounts updated document
  */
-export const setSessionResetIfNecessary = (account, changedFields) => {
+export const setSessionResetIfNecessary = (account, changedFields = {}) => {
```
ptbrowne, line 195: Is this change necessary ? I do not see any added tests ?

packages/cozy-harvest-lib/src/services/budget-insight.spec.js
doubleface: CreateTemporaryToken has an assert for the presence of this attribute. And had to make this modification to allow the test to pass

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
doubleface: Right, and this is not needed. I will keep this bankId in BI policy

packages/cozy-harvest-lib/src/services/budget-insight.js
doubleface: But this is an integer. This is the result of getWebauthBIConnectionId function

packages/cozy-harvest-lib/src/components/OAuthForm.jsx
doubleface: Right, no good reason

ptbrowne (approved):

PR #1119, opened 2020-10-13
fix(Harvest): Better diffentiate bankId

doubleface:
The linxo bankId is the one saved into the account.
I renamed linxo bankId to linxoBankId et BI bankId to BiBankId

### DIFF
--- a/packages/cozy-harvest-lib/src/services/budget-insight.spec.js
+++ b/packages/cozy-harvest-lib/src/services/budget-insight.spec.js
@@ -335,12 +335,12 @@ describe('fetchExtraOAuthUrlParams', () => {
       }
     }
 
-    const onSuccess = ({ id_connector, token }) => {
-      expect(token).toEqual('bi-temporary-access-token-121212')
-      expect(id_connector).toEqual(TEST_BANK_BI_ID)
-    }
-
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/services/budget-insight.spec.js
```diff
@@ -368,7 +368,7 @@ describe('handleOAuthAccount', () => {
       t,
       account: {
         ...account,
-        ...{ auth: { bankId: TEST_BANK_BI_ID } },
+        ...{ auth: { bankId: TEST_BANK_COZY_ID } },
```
ptbrowne, line 371: TEST_BANK_LINXO_ID ?

ptbrowne (commented):
It's a bit weird to see linxo coming back here, I wonder if we should not name it the cozyBankId ?

packages/cozy-harvest-lib/src/services/budget-insight.spec.js
doubleface: I actually prefer to replace all references to Linxo by Cozy

ptbrowne (approved):
OK we still have the get getMappingLinxoBI leaking a bit our history with Linxo, but this can be renamed later if necessary. LGTM 👍

PR #1120, opened 2020-10-14
fix(Harvest): No more permission needed to open the popup

doubleface:
The browser popup blocker blocked the webauth popup because it appeared after an
asynchronous action (to get the temporary token) Now, the submit button is
displayed only when the token is already available and when the user clicks
on the submit button, the popup is displayed synchronously."

### DIFF
--- a/packages/cozy-harvest-lib/src/components/OAuthForm.jsx
+++ b/packages/cozy-harvest-lib/src/components/OAuthForm.jsx
@@ -81,15 +81,17 @@ export class OAuthForm extends PureComponent {
 
     return initialValues ? null : (
       <>
-        <Button
-          className="u-mt-1"
-          busy={isBusy}
-          disabled={isBusy}
-          extension="full"
-          label={t('oauth.connect.label')}
-          onClick={this.handleConnect}
-        />
-        {showOAuthWindow && (!needExtraParams || extraParams) && (
+        {(!needExtraParams || extraParams) && (
+          <Button
+            className="u-mt-1"
+            busy={isBusy}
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/components/OAuthForm.jsx
```diff
@@ -81,15 +81,17 @@ export class OAuthForm extends PureComponent {
...
-        />
-        {showOAuthWindow && (!needExtraParams || extraParams) && (
+        {(!needExtraParams || extraParams) && (
+          <Button
+            className="u-mt-1"
+            busy={isBusy}
+            disabled={isBusy}
+            extension="full"
+            label={t('oauth.connect.label')}
+            onClick={this.handleConnect}
```
ptbrowne, line 91: I wonder if we should not display it as disabled at first, instead of hiding it ? Isn't it weird to have a modal without button ?

packages/cozy-harvest-lib/src/components/OAuthForm.jsx
```diff
@@ -77,20 +77,21 @@ export class OAuthForm extends PureComponent {
...
-          />
-        )}
+        <Button
+          className="u-mt-1"
+          busy={isBusy}
+          disabled={isBusy}
+          extension="full"
+          label={t('oauth.connect.label')}
+          onClick={this.handleConnect}
+        />
```
ptbrowne, line 94: It could be good to have a test for this.

ptbrowne (approved):
LGTM

PR #1124, opened 2020-10-19
feat: Use @cozy/minilog instead of minilog

ptbrowne:
Our own fork solves a problem with octal literals in strict mode


Crash-- (approved):
Thanks, but I'm wondering why `minilog` is still present in the yarn.lock. Do you know why?

ptbrowne:
Good question, it will soon disappear when we remove minilog from cozy-client and cozy-keys-lib :)

PR #1126, opened 2020-10-20
Fix/sharing creation

Crash--:
- N'affiche plus les 2 types de partage quand le partage a été créé avec un rule.policy en read only vu qu'on ne peut pas changer le type de partage, ça ne sert à rien.  (au début, ça se base sur le sharingType dans le premier commit, mais c'est corrigé ensuite) avec l'ajout d'une méthode isReadOnly pour savoir si le partage est readonly ou pas. 
- Utilisation de la nouvelle API de create() et de addRecipients (besoin de https://github.com/cozy/cozy-client/pull/817) 
- Rewording pour ne plus avoir de sharingType / se baser sur le sharingType 
- Refacto pour utiliser les models de cozy-client et non de cozy-doctype

### DIFF
--- a/packages/cozy-sharing/src/DoctypeContact.js
+++ b/packages/cozy-sharing/src/DoctypeContact.js
@@ -1,27 +1,28 @@
+import { models } from 'cozy-client'
 import { Contact as DoctypeContact } from 'cozy-doctypes'
+const ContactModel = models.contact
 
-class Contact extends DoctypeContact {
-  static getInitials(contactOrRecipient, defaultValue = '') {
-    if (Contact.isContact(contactOrRecipient)) {
-      return DoctypeContact.getInitials(contactOrRecipient)
-    } else {
-      const s =
-        contactOrRecipient.public_name ||
-        contactOrRecipient.name ||
-        contactOrRecipient.email
-      return (s && s[0].toUpperCase()) || defaultValue
-    }
-  }
-
-  static getDisplayName(contact, defaultValue = '') {
-    if (Contact.isContact(contact)) {
-      return DoctypeContact.getDisplayName(contact)
-    } else {
-      return (
-        contact.public_name || contact.name || contact.email || defaultValue
-      )
-    }
+const isContact = candidate => {
+  return candidate._type === 'io.cozy.contacts'
+}
+export const getInitials = (contactOrRecipient, defaultValue = '') => {
+  if (isContact(contactOrRecipient)) {
+    return ContactModel.getInitials(contactOrRecipient)
+  } else {
+    const s =
+      contactOrRecipient.public_name ||
+      contactOrRecipient.name ||
+      contactOrRecipient.email
+    return (s && s[0].toUpperCase()) || defaultValue
   }
 }
 
-export default Contact
+export const getDisplayName = (contact, defaultValue = '') => {
+  if (isContact(contact)) {
+    return ContactModel.getDisplayName(contact)
+  } else {
+    return contact.public_name || contact.name || contact.email || defaultValue
+  }
+}
+
+export default DoctypeContact
--- a/packages/cozy-sharing/src/DoctypeContact.js
+++ b/packages/cozy-sharing/src/DoctypeContact.js
@@ -9,6 +9,7 @@ export const getInitials = (contactOrRecipient, defaultValue = '') => {
   if (isContact(contactOrRecipient)) {
     return ContactModel.getInitials(contactOrRecipient)
   } else {
+    // @todo Extract to RecipientModel ?
     const s =
       contactOrRecipient.public_name ||
       contactOrRecipient.name ||
@@ -21,6 +22,7 @@ export const getDisplayName = (contact, defaultValue = '') => {
   if (isContact(contact)) {
     return ContactModel.getDisplayName(contact)
   } else {
+    // @todo Extract to RecipientModel ?
     return contact.public_name || contact.name || contact.email || defaultValue
   }
 }
--- a/packages/cozy-sharing/src/DoctypeContact.spec.js
+++ b/packages/cozy-sharing/src/DoctypeContact.spec.js
@@ -1,4 +1,4 @@
-import Contact from './DoctypeContact'
+import { getDisplayName, getInitials } from './DoctypeContact'
 
 describe('Contact model', () => {
   describe('getInitials method', () => {
@@ -7,7 +7,7 @@ describe('Contact model', () => {
         name: 'whatever',
         public_name: 'janedoe'
       }
-      const result = Contact.getInitials(recipient)
+      const result = getInitials(recipient)
       expect(result).toEqual('J')
     })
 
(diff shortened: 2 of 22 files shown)


packages/cozy-sharing/src/components/EditableSharingModal.jsx
```diff
@@ -33,7 +33,8 @@ export const EditableSharingModal = ({
           revokeSharingLink,
           hasSharedParent,
           hasSharedChild,
-          revokeSelf
+          revokeSelf,
+          getSharingType
```
ptbrowne: It seems like a smell that we need to have all this in the context ? Why is it necessary ?

packages/cozy-sharing/src/DoctypeContact.js
```diff
@@ -1,27 +1,28 @@
...
-    } else {
-      return (
-        contact.public_name || contact.name || contact.email || defaultValue
-      )
-    }
+export const getDisplayName = (contact, defaultValue = '') => {
+  if (isContact(contact)) {
+    return ContactModel.getDisplayName(contact)
+  } else {
+    return contact.public_name || contact.name || contact.email || defaultValue
```
ptbrowne, line 24: Should the two else branch be in a Recipient model ? Would it make sense ?

packages/cozy-sharing/src/DoctypeContact.js
```diff
@@ -1,27 +1,28 @@
+import { models } from 'cozy-client'
```
ptbrowne, line 1: For my information, what was the warning that was solved by using cozy-client models ?

packages/cozy-sharing/src/components/ShareModal.jsx
```diff
@@ -86,7 +86,7 @@ export const ShareModal = ({
                 createContact={createContact}
                 onShare={onShare}
                 needsContactsPermission={needsContactsPermission}
-                sharingType={sharingType}
+                sharing={sharing}
```
ptbrowne, line 89: Does the DumbShareByEmail already can handle sharing ?

packages/cozy-sharing/src/components/ShareByEmail.jsx
```diff
@@ -120,22 +136,24 @@ export class ShareByEmail extends Component {
...
-  getSharingTypes = () => {
-    const { t } = this.props
-    const { sharingType } = this.state
-    const twoWay = {
-      value: 'two-way',
+  getSharingOptions = () => {
+    const { t, sharing, document } = this.props
+    const isSharingReadOnly = sharing
+      ? isReadOnlySharing(sharing, document._id)
+      : false
```
ptbrowne, line 143: For my culture, in which case can the `sharing` prop be undefined ?

packages/cozy-sharing/src/components/ShareByEmail.jsx
```diff
@@ -185,7 +203,7 @@ ShareByEmail.propTypes = {
   sharingDesc: PropTypes.string.isRequired,
   onShare: PropTypes.func.isRequired,
   createContact: PropTypes.func.isRequired,
-  sharingType: PropTypes.oneOfType([PropTypes.string, PropTypes.bool])
+  sharing: PropTypes.object
```
ptbrowne, line 206: Can you add a docstring and an explanation on why it's not required ?

packages/cozy-sharing/src/DoctypeContact.spec.js
```diff
@@ -1,4 +1,4 @@
-import Contact from './DoctypeContact'
+import { getDisplayName, getInitials } from './DoctypeContact'
```
ptbrowne, line 1: Nit: This commit could be fixed up with the commit improving getDisplayName.

packages/cozy-sharing/src/components/SharedStatus.spec.js
```diff
@@ -26,7 +26,6 @@ describe('SharedStatus component', () => {
             name: '2'
           }
         ]}
-        t={x => x}
```
ptbrowne, line 29: Same here, I think this commit could be fixed up.

packages/cozy-sharing/src/components/ShareByEmail.spec.jsx
```diff
@@ -22,13 +22,12 @@ describe('ShareByEmailComponent', () => {
...
     await comp.instance().share()
-    expect(onShare).toHaveBeenCalledWith(
-      {
-        id: 'doc_id'
-      },
-      [{ email: 'quentin@cozycloud.cc', id: 1 }],
-      'two-way',
-      'test'
-    )
+    expect(onShare).toHaveBeenCalledWith({
```
ptbrowne, line 25: Commit can be fixed up.

ptbrowne (commented):
I have some remarks.

packages/cozy-sharing/src/components/ShareModal.jsx
ptbrowne: I think this is solved in the next commit, the next commit could be fixed up.

packages/cozy-sharing/src/components/ShareByEmail.jsx
Crash--: If the sharing is not created. We display the same modal for sharing creation and sharing edition

packages/cozy-sharing/src/components/EditableSharingModal.jsx
Crash--: Cette histoire de passer le sharingType est remplacée par sharing par la suite

Mais aujourd'hui on passe l'ensemble des méthodes du sharing car on ne passait jamais le sharing en tant que tel. Donc on utilisait les méthodes du context pour récupérer les info en se basant sur le state. 

Maintenant que le sharing est passé, on va pouvoir commencer à faire des méthodes pour le manipuler sans dépendre du state et ce sera certainement mieux. 

Mais ce sera pour une autre fois

packages/cozy-sharing/src/DoctypeContact.js
Crash--: Commit edited but "Replace deprecated cozy-doctypes/contactModel methods"

packages/cozy-sharing/src/DoctypeContact.js
Crash--: I don't know, I added a todo for the next time

packages/cozy-sharing/src/components/ShareModal.jsx
Crash--: Hard to squash commits together here. I prefer to not spend to much time by splitting & rewriting them. This commit was before we removed the concept of sharingType and brings few things. It reflects the path I followed to resolve the issue

ptbrowne:
I had forgotten the last 2 comments in a pending review, sorry for that. LGTM though.

packages/cozy-sharing/src/components/ShareModal.jsx
ptbrowne: > It reflects the path I followed to resolve the issue

Yes I can see that :D 

By doing so it's harder for the reviewer to understand what has been done and the review quality is poorer as a result.

PR #1136, opened 2020-11-03
[green CI] Codemod improvements + svgr icons codemod

ptbrowne:
[green CI] See below, there is a check that is yellow but there are two other green checks, so everything looks good to me

- Merged @cozy/codemods and cozy-codeshift, I don't know why there were the two, surely an error from my part
- Reorganized helpers and utilities, utilities are now organized by concerns
- Simplify utililisation of helpers: we do not need to use makeUtils anymore. (previously makeUtils was used to pass `j` from the transform file to the utilities, but it was unnecessary and more complicated than to simply import `j` from 'jscodeshift')

```patch
- const utils = makeUtils(j)
- utils.imports.add(...)
+ const imports = require('../imports') // or const imports = require('@cozy/codemods/src/imports')
+ imports.add(...)
```

- Moved codemods from each repo to this repo : previously we could have codemods in cozy-client, in cozy-ui, in cozy-flags, but this makes it difficult to refactor the codemods. I think it will be simpler to have all our codemods in this repo. It's a tradeoff since one should work on a codemod in this repo if making a breaking change in another repo.
- Add SVGr icon codemods for cozy-ui

```patch
  import Icon from 'cozy-ui/transpiled/react/Icon'
- <Icon icon='left' />
+ import LeftIcon from 'cozy-ui/transpiled/react/Icons/Left'
+ <Icon icon={LeftIcon} />
```


packages/cozy-codemods/src/transforms/__testfixtures__/replace-icons-svgr.output.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import CustomIcon from './CustomIcon'
+
+import MagnifierIcon from "cozy-ui/transpiled/react/Icons/Magnifier";
```
Crash--, line 5: Can we have in output? If not, not a big deal since our linter will fix it. 

```jsx
import MagnifierIcon from 'cozy-ui/transpiled/react/Icons/Magnifier'
```

packages/cozy-codemods/src/transforms/transform-typography.js
```diff
@@ -0,0 +1,86 @@
...
+      root,
+      {
+        default: 'Typography'
+      },
+      'cozy-ui/transpiled/react/Typography'
+    )
+  }
+
+  return root.toSource()
+}
```
Crash--, line 86: Didn't know for this one :o

Crash-- (approved):

packages/cozy-codemods/src/transforms/__testfixtures__/replace-icons-svgr.output.js
ptbrowne: I don't see how to easily lint inside the transform runner how jest. I think I wrote in the commit that indeed, it was not veyr important to have semicolons in the output since we run our eslint afterwards.

PR #1141, opened 2020-11-04
Fix/share modal

Crash--:
Before we had only one modal to rule them all. The result was that we had CSS hacks in all directions to properly manage the display for all cases (only sharable by link, sharable cozy to cozy and by link, sharable cozy to cozy but disable and by link). 

The main idea here is to split the modal. One for sharable only by link and one for cozy-to-cozy (even disabled).

Few screenshots: 

Capture d’écran 2020-11-04 à 08 51 24
Capture d’écran 2020-11-04 à 10 28 26
Capture d’écran 2020-11-04 à 10 30 45
Capture d’écran 2020-11-04 à 10 30 48

Capture d’écran 2020-11-04 à 08 51 17

### DIFF
--- a/packages/cozy-sharing/src/components/ShareByLink.jsx
+++ b/packages/cozy-sharing/src/components/ShareByLink.jsx
@@ -109,7 +109,7 @@ class ShareByLink extends React.Component {
 
   render() {
     const { loading, menuIsOpen } = this.state
-    const { checked, documentType, permissions, t } = this.props
+    const { checked, documentType, permissions, t, popperOptions } = this.props
 
     const hasReadOnlyPermissions = checkIsReadOnlyPermissions(permissions)
 
@@ -168,6 +168,7 @@ class ShareByLink extends React.Component {
                     placement="bottom-end"
                     containerElRef={this.containerRef}
                     anchorElRef={this.containerRef}
+                    popperOptions={popperOptions ? popperOptions : undefined}
                   >
                     <ActionMenuItem
                       left={
--- a/packages/cozy-sharing/src/components/ShareByLink.jsx
+++ b/packages/cozy-sharing/src/components/ShareByLink.jsx
@@ -168,7 +168,7 @@ class ShareByLink extends React.Component {
                     placement="bottom-end"
                     containerElRef={this.containerRef}
                     anchorElRef={this.containerRef}
-                    popperOptions={popperOptions ? popperOptions : undefined}
+                    popperOptions={popperOptions}
                   >
                     <ActionMenuItem
                       left={
--- a/packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
+++ b/packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
@@ -36,6 +36,7 @@ const ShareDialogCozyToCozy = ({
   const { t } = useI18n()
   return (
     <FixedDialog
+      disableEnforceFocus
       opened={true}
       onClose={onClose}
       title={t(`${documentType}.share.title`, { name: document.name })}
--- a/packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
+++ b/packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
@@ -37,7 +37,7 @@ const ShareDialogCozyToCozy = ({
   return (
     <FixedDialog
       disableEnforceFocus
-      opened={true}
+      open={true}
       onClose={onClose}
       title={t(`${documentType}.share.title`, { name: document.name })}
       content={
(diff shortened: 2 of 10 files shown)


Crash--:
I created #1142 to create a standalone app / example to use all of those modales / components

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
```diff
@@ -0,0 +1,106 @@
...
+  groups,
+  createContact,
+  onShare,
+  needsContactsPermission,
+  sharing,
+  showWhoHasAccess,
+  isOwner,
+  onRevoke,
+  onRevokeSelf
+}) => {
```
ptbrowne, line 35: It seems like a smell to have so much props :/

packages/cozy-sharing/src/components/ShareByLink.jsx
```diff
@@ -168,6 +168,7 @@ class ShareByLink extends React.Component {
                     placement="bottom-end"
                     containerElRef={this.containerRef}
                     anchorElRef={this.containerRef}
+                    popperOptions={popperOptions ? popperOptions : undefined}
```
ptbrowne, line 171:
```suggestion
                    popperOptions={popperOptions}
```

Is there really a need for ternary condition ?

packages/cozy-sharing/src/components/ShareModal.jsx
```diff
@@ -46,72 +37,47 @@ export const ShareModal = ({
...
+          contacts={contacts}
+          groups={groups}
+          createContact={createContact}
+          onShare={onShare}
+          needsContactsPermission={needsContactsPermission}
+          sharing={sharing}
+          showWhoHasAccess={showWhoHasAccess}
+          isOwner={isOwner}
+          onRevoke={onRevoke}
+          onRevokeSelf={onRevokeSelf}
```
ptbrowne, line 77: Sorting the props would make them more readable.

ptbrowne (approved):
I think that it's good to detangle the components. So I think the code is going in the right direction.

I wonder if we find a way to reduce the number of props, maybe by grouping them together if they have the same concerns. I've also seen the contacts/groups props, if this is data, maybe the data could be used via redux selectors instead of props ?

A sharing context or hook could also maybe help ?

Anyway the state of the code seems better now so LGTM 👍

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
Crash--: I think the previous goal was to have "Dumb" components. I don't know why exactly but I think it is done like that to do not have to mock anything in our test to only pass the wanted props. 

This DumbComponent, is called in context of... a SharingConsumer. For instance: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-sharing/src/index.jsx#L509-L521 and then called itself in... an other SharingConsumer: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-sharing/src/components/EditableSharingModal.jsx#L21-L36

Today there is no plan to change this stuff, I am open to all ideas to improve this code. However, I'm going to create a kind of playground to make it easier to refacto and make sure that nothing breaks.

ptbrowne: > I am open to all ideas to improve this code. 

```jsx
 <DumbShareModal
    // can come from queryConnect or a redux selector
    contacts={contacts}

    // Not necessary
    createContact={contact => client.create(Contact.doctype, contact)}

    // Necessary
    document={document}

    // Can it be removed in favor of be document._type ? 
    documentType={documentType}


    groups={groups}
    
    // hasSharedChild could have the condition if (!documentPath) { return false } and then we can remove this prop
    hasSharedChild={documentPath && hasSharedChild(documentPath)}

    // hasSharedParent could have the condition if (!documentPath) { return false } and then we can remove this prop
    hasSharedParent={documentPath && hasSharedParent(documentPath)}

    // Not necessary, this should be done by the component
    isOwner={isOwner(document.id)}

    // Not necessary, this should be done by the component
    link={getSharingLink(document.id)}
    onRevoke={revoke}
    onRevokeLink={revokeSharingLink}
    onRevokeSelf={revokeSelf}
    onShare={share}
    onShareByLink={shareByLink}
    onUpdateShareLinkPermissions={updateDocumentPermissions}

    // Not necessary, can be done by the component
    permissions={getDocumentPermissions(document.id)}
  
    // Not necessary, can be done by the component
    recipients={getRecipients(document.id)}

    // Not necessary, can be done by the component
    sharing={getSharingForSelf(document.id)} />


 <DumbShareModal
    document={document}
    link={getSharingLink(document.id)}
    onRevoke={revoke}
    onRevokeLink={revokeSharingLink}
    onRevokeSelf={revokeSelf}
    onShare={share}
    onShareByLink={shareByLink}
    onUpdateShareLinkPermissions={updateDocumentPermissions} />
```

We could pair on something like that :)

PR #1146, opened 2020-11-10
Improve Harvest

ptbrowne:
- Use new Dialogs
- Use SVGr icons


packages/cozy-harvest-lib/src/components/AccountForm/ReadOnlyIdentifier.jsx
```diff
@@ -8,6 +8,8 @@ import { Media, Img, Bd } from 'cozy-ui/transpiled/react/Media'
```
Crash--, line 8: You can set the peerDep to 40 for UI

Crash-- (approved):

PR #1149, opened 2020-11-12
Feat/avatar has icon

Crash--:
Replace avatar icon="left" 

Is there a way to pass, I don't know, an array to `find()` in order to manage in only one method all our components with `icon` props instead of duplicating the code for each one?

### DIFF
--- a/packages/cozy-codemods/src/transforms/replace-icons-svgr.js
+++ b/packages/cozy-codemods/src/transforms/replace-icons-svgr.js
@@ -70,5 +70,36 @@ export default function replaceSvgrIcons(file, api) {
       }
     })
 
+  root
+    .find(j.JSXOpeningElement, {
+      name: {
+        name: 'Avatar'
+      }
+    })
+    .forEach(path => {
+      const iconAttr = path.node.attributes.find(
+        attr => attr.name && attr.name.name === 'icon'
+      )
+
+      if (iconAttr && iconAttr.value.type === 'Literal') {
+        const componentName = iconNameToComponentName(iconAttr.value.value)
+
+        imports.ensure(
+          root,
+          {
+            default: `${componentName}Icon`
+          },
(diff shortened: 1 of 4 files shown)


ptbrowne:
> Is there a way to pass, I don't know, an array to find() in order to manage in only one method all our components with icon props instead of duplicating the code for each one?

```
for (let component of [Avatar, Button, Icon]) {
  root.find(component).forEach(...)
}
```

?

packages/cozy-codemods/src/transforms/replace-icons-svgr.js
```diff
@@ -70,5 +70,36 @@ export default function replaceSvgrIcons(file, api) {
...
+        imports.ensure(
+          root,
+          {
+            default: `Icon`
+          },
+          `cozy-ui/transpiled/react/Icon`
+        )
+        iconAttr.value = `{<Icon icon={${componentName}Icon} />}`
+      }
+    })
```
JF-Cozy, line 103: Est-ce que ce ne serait pas intéressant de mettre une petite doc / commentaire pour expliquer pourquoi on fait ça ?

ptbrowne: A doc for who ?

Crash--:
> for (let component of ['Avatar', 'Button', 'Icon']) {
  root.find(j.JSXOpeningElement, { name: { name: component }}).forEach(...)
}

It was not clear enough, my question was more about find() itself not about any wrapper. I'll read jscodemod doc / code to see how it works under the wood

packages/cozy-codemods/src/transforms/replace-icons-svgr.js
JF-Cozy: J'avais pas vu qu'on était dans `cozy-codemods` :)

Crash--:
done in #1170

PR #1151, opened 2020-11-12
feat: Ability to update user config

ptbrowne:
User config is used to store user data before a transfer (KYC
regulations)

### DIFF
--- a/packages/cozy-harvest-lib/src/services/bi-http.js
+++ b/packages/cozy-harvest-lib/src/services/bi-http.js
@@ -73,6 +73,20 @@ const encodeToForm = rawForm => {
   return formData
 }
 
+export const getBIUserConfig = async (config, biAccessToken) => {
+  return await biRequest('GET', '/users/me/config', config, null, biAccessToken)
+}
+
+export const updateBIUserConfig = async (config, userConfig, biAccessToken) => {
+  return await biRequest(
+    'POST',
+    '/users/me/config',
+    config,
+    encodeToForm(userConfig),
+    biAccessToken
+  )
+}
+
 export const createBIConnection = async (
   config,
   encryptedAuth,
--- a/packages/cozy-harvest-lib/src/services/budget-insight.js
+++ b/packages/cozy-harvest-lib/src/services/budget-insight.js
@@ -15,6 +15,8 @@ import { waitForRealtimeEvent } from './jobUtils'
 import {
   createBIConnection,
   updateBIConnection,
+  getBIUserConfig,
+  updateBIUserConfig,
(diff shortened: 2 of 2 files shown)


packages/cozy-harvest-lib/src/services/bi-http.js
```diff
@@ -73,6 +73,20 @@ const encodeToForm = rawForm => {
   return formData
 }
 
+export const getBIUserConfig = async (config, biAccessToken) => {
+  return await biRequest('GET', '/users/me/config', config, null, biAccessToken)
+}
+
+export const updateBIUserConfig = async (config, userConfig, biAccessToken) => {
```
Crash--, line 80: nit: having the same arg order than getBIUserConfig can be great (config, biAccessToken, userConfig) ?

packages/cozy-harvest-lib/src/services/bi-http.js
```diff
@@ -73,6 +73,20 @@ const encodeToForm = rawForm => {
   return formData
 }
 
+export const getBIUserConfig = async (config, biAccessToken) => {
+  return await biRequest('GET', '/users/me/config', config, null, biAccessToken)
```
Crash--, line 77: nit: having on object for `config / null / biAccessToken` can be easier to maintain and understand

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -99,7 +101,7 @@ export const isBudgetInsightConnector = konnector => {
   )
 }
 
-const createTemporaryToken = async ({ client, biBankId, konnector }) => {
+export const createTemporaryToken = async ({ client, biBankId, konnector }) => {
```
Crash--, line 104: is it really used outside of the file?

Crash-- (approved):

packages/cozy-harvest-lib/src/services/bi-http.js
ptbrowne: Actually, the order is consistent, it's just that the biAccessToken is always last. I am not very happy with it either... I keep you advice in mind for next time.

packages/cozy-harvest-lib/src/services/budget-insight.js
ptbrowne: Not anymore, fixed, thanks :)

packages/cozy-harvest-lib/src/services/bi-http.js
ptbrowne: Agreed, will do it next time I need to add something, I am not very happy with this lib.

PR #1152, opened 2020-11-13
feat(Harvest): Ignore soft deleted ciphers

doubleface:
Ignore ciphers deleted by cozy-pass when proposing import on new account creation


packages/cozy-harvest-lib/src/components/VaultCiphersList/VaultCiphersList.spec.jsx
```diff
@@ -35,6 +35,40 @@ describe('when there is some ciphers', () => {
     expect(onSelect).toHaveBeenCalledWith(ciphers[0])
   })
 
+  it('should ignore trashed ciphers', async () => {
```
ptbrowne, line 38:
```suggestion
  it('should ignore soft deleted ciphers', async () => {
```

I think trashed here could indicate that we could recover the ciphers with the trash from drive.

ptbrowne (approved):

PR #1164, opened 2020-11-20
Dependency cleaning

ptbrowne:
I tried to clean up a bit our dependencies.

- Updated babel dependencies so we have the same deps everywhere.
- Be more lenient in our peer dependency version to have less warnings
  since the vast majority of time, our breaking dependencies do not
  break the cozy-libs packages depending on them
- Do not pin dependencies since it makes less sense to pin dependencies
  of libs. See https://docs.renovatebot.com/dependency-pinning/ for
  more information on why pinning dependencies is a bad idea for libs
- Ran yarn deduplicate to find and remove historical duplicates.
  See https://github.com/atlassian/yarn-deduplicate for more information
  on what type of duplicates yarn-deduplicates removes.
- Removed dependencies where possible
- Added a small checker that checks package.json for the above rules and
  reports if there is a problem. I am looking for feedback on this, the
  checker has not been added to travis yet but it could be. I am for example wondering
  if instead of using >= for all our peer dependencies we could only do that for
  dependencies where we now we have less risk (cozy-client and cozy-ui) for example
  and stay with ^ for other peer dependencies. In yarn 2, the concept of workspace
  constraints is there also, we could move to it in the future if this checker proves
  useful. 

Since I did not change code, only updated and moved dependencies, and the test pass, I do not
think I introduced breaking changes. The riskiest I guess is the update of babel-preset-cozy-app dependencies.


packages/playgrounds/package.json
```diff
@@ -24,8 +24,10 @@
     "redux": "3.7.2"
   },
   "devDependencies": {
-    "@babel/runtime": "^7.4.5",
     "babel-polyfill": "^6.26.0",
     "parcel": "^1.12.3"
+  },
+  "peerDependencies": {
+    "@babel/runtime": "7.12.5"
```
Crash--: Since `playgrounds` is an "application", babel/runtime should be in dep?

scripts/check-package.js
```diff
@@ -0,0 +1,173 @@
...
+    }
+  }
+}
+
+if (require.main === module) {
+  main().catch(e => {
+    console.error(e)
+    process.exit(1)
+  })
+}
```
Crash--, line 173: Wondering, should this script be in cozy-repo-doctor? Some functions can be very specific to cozy-lib's mono repo, but checking if the peer is met or else can be pretty generic

Crash-- (approved):
Thanks for the work @ptbrowne! Highly appreciated

scripts/check-package.js
ptbrowne: I also wondered about that, I think it could be a good idea. Would you be up for it ? :D

ptbrowne: I've added an issue because it is not so clear now to me what would be the advantages and how we could integrate the two : https://github.com/cozy/cozy-libs/issues/1165, thanks :)

doubleface (approved):

PR #1167, opened 2020-11-24
Solve dependency warning

ptbrowne:
When executing yarn test in cozy-harvest-lib, there was a
warning about two different versions of @testing-library/react
in dependencies and devDependencies.

- Removed @testing-library/react from dependencies
- Upgrade @testing-library/react in cozy-scanner to have the
  same version
- Fix the small issues that the upgrade caused


Crash-- (approved):
Nice!

PR #1171, opened 2020-11-26
Fix/scanner

Crash--:
- Use fixed dialog to have a better display on mobile
- remove the use of modal 
- fix an issue on Android (with a workaround for now)

### DIFF
--- a/packages/cozy-scanner/package.json
+++ b/packages/cozy-scanner/package.json
@@ -57,5 +57,8 @@
     ],
     "testEnvironment": "jest-environment-jsdom-sixteen"
   },
-  "sideEffects": false
+  "sideEffects": false,
+  "dependencies": {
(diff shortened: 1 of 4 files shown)


packages/cozy-scanner/package.json
```diff
@@ -57,5 +57,8 @@
     ],
     "testEnvironment": "jest-environment-jsdom-sixteen"
   },
-  "sideEffects": false
+  "sideEffects": false,
+  "dependencies": {
+    "cozy-device-helper": "^1.11.0"
```
ptbrowne, line 62: Why is this necessary ?

ptbrowne: Ah no sorry my bad

ptbrowne (approved):

PR #1179, opened 2020-12-04
feat(Harvest): Reuse unlinked accounts

doubleface:
Some accounts are not linked to any trigger anymore.
This is due to old bugs or when the user closes his window
too soon.

This PR will reuse these accounts, related to the same connector (via
account_type), fill them with new identifiers and link them to a new
trigger.

### DIFF
--- a/packages/cozy-harvest-lib/src/connections/accounts.js
+++ b/packages/cozy-harvest-lib/src/connections/accounts.js
@@ -195,9 +195,14 @@ export const fetchAccountsFromTriggers = async (client, triggers) => {
  * @param  {Array}  triggers io.cozy.triggers documents
  */
 export const fetchAccountsWithoutTriggers = async (client, triggers) => {
-  const accountCol = client.collection('io.cozy.accounts')
-  const accountIdToTrigger = keyBy(triggers, triggersModel.getAccountId)
-  return (await accountCol.all({ limit: null })).data.filter(
-    account => !accountIdToTrigger[account.id]
+  const triggerAccountIds = triggers.map(trigger =>
+    triggersModel.getAccountId(trigger)
   )
+  return (await client.query(
+    Q(ACCOUNTS_DOCTYPE).where({
+      _id: {
+        $nin: triggerAccountIds
+      }
+    })
+  )).data
 }
--- a/packages/cozy-harvest-lib/src/connections/accounts.js
+++ b/packages/cozy-harvest-lib/src/connections/accounts.js
@@ -187,3 +187,17 @@ export const fetchAccountsFromTriggers = async (client, triggers) => {
     .filter(Boolean)
     .map(account => ({ account, trigger: accountIdToTrigger[account._id] }))
 }
+
+/**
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/src/connections/accounts.js
```diff
@@ -187,3 +187,17 @@ export const fetchAccountsFromTriggers = async (client, triggers) => {
...
+ *
+ * @param  {Object}  client  CozyClient
+ * @param  {Array}  triggers io.cozy.triggers documents
+ */
+export const fetchAccountsWithoutTriggers = async (client, triggers) => {
+  const accountCol = client.collection('io.cozy.accounts')
+  const accountIdToTrigger = keyBy(triggers, triggersModel.getAccountId)
+  return (await accountCol.all({ limit: null })).data.filter(
+    account => !accountIdToTrigger[account.id]
+  )
```
Crash--, line 202: Can't we use a mango selector instead of fetching all the accounts? Does `fetchAccountsWithoutTriggers` will be called often? Since expensive 

Something like should work, no ? 
where({$not: { account : { $in : accountIdToTrigger } } }

packages/cozy-harvest-lib/src/connections/accounts.js
```diff
@@ -187,3 +187,17 @@ export const fetchAccountsFromTriggers = async (client, triggers) => {
...
+
+/**
+ * Returns the list of accounts which do not have any associated trigger
+ *
+ * @param  {Object}  client  CozyClient
+ * @param  {Array}  triggers io.cozy.triggers documents
+ */
+export const fetchAccountsWithoutTriggers = async (client, triggers) => {
+  const accountCol = client.collection('io.cozy.accounts')
+  const accountIdToTrigger = keyBy(triggers, triggersModel.getAccountId)
```
Crash--, line 199:
```suggestion
  const accountsIdToTrigger = keyBy(triggers, triggersModel.getAccountId)
```

packages/cozy-harvest-lib/src/connections/accounts.js
doubleface: This function is called only when creating a new account. If I had a selector to the request, I think I would have to handle paging and then I think this is a good compromise since this request is not so expensive and not called so often.

packages/cozy-harvest-lib/src/connections/accounts.js
doubleface: This map is made to associate a given accountId to the corresponding trigger. Sorry, I don't think I should change this name :-)

packages/cozy-harvest-lib/src/connections/accounts.js
ptbrowne: >  If I had a selector to the request,

You are already adding a selector here right ?

doubleface: Yes, I finally added a selector and used queryAll finally, which is fine

PR #1180, opened 2020-12-04
feat: Use partial index to speed up contact search

paultranvan:
This can greatly speed up contact search for sharing.

With 5K contacts, including only 1 relevant contact (having an email for instance) : 
* Without partial index : ~4700 ms
* With partial index : ~20 ms 

In any case, the query will be faster as the partial index filter out contacts without email/url and deleted contacts.

### DIFF
--- a/packages/cozy-sharing/package.json
+++ b/packages/cozy-sharing/package.json
@@ -41,7 +41,7 @@
     "babel-jest": "26.6.3",
     "babel-plugin-css-modules-transform": "^1.6.2",
     "babel-plugin-inline-react-svg": "^1.1.0",
-    "cozy-client": "16.0.0",
+    "cozy-client": "16.8.0",
     "cozy-ui": "40.9.1",
     "enzyme": "3.11.0",
     "enzyme-adapter-react-16": "1.15.2",
--- a/packages/cozy-sharing/src/components/EditableSharingModal.jsx
+++ b/packages/cozy-sharing/src/components/EditableSharingModal.jsx
@@ -97,6 +97,7 @@ const contactsQuery = () =>
       ]
     })
     .indexFields(['_id'])
+    .limitBy(1000)
 
 const groupsQuery = () => Q(Group.doctype)
 
--- a/packages/cozy-sharing/src/components/EditableSharingModal.jsx
+++ b/packages/cozy-sharing/src/components/EditableSharingModal.jsx
@@ -8,6 +8,7 @@ import SharingContext from '../context'
 import ContactsAndGroupsDataLoader from './ContactsAndGroupsDataLoader'
 import { default as DumbShareModal } from './ShareModal'
 import { useFetchDocumentPath } from './useFetchDocumentPath'
+
 export const EditableSharingModal = ({
   contacts,
   document,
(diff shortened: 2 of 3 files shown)


packages/cozy-sharing/src/components/EditableSharingModal.jsx
```diff
@@ -97,6 +97,7 @@ const contactsQuery = () =>
       ]
     })
     .indexFields(['_id'])
+    .limitBy(1000)
```
Crash--, line 103: Can you explain why this change? Does that mean we only fetch the first 100 (now 1000) contacts? Without any pagination? If yes, I don't know if this fix is the right one

paultranvan: From my tests, we had a pagination with a limit of 100, so I assumed there is some mechanism somewhere making the pagination?

Crash--: 100 is the default behavior from the stack. But we fetchMore by hands to fetch all the contacts here https://github.com/cozy/cozy-libs/blob/master/packages/cozy-sharing/src/components/ContactsAndGroupsDataLoader.jsx . So I assume this modification is good. 

But can you make an explanation in the commit message to say why this change? Since by default we only rely on the first 100 rows returned by the stack? We assume we can get the first 1000 because of the partial index?

packages/cozy-sharing/package.json
```diff
@@ -41,7 +41,7 @@
     "babel-jest": "26.6.3",
     "babel-plugin-css-modules-transform": "^1.6.2",
     "babel-plugin-inline-react-svg": "^1.1.0",
-    "cozy-client": "16.0.0",
+    "cozy-client": "16.8.0",
```
Crash--, line 44: If you take a look at the definition, you'll see that this change is not a "chore" commit https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#type

"feat: Upgrade cozy-client to have access to partialIndex" will be more meaningful

packages/cozy-sharing/src/components/EditableSharingModal.jsx
```diff
@@ -75,7 +76,9 @@ const contactsQuery = () =>
     .where({
       _id: {
         $gt: null
-      },
+      }
```
Crash--, line 79: I think you can also add your PR's description in this commit ;).

packages/cozy-sharing/src/components/EditableSharingModal.jsx
paultranvan: I set 1000 as a limit to be consistent with [contacts](https://github.com/cozy/cozy-contacts/blob/master/src/helpers/queries.js#L64). And, if 1000 contacts match, I think it's better in this case to have 1 big query of 1000 contacts than 10 of 100 contacts, as we quickly want all the contacts to search on them.
I'll add explanations in the commit.

packages/cozy-sharing/package.json
paultranvan: Fair enough :+1:

Crash-- (approved):
😘

PR #1181, opened 2020-12-07
Fix dialog not opening

ptbrowne:
After an update of cozy-ui, it is necessary to put the `open` prop
on Dialogs.

- [x] Added tests to test that confirm dialog modals are correctly showing


LucsT:
Good for me.

PR #1193, opened 2020-12-22
fix(Harvest): Remove check on identifiers for account rescue

doubleface:
When account is soft deleted by cozy pass, identifiers are removed on
the account and checking identifier to restore an account has no
meaning since identifiers will be overwritten just after.


paultranvan (approved):
Thanks!

PR #1194, opened 2020-12-23
feat: Unshare ciphers after account delete request

paultranvan:
When deleting a konnector account with a vault enabled, the cipher must be "unshared" with the cozy org, i.e. it must re-encrypted in end-to-end with the user key, both for security and UX (in Pass, the cipher is still marked as shared with cozy).

To this end, the user needs to unlock the vault when requesting an account deletion ; then the account is removed and the cipher is re-encrypted


packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/index.jsx
```diff
@@ -87,6 +93,18 @@ const ConfigurationTab = ({
 
   const hasLoginError = error && error.isLoginError()
 
+  const handleDeleteConfirm = async () => {
+    setRequestDeletion(false)
+    const extensionInstalled = await CozyUtils.checkHasInstalledExtension(
```
ptbrowne, line 98: If the user is on a device that does not have the extension installed, the cipher should be unshared right ?

paultranvan: Yes, but `checkHasInstalledExtension` does not says if the device has the extension, it says if it's been installed once, and so if the vault is set up.

ptbrowne: Shouldn't the condition be along the line of `hasVault && vault.isLocked == true` ?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx
```diff
@@ -95,14 +96,16 @@ const DumbKonnectorAccountTabs = props => {
...
+          <HarvestVaultProvider>
+            <ConfigurationTab
+              konnector={konnector}
+              account={account}
+              flow={flow}
+              addAccount={addAccount}
+              onAccountDeleted={onAccountDeleted}
+              showNewAccountButton={showNewAccountButton}
+            />
+          </HarvestVaultProvider>
```
ptbrowne, line 108: Could HarvestVaultProvider be at the route level ? I don't see why it should only wrap the ConfigurationTab ?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/index.jsx
ptbrowne: OK, weird naming then :/ The vault can already be unlocked right, can we check if it's unlocked ?

paultranvan: Yes the naming is definitely not ideal, it's because it's based on a field `extensionInstalled` in the settings, that was originally used for the Cozy Pass web app onboarding, but was later recycled to know if the vault was used or not. 

> The vault can already be unlocked right, can we check if it's unlocked 

This is actually a subject we're hoping to address for next year, but it's probably out of scope here, I understood it has security implications

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx
paultranvan: I guess it makes sense. So you mean wrapping here typically? https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/components/Routes.jsx#L22

ptbrowne: Yes.

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/index.jsx
ptbrowne: OK fine, thanks.

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx
paultranvan: Done in c753a08

PR #1199, opened 2021-01-08
fix: Close edit contract modal on escape

ptbrowne:
onClose was overriden by the dialogProps, we need to pass onClose to
useCozyDialog


ptbrowne:
Merged after green CI since this is super small

PR #1200, opened 2021-01-11
Fix: Sanitize account identifier to create folder path

### LINKED ISSUE #1192: Non-escaped characters in login when saving files creating sub-folders
Hi Cozy team,

I am developing a konnector to save files from another drive service (Reconnect) into my Cozy Drive. For that, I use the saveFiles function. 
I noticed that the files are stored in Administrative/konnector-name/konnector-login but in my case, the konnector-login contains "/" characters and so the files are stored in sub-folders. For example : Adminitrative/konnector-name/login-part1/login-part2/... 

**For example :**
Here is my login page for my konnector
image

The connection succeeded
image

But my files are stored in a sub-folder like this :

```
Drive
  └─ Administrative
          └─ Reconnect
                  └─ julien.pilote.01
                          └─ 01
                                  └─ 1975
                                          └─ myfiles 
```

I think that "/" characters in the user login should be escaped when saving files in the Cozy Drive in order to looklike this :

```
Drive
  └─ Administrative
          └─ Reconnect
                  └─ julien.pilote.01/01/1975
                          └─  myfiles 
```

ptbrowne:
fix #1192

If / are kept in the account identifier, nested folders might
be created if the account identifier contains slashes.


packages/cozy-harvest-lib/src/helpers/konnectors.spec.js
```diff
@@ -63,6 +63,30 @@ describe('Konnectors Helpers', () => {
...
+        buildFolderPath(fixtures.konnector, {
+          ...fixtures.account,
+          identifier: 'login',
+          auth: {
+            login: 'identifier/with/slashes'
+          }
+        })
+      ).toEqual('/Administrative/Test Konnector/identifier-with-slashes')
+    })
+
```
Crash--, line 89: i think this commit should be squashed with the next one?

ptbrowne: Yes :)

Crash-- (approved):

PR #1201, opened 2021-01-13
feat: Main harvest modal is hidden when vault unlock is opened

ptbrowne:
Previously, when the vault unlocker was displayed, it was displayed over 
the main harvest modal. The correct hiding of the main modal behind the
vault unlocker was difficult, and we could not use a narrower VaultUnlocker
since it would not hide completley the main modal. The problem was then to
hide the main modal when the vault unlocker is displayed.

The structure of the code made it difficult as the vault unlocker was a 
child of the main Harvest modal. Hiding the main modal would hide the vault
unlocker :/

Here, the code is reorganized so that TriggerManager is not responsible to
display the VaultUnlocker. Instead, it asks via its context, a component
near the root of Harvest (VaultUnlockPlaceholder) to display the
VaultUnlocker when necessary. At the same time, the HarvestDialog checks,
via the same context, if the VaultUnlocker is shown and hides itself if
true.

\+ smaller commits that are self explanatory


packages/cozy-harvest-lib/src/components/KonnectorAccounts.jsx
```diff
@@ -110,7 +109,7 @@ export class KonnectorAccounts extends React.Component {
     if (this.state.boundaryErrorInfo) {
       return (
         <div className="u-error">
-          <Bold>{this.state.boundaryError.message}</Bold>
+          <b>{this.state.boundaryError.message}</b>
```
Crash--, line 112: Why not <Typography variant="h6" > as stated here in the ui's doc: 
Capture d’écran 2021-01-13 à 09 06 44

packages/cozy-harvest-lib/src/components/hooks/useMaintenanceStatus.jsx
```diff
@@ -1,6 +1,14 @@
...
 import { Registry } from 'cozy-client'
 import get from 'lodash/get'
+import memoize from 'lodash/memoize'
+
+const memoizedFetchApp = memoize(
+  (registry, slug) => {
+    return registry.fetchApp(slug)
+  },
+  (registry, slug) => slug
+)
```
Crash--, line 11: I understand that you did that because this fetch can be slow (is it correct?) but having it memoized forever makes me inconfortable. I know that this is something we don't change often, but... 

I think the fix should rather be on the registry side of cozy-client so that it uses the redux store to persist data. So we would make sure to read the store first?

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -218,6 +199,32 @@ export class DumbTriggerManager extends Component {
...
+      konnector,
+      showUnlockForm,
+      onVaultDismiss,
+      vaultClosable,
+      vaultClient
+    } = this.props
+    const konnectorPolicy = findKonnectorPolicy(konnector)
+    const isVaultLocked = await vaultClient.isLocked()
+    if (konnectorPolicy.saveInVault) {
+      window.vaultClient = vaultClient
```
Crash--, line 213: Is it your new way to debug your code ? :D

packages/cozy-harvest-lib/src/components/vaultUnlockContext.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const withVaultUnlockContext = Component => {
+  const Wrapper = props => {
+    const vaultUnlockContextValue = useVaultUnlockContext()
+    return <Component {...props} {...vaultUnlockContextValue} />
+  }
+  Wrapper.displayName = `withVaultUnlockContext(${Component.displayName ||
+    Component.name})`
+  return Wrapper
+}
```
Crash--, line 64: I don't think so, but is it a breaking change? Harvest is used outside of Cozy ;)

Crash-- (approved):

packages/cozy-harvest-lib/src/components/KonnectorAccounts.jsx
ptbrowne: Done

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
ptbrowne: Yes ;) Removed, thanks

packages/cozy-harvest-lib/src/components/vaultUnlockContext.jsx
ptbrowne: I have added a BREAKING CHANGE mention

packages/cozy-harvest-lib/src/components/hooks/useMaintenanceStatus.jsx
ptbrowne: It's memoized until the user reloads the page, to have this functionality in cozy-client would be a much bigger endeavour. If you are unconfortable with this, I prefer to remove this commit.

Crash--: Yeah this is what I meant when using "forever" (until the user reloads the page). 

I'll create an issue on Cozy-Client to think about changing the Registry behavior. In the meantime, I just rise this potential issue, you deal with konnectors more often than me, so if you decide that we can memoize the result since we don't change the konnector's status pretty often, let's keep it 👍

PR #1205, opened 2021-01-18
Fix deletion of account

ptbrowne:
After account deletion, route was not correctly reset and 
subsequent navigation was broken. To remedy this problem, the
modal is closed optmistically.

### DIFF
--- a/packages/cozy-harvest-lib/src/locales/fr.json
+++ b/packages/cozy-harvest-lib/src/locales/fr.json
@@ -364,7 +364,9 @@
     },
     "updateAccount": {
       "general-subheader": "Général",
-      "identifiers": "Identifiants"
+      "identifiers": "Identifiants",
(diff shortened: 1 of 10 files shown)


packages/cozy-harvest-lib/src/locales/fr.json
```diff
@@ -364,7 +364,9 @@
     },
     "updateAccount": {
       "general-subheader": "Général",
-      "identifiers": "Identifiants"
+      "identifiers": "Identifiants",
+      "delete-account-success": "Le compte a été déconnecté",
+      "delete-account-error": "Le compte n'a pas pu être déconnecté, merci de réessayer plus tard, ou contactez notre support."
     }
   },
   "oauth": {
```
Crash--, line 372: I understand the fact you display an Alerter, but I don't see anything related to closing a dialog in this commit ? 🤔

Crash-- (approved):

PR #1210, opened 2021-01-20
Fix: Open KonnectorModal when no vault has been created

ptbrowne:
Due to a difference between the code checking if the
vault should be unlocked in keys and in harvest, the
konnector modal would open and close in an infinite
loop.

To fix this problem, I thought it would be simpler if
the code for the VaultUnlock system was moved to
cozy-keys-libs, so that only 1 lib is responsible for
the condition to unlock the vault.


Crash-- (approved):
Nice move 👍

PR #1213, opened 2021-01-22
feat: Add sentry integration to harvest connection flow

ptbrowne:
First step in the integration of Sentry. Next step would be to
integrate a React error boundary connected to the hub, wrapping all
the Harvest routes. Since sentry React integration uses the global hub,
we cannot use directly.

We use a Hub instead of a global Sentry since harvest is used by other apps.

### DIFF
--- a/packages/cozy-harvest-lib/src/sentry.js
+++ b/packages/cozy-harvest-lib/src/sentry.js
@@ -1,0 +1,21 @@
+import * as Sentry from '@sentry/browser'
+
+/**
+ * Creates a Sentry client connected to the Harvest project
+ *
+ * @param {SentryClient}
+ */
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/sentry.js
```diff
@@ -0,0 +1,21 @@
...
+  integrations: Sentry.defaultIntegrations,
+  beforeSend(event) {
+    if (process.env.NODE_ENV === 'production') {
+      return null
+    }
+    return event
+  }
+})
+
+export const hub = new Sentry.Hub(client)
```
Crash--, line 19: nice, didn't know about that!

Crash-- (approved):

PR #1214, opened 2021-01-25
fix(Harvest): Remove wrong 80s job watcher delay

doubleface:
There is already a default 8s value and 80s to wait for a login success is too long.


Crash-- (approved):
Thanks!

PR #1217, opened 2021-01-27
Fix bottom content swipeable view for Harvest

flohhhh:
- Remove updateSwiperHeight func cause not used
- Add .idea folder in gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 dist
 coverage
 docs/build
+.idea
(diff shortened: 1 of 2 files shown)


ptbrowne:
I do not understand why the function was not used, I am wondering if this would be a version problem with react-swipeable-view.

flohhhh:
Je voulais dire plus utilisé maintenant 🙃

ptbrowne (approved):
Do you know the cause of the bug ? Why didn't the dom mutations method work ?

flohhhh:
updateSwiperHeight was not called when there was a tab modification and I tried many ways to call updateHeight but no change.

I have not investigated more time on it

flohhhh:
For information: https://www.gitmemory.com/issue/oliviertassinari/react-swipeable-views/535/520230340

PR #1219, opened 2021-01-28
feat: Add a reconnect button inside trigger info to reconnect account

ptbrowne:
When a 2FA is needed, the user must go through the form to reconnect. 
Otherwise, by using the "update" button inside the data tab, only the job
will be relaunched, not the client side dance with budget insight. Since
this is the client side dance that triggers the 2FA modal, we need to
redirect the user to the connection form.


ptbrowne:
Duplicate.

PR #1224, opened 2021-02-01
docs: cozy-authentication

Crash--:
Add a few lines on how to use it with an exemple.

Also removed unused prop


packages/cozy-authentication/README.md
```diff
@@ -1,10 +1,56 @@
...
+          onLogout={() => alert('logout')}
+          appIcon={require('icon.png')}
+        >
+            <Route>
+                <Route path="path1" component={...} />
+                <Route path="path2" component={...} />
+            </Route>
+        </MobileRouter>
+}
+```
```
ptbrowne, line 49: I am not sure on having a static example here. I think they too often drift and become wrong. Do you think we could just keep the link to the playground example ?

ptbrowne:
Ping ? If you think the static example is OK, let's merge ?

PR #1227, opened 2021-02-03
fix: Don't crash if there is no value inside a permission

Crash--:
We have an use case when a permission looks like the one added
 in the fixture. Resulting by trying to iterate on undefined...

So if there is no value, there is no need to iterate on it since we
can't sort documents by id... 

Since we don't know exactly how to deal with this issue yet, I 
returned an empty [] as a quick fix.

### DIFF
--- a/packages/cozy-sharing/src/state.js
+++ b/packages/cozy-sharing/src/state.js
@@ -367,7 +367,7 @@ export const hasSharedChild = (state, documentPath) => {
 }
 
 // helpers
-const getSharedDocIds = doc =>
+export const getSharedDocIds = doc =>
   doc.type === 'io.cozy.sharings'
     ? getSharingDocIds(doc)
     : getPermissionDocIds(doc)
@@ -384,9 +384,16 @@ export const getSharingDocIds = sharing => {
   return docs
 }
 
-const getPermissionDocIds = perm =>
+// Some permissions can not have values since they can
+// be on a global doctype. In that case, we can't sort
+// them by id
+export const getPermissionDocIds = perm =>
   Object.keys(perm.attributes.permissions)
-    .map(k => perm.attributes.permissions[k].values)
+    .map(k =>
+      perm.attributes.permissions[k].values
+        ? perm.attributes.permissions[k].values
+        : []
+    )
     .reduce((acc, val) => [...acc, ...val], [])
 
 const areAllRecipientsRevoked = sharing =>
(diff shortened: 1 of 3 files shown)


packages/cozy-sharing/src/state.js
```diff
@@ -384,9 +384,16 @@ export const getSharingDocIds = sharing => {
...
+// them by id
+export const getPermissionDocIds = perm =>
   Object.keys(perm.attributes.permissions)
-    .map(k => perm.attributes.permissions[k].values)
+    .map(k =>
+      perm.attributes.permissions[k].values
+        ? perm.attributes.permissions[k].values
+        : []
+    )
     .reduce((acc, val) => [...acc, ...val], [])
```
ptbrowne, line 397: Nit: mapValues + flatten also does the job.

flatten(mapValues(perm.attributes.permissions, x => x || []))

packages/cozy-sharing/src/state.spec.js
```diff
@@ -487,6 +489,16 @@ describe('getSharedDocIdsBySharings method', () => {
   })
 })
 
+describe('getPermissionDocIds method', () => {
+  it('should return the permission by docsid', () => {
```
ptbrowne, line 493:
```suggestion
  it('should return the docs ids for the permission', () => {
```

packages/cozy-sharing/src/state.spec.js
```diff
@@ -487,6 +489,16 @@ describe('getSharedDocIdsBySharings method', () => {
...
+    const docs = getPermissionDocIds(PERM_2)
+    expect(docs).toEqual(['folder_2'])
+  })
+  it('should return nothing if there is no document', () => {
+    const docs = getPermissionDocIds(PERM_WIHTOUT_DOC)
+    expect(docs).toEqual([])
```
ptbrowne, line 499: It seems a bit surprising here to return an empty list when the permission is on the whole doctype. What happens if the permission has really no documents. Does it switch to global permission ? I think returning null here would be clearer.

ptbrowne (approved):

packages/cozy-sharing/src/state.spec.js
Crash--: Let me create the issue to discuss this thing because it's not clear to me how to correctly deal with that issue yet. For now, let's just fix the crash. (I edited the PR's description)

PR #1240, opened 2021-02-10
feat(Harvest): Set the current app as default redirect slug for OAuth

doubleface:
After the OAuth dance, the stack redirects to the home application by
default.

This sets the current application as the default redirect

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/oauth.js
+++ b/packages/cozy-harvest-lib/src/helpers/oauth.js
@@ -3,6 +3,7 @@ import uuid from 'uuid/v4'
 import * as konnectors from './konnectors'
 import CozyClient from 'cozy-client'
 import CozyRealtime from 'cozy-realtime'
+import get from 'lodash/get'
 
 export const OAUTH_REALTIME_CHANNEL = 'oauth-popup'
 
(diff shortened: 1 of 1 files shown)


packages/cozy-harvest-lib/src/helpers/oauth.js
```diff
@@ -128,6 +129,11 @@ export const getOAuthUrl = ({
   return oAuthUrl
 }
 
+const getAppSlug = () => {
+  const root = document.querySelector('[role=application]')
+  return get(root, 'dataset.cozyAppSlug')
```
Crash--, line 134: Querying the DOM like that involves that we only use Harvest within a DOM served by the cozy-stack. 

If we plan to use it in our next project, client side, we should not rely on that fact, right? 

Since we're in a context with a CozyClient instance https://github.com/cozy/cozy-libs/blob/2e552961595a1cab8d593cc9e014ca0d95f1b68e/packages/cozy-harvest-lib/src/helpers/oauth.js#L147 we can get the current application by reading client.appMetadata.slug

doubleface: Nice! Thanks!

ptbrowne (approved):

PR #1251, opened 2021-02-17
fix(Harvest): Missing account param for fetchExtraOAuthUrlParams

doubleface:
This caused an assert to fail in createTemporaryToken function in
BI service.

This error was visible with webauth connectors.


Crash-- (commented):
Can we add a test to prevent this kind of issue?

doubleface:
@Crash-- @ptbrowne Here are the unit tests

ptbrowne (approved):

PR #1259, opened 2021-02-24
feat(cozy-doctypes): Avoid a log for each transaction

doubleface:
In some cases (ex nestor connector with many transactions), the banking
reconciliator can generate a lot of logs, which is suspected to case
memory errors in connector's worker.

This logProgress added option allows to not log each transaction with
the `null` value and this option can be chosen by the connector.

### DIFF
--- a/packages/cozy-doctypes/src/banking/BankingReconciliator.js
+++ b/packages/cozy-doctypes/src/banking/BankingReconciliator.js
@@ -28,6 +28,26 @@ class BankingReconciliator {
     return { savedAccounts, reconciliatedAccounts }
   }
 
+  /**
+   * @typedef ReconciliatorResponse
+   * @attribute {Array<BankAccount>} accounts
+   * @attribute {Array<BankTransactions>} transactions
+   */
+
+  /**
+   * @typedef ReconciliatorSaveOptions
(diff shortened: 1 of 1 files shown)


ptbrowne (approved):

packages/cozy-doctypes/src/banking/BankingReconciliator.js
```diff
@@ -28,6 +28,14 @@ class BankingReconciliator {
     return { savedAccounts, reconciliatedAccounts }
   }
 
+  /**
+   * Save new accounts and transactions
+   *
+   * @param {array} fetchedAccounts
+   * @param {array} fetchedTransactions
+   * @param  {[type]} options   { logProgress: function }
+   * @returns {[type]} { accounts: savedAccounts, transactions: savedTransactions }
```
ptbrowne, line 37: Please add a proper type instead of "[type]", maybe with a @typedef.

```
    * @param {Array<BankAccount>} fetchedAccounts
    * @param {Array<BankTransactions>} fetchedTransactions

@typedef ReconciliatorResponse
@attribute {Array<BankAccount>} accounts
@attribute {Array<BankTransactions>} transactions
```

PR #1263, opened 2021-02-27
feat(Harvest): Handle multiple decoupled states

doubleface:
Some BI connectors may have multiple sources and sometimes, both sources
may need a decoupled 2FA. Ex Crédit Mutuel.

In case of a double decoupled 2FA before, harvest would send LOGIN_OK
event after the first decoupled 2FA.

Now, harvest will loop until the connection state is different from
'decoupled' or 'additionalInformationNeeded'

### DIFF
--- a/packages/cozy-harvest-lib/src/services/budget-insight.js
+++ b/packages/cozy-harvest-lib/src/services/budget-insight.js
@@ -267,8 +267,9 @@ export const resumeBIConnection = flow => {
  * @param  {ConnectionFlow} options.flow
  */
 export const finishConnection = async ({ flow }) => {
+  let connectionError
   while (flow.getData().biConnection.error) {
-    const connectionError = flow.getData().biConnection.error
+    connectionError = flow.getData().biConnection.error
     if (connectionError === DECOUPLED_ERROR) {
       const twoFAOptions = { type: 'app', retry: false }
       await flow.saveTwoFARequest(twoFAOptions)
@@ -283,7 +284,8 @@ export const finishConnection = async ({ flow }) => {
       await flow.waitForTwoFA()
       logger.debug('Finished waiting for 2FA...')
     } else {
-      throw convertBIErrortoKonnectorJobError(flow.getData().biConnection.error)
+      connectionError = flow.getData().biConnection.error
+      throw convertBIErrortoKonnectorJobError(connectionError)
     }
   }
   flow.triggerEvent(LOGIN_SUCCESS_EVENT)
--- a/packages/cozy-harvest-lib/src/services/budget-insight.js
+++ b/packages/cozy-harvest-lib/src/services/budget-insight.js
@@ -244,35 +244,49 @@ export const resumeBIConnection = flow => {
 }
 
 /**
- * Checks for 2FA or decoupled requests
+ * @typedef biConnection
+ * @property {number} id
+ * @property {number} id_user
+ * @property {number} id_connector
+ * @property {string|null} state  - ( wrongpass | additionalInformationNeeded | websiteUnavailable | actionNeeded | SCARequired | decoupled | passwordExpired | webauthRequired | rateLimiting | bug). Null indicates a success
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -244,35 +244,49 @@ export const resumeBIConnection = flow => {
...
+ *
+ * @param  {Object} options
+ * @param  {biConnection} options.biConnection
+ * @param  {ConnectionFlow} options.flow
  */
-export const finishConnection = async ({ biConnection, flow }) => {
-  if (biConnection.error) {
-    if (biConnection.error === DECOUPLED_ERROR) {
+export const finishConnection = async ({ flow }) => {
+  while (flow.getData().biConnection.error) {
```
Crash--, line 270: I don't know the code too much to have a strong opinion, but with this kind of loop we can get stuck for a long time? Can we have some kind of delay, for instance, in order to throw an error after X minutes ?

doubleface: Since the in this case, the loop is asynchronous and not connector is launched yet. The user still has control on the navigation.
I don't think this is necessary here.
I wondered if i should add a limit on the number of loops but couldn't define a proper limit and proper associated error message.

Crash-- (approved):

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -244,35 +244,49 @@ export const resumeBIConnection = flow => {
...
-        await flow.waitForTwoFA()
-      }
-      flow.triggerEvent(LOGIN_SUCCESS_EVENT)
+      await flow.saveTwoFARequest(twoFAOptions)
+      logger.debug('Waiting for 2FA...')
+      await flow.waitForTwoFA()
+      logger.debug('Finished waiting for 2FA...')
     } else {
-      throw convertBIErrortoKonnectorJobError(biConnection.error)
+      throw convertBIErrortoKonnectorJobError(flow.getData().biConnection.error)
```
ptbrowne, line 286: `const mostRecentConnectionError = flow.getData().biConnection.error` ?

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -244,35 +244,49 @@ export const resumeBIConnection = flow => {
...
+ * @param  {Object} options
+ * @param  {biConnection} options.biConnection
+ * @param  {ConnectionFlow} options.flow
  */
-export const finishConnection = async ({ biConnection, flow }) => {
-  if (biConnection.error) {
-    if (biConnection.error === DECOUPLED_ERROR) {
+export const finishConnection = async ({ flow }) => {
+  while (flow.getData().biConnection.error) {
+    const connectionError = flow.getData().biConnection.error
```
ptbrowne, line 271: Nit: connectionError could be declared with a let outside the loop not to have twice the `flow.getData().biConnection.error` assignment.

```
let connectionError
while (connectionError = flow.getData().biConnection.error) {
  ...

  // later on
  connectionError = flow.getData().biConnection.error
  throw convertBIErrortoKonnectorJobError(connectionError)
}
```

packages/cozy-harvest-lib/src/services/budget-insight.js
ptbrowne: I think it would make it clearer that after the wait, the connectionError could have changed and needs to be refreshed.

ptbrowne (approved):
LGTM, thanks !

packages/cozy-harvest-lib/src/services/budget-insight.js
doubleface: Since the flow can be updated in most of the previous lines, in both decoupled and additionalInformationNeeded cases. I don't see the point in doing this.

PR #1266, opened 2021-03-02
feat: Update encrypted tests

ptbrowne:
Re-encrypt tests with password from password store


doubleface (approved):
Tested with password-store password and it worked

PR #1267, opened 2021-03-02
cozy-sharing : muiv4

JF-Cozy:
Update to muiv4 for cozy-sharing.

BREAKING CHANGE: The application needs to use MUI v4 to works properly with this version of cozy-sharing


Crash--:
can you add a breaking change plz? Since apps need to be upgraded to mui v4, it's a Breaking Chance (since v4 brings a few BC on its own).

ptbrowne (approved):

PR #1269, opened 2021-03-04
fix: Don't require react in a node context

Crash--:


ptbrowne (approved):
Suggestion: In the node file (index.js), I think we could

* remove the typeof window !== undefined block, this way we have no conditional `require`
* without conditional require, we can use `import/export` instead of `require/module.exports` in both files.

If not possible, LGTM

Crash--:
I'll not have the time to do this today, but will make it tomorrow. I'll merge it in order to be able to update cozy-client & konnector-libs

PR #1270, opened 2021-03-04
feat: Add shortLabel prop to ShareButton

JF-Cozy:
To show short label when "shared with me" or "shared by me". It shows only "shared" in those cases.

### DIFF
--- a/packages/cozy-sharing/locales/en.json
+++ b/packages/cozy-sharing/locales/en.json
@@ -60,6 +60,7 @@
         }
       },
       "sharedByMe": "Shared by me",
+      "shared": "Shared",
       "sharedWithMe": "Shared with me",
       "sharedBy": "Shared by %{name}",
       "sharedByLink": "Shared by link",
@@ -141,6 +142,7 @@
         }
       },
       "sharedByMe": "Shared by me",
+      "shared": "Shared",
       "sharedWithMe": "Shared with me",
       "sharedBy": "Shared by %{name}",
       "shareByLink": {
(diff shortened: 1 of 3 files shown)


Crash-- (changes requested):
I don't think we should add this kind of props. 

WDYT about : 
```
<ShareButton 
labels={{
'sharedByMe': ...,
'shareWithMe': ...,
'shareCTA':
}} /> 
```
Or we could also export ShareButton without the withLocales. Like that you can pass the t you want without adding any props

ptbrowne:
I wonder if `shared={true|false}` prop wouldn't be better, and then the label could be conditioned on this prop ? I think it would be a bit closer to the business rule ? I do not know so well the usecase so I may be mistaken.

Crash--:
I don't think it's a good idea to add something like `shared={true|false}` . Does that mean that the document is shared? Or does that mean we want to display "shared" everywhere? 

I've an other option : dealing with the isMobile case directly in the ShareButton. I don't think we will one day display "Partagé avec moi" on a mobile screen. It really takes a lot a space. And since this component can be added everywhere, I'm wondering if the component should not deal itself with that case

After reviewing the first proposal, I think we can enhance it and makes it better by using "useShortLabel" true / false instead of shortLabel. At first, I though we're passing the "short label string" to that props

JF-Cozy:
> After reviewing the first proposal, I think we can enhance it and makes it better by using "useShortLabel" true / false instead of shortLabel. At first, I though we're passing the "short label string" to that props

Oui très bien, c'est effectivement plus clair

Crash-- (approved):

Issue #1272, opened 2021-03-05
cozy-flags


Crash--:
comments from @ptbrowne : 

 In the node file (index.js), I think we could
remove the typeof window !== undefined block, this way we have no conditional require
without conditional require, we can use import/export instead of require/module.exports in both files.
If not possible, LGTM

PR #1273, opened 2021-03-05
fix: Remove useless block for node

paultranvan:
See https://github.com/cozy/cozy-libs/issues/1272

### DIFF
--- a/packages/cozy-flags/src/index.js
+++ b/packages/cozy-flags/src/index.js
@@ -6,4 +6,4 @@ if (typeof __ENABLED_FLAGS__ !== 'undefined') {
   flag.enable(__ENABLED_FLAGS__)
 }
 
-module.exports = flag
+export default flag
--- a/packages/cozy-flags/src/index.js
+++ b/packages/cozy-flags/src/index.js
@@ -2,11 +2,6 @@
 
 import flag from './flag'
 
-if (typeof window !== 'undefined') {
-  flag.connect = require('./connect').default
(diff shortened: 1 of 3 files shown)


packages/cozy-flags/src/index.js
```diff
@@ -2,11 +2,6 @@
 
 import flag from './flag'
 
-if (typeof window !== 'undefined') {
-  flag.connect = require('./connect').default
-  flag.FlagSwitcher = require('./FlagSwitcher').default
-}
-
```
ptbrowne, line 9: can you use `export default flag` below ? and also remove the typeof check in the other file so that we can have

```
import connect from './connect'

flag.connect = connect
```

Using import and export exclusively is necessary for correct unused code pruning.

paultranvan: > remove the typeof check in the other file 

not sure what you mean by that? I assume you mean `index.browser.js` by "other file", maybe there is no `if (typeof window !== 'undefined') {`, if this is what you meant?

ptbrowne: Ah yes, no if, there is only the "module.exports/require" problem (require should be import/module.exports should be export).

paultranvan: Does 044db5e look good?

ptbrowne (approved):

packages/cozy-flags/src/index.js
ptbrowne: Yep, thanks.

Issue #1275, opened 2021-03-08
Choose a data card for a konnector


ptbrowne:
State right now

Right now, the approach to show datacards for a konnector is 

- Check permissions and see if there is a data card for it

Each doctype can be mapped to a doctype in harvest = harvest knows that for io.cozy.timeseries.geojson,
it should be display a GeoDataCard.

Questions:

- What happens when we'll have multiple permissions that have a data card (files and bills) ?
- Should the developer of the konnector be in control of what data cards are shown for its connector ?

Solutions:

- Choose only the first permission that has a viewer
- Have a ranking on viewers = if the konnector fetches bills and files, we only show bills (bills > files in our hypothetical ranking)
- Show all viewers in a tab like interface
- Manifest of the konnector could express which datacard are to be shown

Even if the manifest of the konnector could express which data card is to be shown, I still think that we should have a default behavior that makes sense without the developer of the konnector knowing about data cards (so that we do not have to go through all the manifests and add this behavior ourselves).

PR #1279, opened 2021-03-10
Files data card

ptbrowne:
- The permission "io.cozy.files" triggers the showing of this card. This will surely change in the future. A workshop is planned on how to decide whether to display a datacard, see https://github.com/cozy/cozy-libs/issues/1275
- Shows the last 5 files imported by the konnector for the account that is shown
- A click on the file brings the viewer
- useRealtime is used so that the list of files is updated when the konnector
  imports data
- This functionality is in testing right now and the flag harvest.datacards.files
  needs to be set

Next steps:

- Add an animation for new items entering the list. This will make the realtime
  more obvious
- Repair the "running" button of harvest, right now, after reopening the modal,
  the state is not good
- Add tests. Since I was in an exploratory phase and since we do not know yet
  if this code will survive, I plan to add tests when the flag is removed.
- Use bills instead of files, this way we'll have the amount/date of the bill

Those steps will be done in a future PR.
This future is testable on the following adhoc build:
  - git://github.com/cozy/cozy-home.git#build-files-datacard
  - flag('harvest.datacards.files', true)


packages/cozy-realtime/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "3.12.2",
   "description": "Realtime interactions with cozy-stack using Websocket",
   "main": "dist/index.js",
+  "main": "dist/index.browser.js",
```
Crash--, line 6: We now have two "main" entry point. I don't think this is intended?

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
```diff
@@ -0,0 +1,120 @@
...
+    .where({
+      'cozyMetadata.sourceAccount': accountId
+    })
+    .sortBy([
+      { 'cozyMetadata.sourceAccount': 'desc' },
+      { 'cozyMetadata.createdAt': 'desc' }
+    ])
+    .indexFields(['cozyMetadata.sourceAccount', 'cozyMetadata.createdAt'])
+    .limitBy(5),
+  as: `io.cozy.accounts/${accountId}/io.cozy.files`,
```
Crash--, line 151: You should probably put more information in the `as` attribute like: sortBy / limit ...

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
```diff
@@ -0,0 +1,120 @@
...
+  fetchPolicy: CozyClient.fetchPolicies.olderThan(30 & 1000)
```
Crash--, line 152: Is it really wanted to use this operator? Seems like a typo

packages/cozy-harvest-lib/src/datacards/mime-utils.js
```diff
@@ -0,0 +1,54 @@
...
+    return existingType ? mappingMimetypeSubtype[existingType[0]] : undefined
+  }
+  return undefined
+}
+
+export const getFileIcon = file => {
+  const subtype =
+    mappingMimetypeSubtype[file.class] || mappingMimetypeSubtype.text
+  return iconsBySubType[subtype]
+}
```
Crash--, line 53: Is it coming from drive? Should we extract and mutualize this kind of utilities somewhere?

packages/cozy-harvest-lib/src/models/getRelatedAppsSlugs.js
```diff
@@ -29,7 +29,7 @@ export const relatedAppsConfiguration = [
     slug: 'drive',
     priority: 0,
     predicate: ({ trigger }) => {
-      return !!get(trigger, 'message.folder_to_save')
+      return false && !!get(trigger, 'message.folder_to_save')
```
Crash--: to be removed

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
```diff
@@ -39,8 +46,51 @@ const LoadingFileListItem = ({ divider }) => {
...
+const FileMenu = ({ file, anchorEl, onClose }) => {
+  return (
+    <Menu
+      id={`file-menu-${file._id}`}
+      anchorEl={anchorEl}
+      keepMounted
+      open={Boolean(anchorEl)}
+      onClose={onClose}
+    >
+      <MenuItem onClick={onClose}>Open in Drive</MenuItem>
```
Crash--: translate()

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
```diff
@@ -39,8 +46,51 @@ const LoadingFileListItem = ({ divider }) => {
   )
 }
 
+const FileMenu = ({ file, anchorEl, onClose }) => {
+  return (
+    <Menu
```
Crash--: We should use UI's ActionMenu in order to have the right display on mobile device

packages/cozy-harvest-lib/package.json
```diff
@@ -31,6 +31,7 @@
     "cozy-bi-auth": "0.0.19",
     "cozy-doctypes": "^1.79.0",
     "cozy-logger": "^1.7.0",
+    "cozy-realtime": "3.12.2",
```
Crash--, line 34: For now, realtime was a peerDep of harvest. Do we want to add it as a dep?

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
```diff
@@ -181,8 +184,30 @@ const makeQueryFromProps = ({ accountId }) => ({
...
+  useRealtime(
+    client,
+    {
+      'io.cozy.files': {
+        created: onFileCreation,
+        updated: onFileCreation
+      }
+    },
+    [onFileCreation]
+  )
```
Crash--, line 177: I think we don't have to do this. We can use: 

`<RealTimeQueries doctype='io.cozy.files' />`. And since this component will dispatch `CozyClient's Mutation`, its store will be updated, and the result of the query will be updated too, no?

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
```diff
@@ -212,13 +220,19 @@ const FileDataCard = ({ filesCol, konnector, accountId }) => {
   const noFiles = hasQueryBeenLoaded(filesCol) && files.length == 0
   const isLoading = isQueryLoading(filesCol)
   return noFiles ? null : (
-    <FileCard files={files} loading={isLoading} konnector={konnector} />
+    <FileCard
+      files={files.slice(0, 5)}
```
Crash--, line 183: Wondering if we have this issue using RealTimequery

Crash-- (commented):
It remains to put all the logic of certifications in the viewer directly since we don't redirect to  Drive's viewer ^^.

Nice work

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
ptbrowne: Removed later on.

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
ptbrowne: This has been removed later one :)

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
ptbrowne: Done

packages/cozy-harvest-lib/package.json
ptbrowne: Nope, removed, kept as peer dependency, thanks !

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
ptbrowne: I added the "filecard" prefix not to have to repeat all the sort/limit etc..

packages/cozy-harvest-lib/src/datacards/mime-utils.js
ptbrowne: I created an issue on cozy-ui: https://github.com/cozy/cozy-ui/issues/1768, I am not planning to do this in this PR.

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
ptbrowne: Yep we still have the issue, queries inside cozy-client do not respect the limitBy parameter of the query.

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
ptbrowne: Nice, I was able to replace my code by RealTimeQueries easily !

Crash-- (approved):
👏

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
Crash--: Issue created https://github.com/cozy/cozy-client/issues/901

PR #1280, opened 2021-03-12
feat: Animate file list item when it is received via realtime

ptbrowne:
Use Slide from @material-ui to animate list items that are
received from realtime while a konnector is running.


Crash-- (approved):
Why not !  👍

PR #1282, opened 2021-03-17
Typography migration

flohhhh:
Migrating some components to the Typography component
for cozy sharing and cozy scanner


packages/cozy-sharing/src/components/ShareByLink.jsx
```diff
@@ -229,9 +245,9 @@ class ShareByLink extends React.Component {
                         this.deleteShareLink()
                       }}
                     >
-                      <Text className="u-pomegranate">
+                      <Typography className="u-pomegranate" variant="body1">
```
ptbrowne, line 248:
```suggestion
                      <Typography className="u-error" variant="body1">
```

ptbrowne: There is another instance of pomegranate in the same file, if it could be also taken care of via a className='u-error', same in Recipient.jsx.

ptbrowne (approved):
LGTM, thanks ! I've made a suggestion to take care of semantic class names for colors, but it could be done in another PR.

PR #1283, opened 2021-03-19
Improve files data card

ptbrowne:
- Refactor data card options to be more like getRelatedAppSlugs (the final
goal is to merge the two)
- Have the file data card be more powerful, not only does it fetch the
  sourceAccount linked files, but also files from the folder_to_save
  prop of the trigger (useful for connectors that do not implement
  the cozyMetadata.sourceAccount attribute)


packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
```diff
@@ -165,17 +165,14 @@ const FileCard = ({ files, loading, konnector, trigger }) => {
...
-      'cozyMetadata.sourceAccount': accountId,
+      dir_id: trigger.message.folder_to_save,
       trashed: false
     })
-    .indexFields(['cozyMetadata.sourceAccount', 'cozyMetadata.createdAt'])
-    .sortBy([
-      { 'cozyMetadata.sourceAccount': 'desc' },
-      { 'cozyMetadata.createdAt': 'desc' }
-    ])
+    .indexFields(['dir_id', 'cozyMetadata.createdAt'])
```
Crash--, line 177: I think you can also index by trashed

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
```diff
@@ -174,23 +176,58 @@ const makeQueryFromProps = ({ accountId, trigger }) => ({
...
+  as: `fileDataCard_io.cozy.files/${trigger.message.folder_to_save}/io.cozy.files`,
   fetchPolicy: CozyClient.fetchPolicies.olderThan(30 * 1000)
 })
 
-const FileDataCard = ({ filesCol, konnector, trigger }) => {
-  const { data: files } = filesCol
+const makeSourceAccountQueryFromProps = ({ accountId }) => ({
+  query: Q('io.cozy.files')
+    .where({
+      'cozyMetadata.sourceAccount': accountId,
```
Crash--, line 187: ~~This should be the only method used, no?~~ explained in the PR description. 

nit: adding the info in the commit message or in a comment can be easier to understand

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
```diff
@@ -174,23 +176,58 @@ const makeQueryFromProps = ({ accountId, trigger }) => ({
...
+  const isLoading =
+    isQueryLoading(folderToSaveFiles) || isQueryLoading(sourceAccountFiles)
+
+  const files = useMemo(() => {
+    return sortBy(uniq([...files1, ...files2], x => x._id), x =>
+      get(x, 'cozyMetadata.createdAt')
+    )
+      .reverse()
+      .slice(0, 5)
+  }, [files1, files2])
```
Crash--, line 223: Shouldn't we use only files from sourceAccountFiles if we have some?

Crash-- (commented):
can you update the readme too? https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/README.md#data-cards

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
ptbrowne: I do not see why we would do that ?

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
ptbrowne: I am not sure ? Is it really necessary, I do not think that a filter on the documents already filtered by the other conditions is going to take long ? What do you think @paultranvan ? Should we put all the fields always ?

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
ptbrowne: Added in the description of the commit.

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
Crash--: Because we can add / move files manually to the `destination folder` and these files will not be imported from the konnector but will be displayed as it it the case?

Crash--: Also, what about if we have 2 accounts for one konnector? If we list by the dest folder, we'll display the same documents twice, no?

ptbrowne:
I've updated the README, thanks.

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
paultranvan: For me it's ok not indexing it, as long as the percent of trashed docs remains low. In a worst case scenario, you might have all your docs trashed, and CouchDB will have to full scan the index to finally return nothing.
A possible alternative here is a partial index : 
```javascript
  query: Q('io.cozy.files')
    .where({
      dir_id: trigger.message.folder_to_save
    })
    .partialIndex(
        $or: [
          {
            trashed: {
              $exists: false
            }
          },
          {
            trashed: false
          }
        ]
      })
    .indexFields(['dir_id', 'cozyMetadata.createdAt'])
    .sortBy([{ dir_id: 'desc' }, { 'cozyMetadata.createdAt': 'desc' }])

 ```
 
 This way, the docs with a `trashed: true` field won't be indexed at all

paultranvan: BTW, we quite often need to filter out trashed docs, and this partial filter is a bit verbose, maybe we could add some helper on the dsl level like `.filterTrashed()` to handle this?

ptbrowne: OK thanks for the explanation, I think the state of the PR is ready then.

@paultranvan I've put your suggestion in an issue along with an idea : https://github.com/cozy/cozy-client/issues/909

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
Crash--: > Yes, in this case, but then there is not other way to discriminate a file that has been downloaded by the connector if there is no sourceAccount.

To me, if the request on the sourceAccount is returning something, we should not run the query on the destination folder since it's here only for a fallback, no? If there is no result on this sourceAccount query then we should query on the dest folder. 

> Also, what about if we have 2 accounts for one konnector? If we list by the dest folder, we'll display the same documents twice, no?

> I am not sure I understand, the documents are deduplicated by their _id, so we cannot have the same documents twice. If a konnector with multiple accounts arrange their files in the same folder, the files will be shown in the "account 1" list and in the "account 2" list, but as for the previous case, if the konnector does not save the sourceAccount (or has not saved the sourceAccount), we have no way of discriminating.

Let's say we have an OAuth account on the Foo konnector, we don't have much informations to create the destination folder, what will be the path resulting of the creation? `/Administrative/Foo/` ? or `/Administrative/Foo/XXX` where XXX is an ID coming from the OAuth connexion? 

My concern was about the fact we have 2 accounts connected on the same folder `/Administrative/Foo`, in this case, we'll ending up by displaying the same files for account 1 and account 2 if we do not discriminate by the sourceAccount (no dedup involved in this case). So this is the same result as 1, if we can discriminate on sourceAccount we should only use that. If the case is not possible since 2 accounts can't share the same destination folder, let's ignore this case (not sure it's more understandable 🤔 )

Crash-- (approved):
Only the remark about the queries but 👍

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
ptbrowne: > To me, if the request on the sourceAccount is returning something, we should not run the query on the destination folder since it's here only for a fallback, no?

If a konnector starts to add a sourceAccount on the data, you do not want the previously saved data to "disappear" from the list right ?

> Let's say we have an OAuth account on the Foo konnector, we don't have much informations to create the destination folder, what will be the path resulting of the creation? /Administrative/Foo/ ? or /Administrative/Foo/XXX where XXX is an ID coming from the OAuth connexion?

I think it depends on the connector, but I think we should have recent konnectors that create the administrative folder once they have enough information.

> My concern was about the fact we have 2 accounts connected on the same folder /Administrative/Foo, in this case, we'll ending up by displaying the same files for account 1 and account 2 if we do not discriminate by the sourceAccount 

Yes, this is true and I personally think it is a better behavior than to only show the files from the administrative folder. We have to keep in mind that as konnector will evolve, at some point the sourceAccounts files will be displayed in priority (thanks to the sort on creation date) over the files from the folder.

PR #1286, opened 2021-03-23
Lot1 config/init modal

flohhhh:
modale init config : 
- les étoiles ne doivent pas apparaitre quandd un autre champs est rempli
- pas d'étoile sur cette modale de config initiale à aucun moment

modale modif 
- quand on met le focus sur le chp mdp les étoiles doivent disparaitre
- quand on a pas le focus sur le chp mdp alors les étoiles doivent être présentes dans le chp mdp
- quand on arrive sur cette modale via bouton "reconnecter" alors le focus doit être sur le champ mdp 
  - dans le cas contraire focus sur chp identifiant


packages/cozy-harvest-lib/package.json
```diff
@@ -31,7 +31,7 @@
     "cozy-bi-auth": "0.0.19",
     "cozy-doctypes": "^1.80.0",
     "cozy-logger": "^1.7.0",
-    "cozy-ui": "^45.3.0",
+    "cozy-ui": "47.4.0",
```
ptbrowne, line 34: Please use a ^ here otherwise the version is unnecessarily restrictive.

- In apps we pin libraries
- In library we use the caret to be more lenient and allow an app to use (hopefull) only 1 version that satisfies all the dependency ranges of libraries.

packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
```diff
@@ -48,6 +48,16 @@ const DumbEditAccountModal = ({
...
    */
   const shouldShow = useTimeout(500)
+
+  const fromReconnectButton = window.location.toString().endsWith('reconnect')
+  // If we come from the reconnect button so focus on secret field
+  const fieldsOptions = {
+    secret: {
+      forceEncryptedPlaceholder: !fromReconnectButton,
+      focus: fromReconnectButton
+    }
```
ptbrowne, line 57: I wonder if it is necessary to have the generic "fieldsOptions" name where we in fact handle only the secret field options. Here it seems that we can do 

```
const fieldsOptions = {
  login: {
    focus: true
  }
}
```

but it does not work. To me either we use a generic name and then make it so that there no surprises or we can use specific names:

```
const fieldOptions = { // Nit: I prefer in names to use only 1 plural "s"
  displaySecretPlaceholder: true,
  focusSecretField: true
}
```

packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
```diff
@@ -193,13 +207,16 @@ export class AccountForm extends PureComponent {
...
       t,
+      fieldsOptions,
       flowState
     } = this.props
     const submitting = flowState.running
     const error = flowState.error
 
     const { fields } = konnector
     const sanitizedFields = manifest.sanitizeFields(fields)
+    manifest.addForceEncryptedPlaceholder(sanitizedFields, fieldsOptions)
```
ptbrowne, line 218: The forceEncryptedPlaceholder option is not supported by the manifest, only by the UI fields, so I would not put the helper function in the `manifest` file.

packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
```diff
@@ -48,6 +48,16 @@ const DumbEditAccountModal = ({
    * vault modal), we delay a bit the appearing of the edit modal via CSS.
    */
   const shouldShow = useTimeout(500)
+
+  const fromReconnectButton = window.location.toString().endsWith('reconnect')
```
ptbrowne, line 52: In banks, the hash history is not used, you should the router context to get a handle on the current route.

packages/cozy-harvest-lib/package.json
flohhhh: Thks, It's changed

packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
flohhhh: Move to fields helper

packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
flohhhh: Agree with you

ptbrowne:
Could you add some tests, the focus/blur logic seems a bit tricky ?

flohhhh:
> Could you add tests, the focus/blur logic seems a bit tricky ?

Sure, that's why it's in draft :)

ptbrowne:
Woops, I had not seen the draft, sorry 😅

packages/cozy-harvest-lib/src/components/AccountForm/index.jsx
```diff
@@ -193,13 +206,16 @@ export class AccountForm extends PureComponent {
...
       t,
+      fieldOptions,
       flowState
     } = this.props
     const submitting = flowState.running
     const error = flowState.error
 
     const { fields } = konnector
     const sanitizedFields = manifest.sanitizeFields(fields)
+    fieldsHelper.addForceEncryptedPlaceholder(sanitizedFields, fieldOptions)
```
ptbrowne, line 217: Nit: fieldHelpers since there will be several helpers ;)

packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
```diff
@@ -23,66 +23,75 @@ import { withMountPointPushHistory } from './MountPointContext'
...
-  trigger,
-  fetching,
-  redirectToAccount
-}) => {
-  const { dialogProps, dialogTitleProps } = useCozyDialog({
-    open: true,
-    size: 'm',
-    onClose: redirectToAccount
-  })
+const DumbEditAccountModal = withRouter(
```
ptbrowne, line 31: Nit: Have you tried to use `useLocation` ? This way there is no HOC involved.

ptbrowne (approved):
LGTM, thanks ! I have some nits if you have 5m ;)

packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
flohhhh: useLocation doest not work

ptbrowne: Do you know why ?

PR #1289, opened 2021-03-25
feat: Allow to pass onCreateOrUpdate error to bulkSave

ptbrowne:
This allows the user of bulkSave to decide what do to do with errors.
The onCreateOrUpdate callback receives (error, doc) as arguments and
its result will be put in the result array returned by bulkSave.
This way, the user can extract errors after the save and decide what
to do with them (maybe retry them)


doubleface (approved):

doubleface:
LGTM, this will fix a lot of conflicts in banking connectors and services.

PR #1290, opened 2021-03-26
fix: Remove default value

flohhhh:
When there is a new account to connect
We had stars when we wrote in identifier field


ptbrowne:
LGTM, I think a test here could be good.

PR #1296, opened 2021-03-31
feat: Add client side connector launcher

doubleface:
This adds a call to any react native launcher if :

1. the window.cozy.clientSideSlugs property contains a value corresponding to the current connector slug
2. window.ClientConnectorLauncher is initialized with `react-native` value

1) Won't be needed when the stack will handle the "clientSide" attribute in the manifest.

The call to the launcher provides the following values : 

- connector manifest
- account
- trigger
- job

Another addition : for client side connectors, the redirect to the success modale is deactivated when the connector finishes or sends the LOGIN_OK signal

### DIFF
--- a/packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
@@ -70,11 +70,19 @@ const NewAccountModal = ({ konnector, onDismiss }) => {
             konnector={konnector}
             onLoginSuccess={trigger => {
               const accountId = triggersModel.getAccountId(trigger)
-              pushHistory(`/accounts/${accountId}/success`)
+              let path = `/accounts/${accountId}`
+              if (trigger.worker !== 'client') {
+                path += '/success'
+              }
+              pushHistory(path)
             }}
             onSuccess={trigger => {
               const accountId = triggersModel.getAccountId(trigger)
-              pushHistory(`/accounts/${accountId}/success`)
+              let path = `/accounts/${accountId}`
+              if (trigger.worker !== 'client') {
+                path += '/success'
+              }
+              pushHistory(path)
             }}
             onVaultDismiss={onDismiss}
             fieldOptions={fieldOptions}
--- a/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
+++ b/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
@@ -496,6 +496,7 @@ export class ConnectionFlow {
    */
   async launch({ autoSuccessTimer = true } = {}) {
     const konnector = this.withClientSide(this.konnector)
+    const computedAutoSuccessTimer = autoSuccessTimer && !konnector.clientSide
 
     logger.info('ConnectionFlow: Launching job...')
     this.setState({ status: PENDING })
@@ -541,7 +542,7 @@ export class ConnectionFlow {
       this.handleJobUpdated.bind(this)
     )
     this.jobWatcher = watchKonnectorJob(this.client, this.job, {
-      autoSuccessTimer
+      autoSuccessTimer: computedAutoSuccessTimer
     })
     logger.info(`ConnectionFlow: Subscribed to ${JOBS_DOCTYPE}:${this.job._id}`)
 
--- a/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
+++ b/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
@@ -410,7 +410,7 @@ export class ConnectionFlow {
       await this.ensureTriggerAndLaunch(client, {
         trigger,
         account,
-        konnector: this.withClientSide(konnector),
+        konnector,
         t
       })
       this.setState({ accountError: null })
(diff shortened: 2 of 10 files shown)


packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -497,6 +497,26 @@ export class ConnectionFlow {
...
+          JSON.stringify({
+            message: 'startLauncher',
+            value: {
+              connector: this.konnector,
+              account: this.account,
+              trigger: this.trigger,
+              job: this.job
+            }
+          })
+        )
```
Crash--, line 514: We should create a `Bridge` or a `CommunicationChannel` between Harvest and the launcher. 

At first we only need to send this `startLauncher` command to a RN environnement, but we can imagine having other channels. 

Also, for now we rely only on realtime socket, but we can easily miss a message on the client side. We would have to listen to event coming from the launcher 

Something similar to CozyRealtime can be imagined: 
```
Bridge.send({
   message: 'startLauncher',
            value: {
              connector: this.konnector,
              account: this.account,
              trigger: this.trigger,
              job: this.job
            }
})

Bridge.subscribe('...')
```

Is it planned in the future?

doubleface: Yes

packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
```diff
@@ -70,11 +70,19 @@ const NewAccountModal = ({ konnector, onDismiss }) => {
...
               const accountId = triggersModel.getAccountId(trigger)
-              pushHistory(`/accounts/${accountId}/success`)
+              let path = `/accounts/${accountId}`
+              if (trigger.worker !== 'client') {
+                path += '/success'
+              }
+              pushHistory(path)
             }}
             onVaultDismiss={onDismiss}
             fieldOptions={fieldOptions}
```
Crash--, line 88: We should definitely add a test for this case.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -368,10 +369,11 @@ export class ConnectionFlow {
       this.trigger = trigger
       this.account = account
       this.konnector = konnector
+
       this.t = t
 
       assert(client, 'No client')
-      const konnectorPolicy = findKonnectorPolicy(konnector)
+      const konnectorPolicy = findKonnectorPolicy(this.konnector)
```
Crash--, line 376: Why do we need this change? It seems that `handleFormSubmit` is call with options.konnector (https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/components/TriggerManager.jsx#L140) 

Sometimes ConnectionFlow is instanciated without `konnector` in the constructor https://github.com/cozy/cozy-libs/blob/ba70b8fb5b7b95196d3bec51942c3a56410712c1/packages/cozy-harvest-lib/src/cli/cli.js#L69 so using this.konnector will result in breaking at least this method

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -477,10 +479,24 @@ export class ConnectionFlow {
...
+   * Add clientSide attribute to the konnector manifest according to the context
+   *
+   * @todo When the stack handles the clientSide attribute, this won't be needed anymore
+   */
+  withClientSide(konnector) {
+    const clientSideSlugs = get(window, 'cozy.clientSideSlugs', [])
+    const clientSide = clientSideSlugs.includes(konnector.slug)
+
+    return { ...konnector, clientSide }
+  }
```
Crash--, line 492: Since the stack handle this attribute can we use it? (https://github.com/cozy/cozy-stack/pull/2976)

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -477,10 +479,24 @@ export class ConnectionFlow {
...
+
   /**
    * Launches the job and sets everything up to follow execution.
    */
   async launch({ autoSuccessTimer = true } = {}) {
+    const konnector = this.withClientSide(this.konnector)
```
Crash--, line 498: I don't think we need this code anymore this the trigger has the `client` attribute, right? (again relying on this.konnector will break stuff see previous comment). 

Generally we should not add a dep to the konnector when launching a trigger

Crash-- (changes requested):
few remarks

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: This attribute is not handled. It is possible to create and run "client" triggers. But if we install a connector with `clientSide` attribute in the manifest, the attribute is still not visible in io.cozy.konnectors doctype.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: Same as https://github.com/cozy/cozy-libs/pull/1296#discussion_r689681060 . Harvest does not know it should create a client trigger if it cannot find the clientSide attribute in io.cozy.konnectors.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
Crash--: Can you propose what you want to add in the manifest / konnectors doctype? I'm sure that when we have that @nono will be happy to add it when installing the connector.

doubleface: https://github.com/cozy/cozy-stack/issues/3103

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: I remember there are some cases where options.konnector is null. I am looking for it at the moment.

doubleface: I could not reproduce the problem and I can't see why option.konnector would be null by reading the code. I remove this change.

packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
doubleface: Here it is

doubleface:
@Crash-- Je ne sais pas si cette PR est mergable compte tenu du fait qu'une partie du code est là pour ajouter l'attribut clientSide et sera donc obsolète une fois la gestion de cet attribut ajoutée dans la stack.

A part ça, je pense avoir répondu à tous les commentaires

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: Done. This is now possible since https://github.com/cozy/cozy-stack/pull/3113

Crash-- (approved):

PR #1298, opened 2021-04-07
feat: Support two fa in launch mode

ptbrowne:
Now, a user can launch the connector via the launch button and have the
two fa modals.

When launching the konnector, since we do not have the BI connection
cached (we do not launch a temporary token job, the front cannot access
budget-insight), we send the two fa codes to the konnector job via
the regular two fa APIs.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
+++ b/packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
@@ -4,14 +4,14 @@ import PropTypes from 'prop-types'
 import logger from '../../logger'
 import Card from 'cozy-ui/transpiled/react/Card'
 import Divider from 'cozy-ui/transpiled/react/MuiCozyTheme/Divider'
-import { ButtonLink } from 'cozy-ui/transpiled/react/Button'
+import Typography from 'cozy-ui/transpiled/react/Typography'
+import Link from '@material-ui/core/Link'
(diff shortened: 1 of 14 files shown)


doubleface (approved):

packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx
```diff
@@ -34,18 +33,19 @@ const WebsiteLinkCard = ({ link }) => {
...
         <Img className="u-pr-1">
-          <Icon icon={GlobeIcon} color={palette['coolGrey']} />
+          <Typography color="textSecondary">
+            <Icon icon={GlobeIcon} />
+          </Typography>
         </Img>
         <Bd>
-          <ButtonLink
-            subtle
+          <Link
```
Crash--, line 41: Shouldn't the label be capitalized? This is the goal of ButtonLink, no?

ptbrowne: The label is not capitalized at the moment, see this screenshot.

image

Crash--: Oh ok, it was right in front of my eyes https://github.com/cozy/cozy-libs/blob/ef670f968b0ef7d0550db3ff5eb8a7d4100bc66c/packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.jsx#L14 🤦

PR #1304, opened 2021-04-15
Small improvements for Harvest

ptbrowne:
feat: Remove references to "cozy" in harvest locales
Makes it easier to customize and whitelabel the UI

Show konnector error in both data and configuration tab in harvest

Open external links in a new tab not to lose the Cozy context


Crash-- (approved):
Looks good. 

I just have a little reservation about removing "Cozy" from sentences. I don't know if we have a guideline for the whole platform cc @poupotte

ptbrowne:
cc @fffflo

PR #1307, opened 2021-04-21
Fix/flag: Cozy-Client import

Crash--:
Since https://github.com/cozy/cozy-libs/pull/1235 we import cozy-client without announcing it. Also, let's be specific for the path to import in order to avoid issues during test / build


flohhhh:
What was the problem ? if there was one ^^

flohhhh (approved):

PR #1317, opened 2021-05-12
fix: Revert change on removal of browser

ptbrowne:
The browser entrypoint was removed by error in a unrelated commit.
The commit in question a0856a5fc4b1701376dddabe9065b521a6a0294a

This had the consequence that we could not access browser only methods
after updating cozy-flags.


Crash--:
Oups wrong button! Wanted to approve, not merging it myself. Sorry 🤷‍♂️

PR #1318, opened 2021-05-21
fix: BI Two FA naming error

doubleface:
There was an error in the naming of the sendTwoFa function
in the BI policy.
This caused 2FA codes not to be sent to BI and harvest was keeping
asking a new code indefinitely.

### DIFF
--- a/packages/cozy-harvest-lib/src/services/budget-insight.js
+++ b/packages/cozy-harvest-lib/src/services/budget-insight.js
@@ -366,7 +366,7 @@ export const onBIAccountCreation = async ({
   return account
 }
 
-export const sendTwoFaCode = (flow, code) => {
+export const sendTwoFACode = (flow, code) => {
   let fields
 
   try {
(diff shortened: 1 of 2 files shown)


Crash--:
@doubleface you should remove some commits 🙏

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -363,7 +363,7 @@ export const onBIAccountCreation = async ({
   return account
 }
 
-export const sendTwoFaCode = (flow, code) => {
+export const sendTwoFACode = (flow, code) => {
```
Crash--, line 366: I think test will fail since in the spec we have
```
   sendTwoFaCode
} from './budget-insight' 
```

doubleface: Thanks, I fixed unit tests

PR #1321, opened 2021-06-11
fix: Label height for Account Form

Crash--:
Before, we set the fontSize to 0 in order to not display the
ObfuscateChar. But by doing so, I do know why, but the
height of the label is not the good one at least on Chromium's
based browser.

To avoid this issue, let's use display none. It will be easier

for Password manager to detect our trick, but we already have
an aria-label saying the same thing. So ...

Before
before

After
after

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm/Field.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/Field.jsx
@@ -3,7 +3,7 @@ import UIField from 'cozy-ui/transpiled/react/Field'
 
 const ObfuscateChar = () => {
   return (
-    <span style={{ fontSize: 0 }} aria-hidden="true">
+    <span style={{ display: 'none' }} aria-hidden="true">
       ·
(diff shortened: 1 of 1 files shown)


Ldoppea (approved):
Looks good to me.
But I cannot reproduce the bug even without the initial hack. Is there a note somewhere on how to reproduce it?

Crash--:
> But I cannot reproduce the bug even without the initial hack

What do you mean by "initial hack"? What's the bug you're trying to reproduce?

Ldoppea:
> > But I cannot reproduce the bug even without the initial hack
> 
> What do you mean by "initial hack"? What's the bug you're trying to reproduce?

My understanding is that `ObfuscateChar` was added to prevent password managers from remembering those fields.
I tried to remove obfuscated chars and yet my password manager still wasn't trying to remember my konnector's fields.

Crash--:
I'll answer you latter @Ldoppea

PR #1324, opened 2021-06-15
feat: Use deep imports

ptbrowne:
This way, we do not import the whole cozy-ui. This is particularly 
useful so that the PdfJsViewer file is not imported and thus, there
is no pdfjs worker file; this pdf worker file would cause problems when
developing locally on Chrome due to security measures.

```
Uncaught DOMException: Failed to construct 'Worker'
```


Crash-- (approved):
Nice

flohhhh (approved):

PR #1325, opened 2021-06-17
fix(doctypes): Enlarge account matching for hidden credit card number

LucsT:
This regexp modification come from a bug in prod when account number is hidden. Often found when account is a credit card.


Crash-- (changes requested):
can you add a test plz ?

flohhhh (approved):

doubleface (approved):

Crash-- (approved):
can you add one test for `************1234`?

Crash--:
ping @LucsT What is the status of this PR?

PR #1326, opened 2021-07-29
Feat/add organization sharing

Ldoppea:
In order to implement folders and ciphers sharing inside of Cozy Pass we need to implement rules that are specific to Bitwarden's organizations

Note: in the Cozy's context, a folder is equivalent to a Bitwarden's organization containing a single collection with the same name. That's why the name "folder" doesn't appears in this new code

What has been done :
- new document type `Organization` has been declared
- translations for this new type has been created (i took every File's translation and copied them)
- made the `ShareByLink` footer optional in ShareModal
- remove the `previewPath` for organizations as they cannot be previewed until the share keys are fully created
- fixed `fetchFilesPaths` to ignore documents that don't have `parentDir` (this is the case for organizations)
- `ShareDialogCozyToCozy` component now switch between `FixedDialog` and `Dialog` based on `showShareByLink` prop value (cf https://github.com/cozy/cozy-libs/pull/1326#issuecomment-893482019 )

This PR is needed for https://github.com/cozy/cozy-pass-web/pull/32


packages/cozy-sharing/src/helpers/files.js
```diff
@@ -5,9 +5,14 @@ export const fetchFilesPaths = async (client, doctype, files) => {
...
-  })
+  const filePaths = files
+    .map(f => parentDirs.data.find(d => d.id === f.dir_id))
+    .filter(parentDir => parentDir !== undefined)
+    .map(parentDir => {
+      const parentDirPath = parentDir.path
+      return parentDirPath === '/'
+        ? `/${parentDir.name}`
+        : `${parentDirPath}/${parentDir.name}`
+    })
```
Ldoppea, line 16: This code is not covered by unit tests, so i'm not sure of side effect I may have introduced.

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
```diff
@@ -88,16 +89,18 @@ const ShareDialogCozyToCozy = ({
...
+            checked={link !== null}
+            document={document}
+            documentType={documentType}
+            link={link}
+            onChangePermissions={onUpdateShareLinkPermissions}
+            onDisable={onRevokeLink}
+            onEnable={onShareByLink}
+            permissions={permissions}
+          />
+        ) : null
```
Ldoppea, line 103: This part is not enough as cozy-ui still display the `actions` container even if empty. I may have to edit cozy-ui later.

packages/cozy-sharing/src/SharingProvider.jsx
```diff
@@ -217,12 +217,15 @@ export class SharingProvider extends Component {
         readOnlyRecipients
       })
 
+    const previewPath =
+      document._id === 'com.bitwarden.organizations' ? undefined : '/preview'
```
Ldoppea, line 221: I would like to extract this `com.bitwarden.organizations` to a const. Ideally in a file accessible from here but also from this file : https://github.com/cozy/cozy-client/pull/990/files#diff-eedaf5c6b0118a4e2a44524e49bf386befb317815ee159d97a2242a578b96724

Does someone know if this place exists? Maybe `cozy-doctypes` ?

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
Crash--: What about not setting `actions` to null? Does UI still display the container if the props is not specified?

packages/cozy-sharing/src/helpers/files.js
Crash--: let's add a few unit tests so :D

packages/cozy-sharing/src/SharingProvider.jsx
Crash--: The future of cozy-doctypes doesn't exist. We want to shutdown this lib. The content should be migrated to cozy-client. We have done some work (in cozyclient/helpers) but still remaining a lot of stuff. 

I think cozy-client is the right place to store this kind of information

Ldoppea: @Crash--  `com.bitwarden.organizations` is also declared in `cozy-stack-client` here : https://github.com/cozy/cozy-client/pull/990/files#diff-eedaf5c6b0118a4e2a44524e49bf386befb317815ee159d97a2242a578b96724R5-R7

Currently:
- `cozy-sharing` imports `cozy-client` but not `cozy-stack-client`
- `cozy-stack-client` doest not import `cozy-client`
- `cozy-client` imports `cozy-stack-client`

Is it ok if I import `cozy-stack-client` package in `cozy-sharing`?
Or do you still prefer to add those doctypes in `cozy-client`? -> I would need to import `cozy-client` in `cozy-stack-client` so.

Ldoppea:
I edited the `ShareDialogCozyToCozy` component to switch between FixedDialog and Dialog based on `showShareByLink` prop value.

This allow to fix only the Title when the `share by link` option is off and prevents having an empty space on the dialog's footer.

Before (fixed empty space displayed on dialog's footer) :
image

After (no fixed empty space displayed on dialog's footer) :
image

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
Ldoppea: Fixed with : https://github.com/cozy/cozy-libs/pull/1326#issuecomment-893482019

packages/cozy-sharing/src/SharingProvider.jsx
```diff
@@ -217,12 +217,15 @@ export class SharingProvider extends Component {
         readOnlyRecipients
       })
 
+    const previewPath =
+      document._type === 'com.bitwarden.organizations' ? undefined : '/preview'
+
```
Crash--, line 222: I think I already commented something different but here is my new mind 🙄 : 

previewPath should be set by the caller. Like that, we can do what we want from the app's land. 

```jsx
 <SharingProvider doctype="io.cozy.files" documentType="Files" previewPath="/preview">
```

packages/cozy-sharing/src/SharingProvider.jsx
Crash--: That's a lot of thinking for not much. And since we'll move this string to the caller, no need to think about it anymore :D

packages/cozy-sharing/src/SharingProvider.jsx
Ldoppea: Seems doable. Maybe we should set a default to "/preview" to maintain retro-compatiblity?

Crash--: I think so, yes!

packages/cozy-sharing/src/SharingProvider.jsx
Ldoppea: https://www.youtube.com/watch?v=18SNR9c09is

packages/cozy-sharing/src/SharingProvider.jsx
Ldoppea: Implemented in https://github.com/cozy/cozy-libs/pull/1326/commits/66bd98e8f26aeac6c645ea604c13bf2ccc91aeaf

I never unit-tested React Providers before, do you know an easy way to test this part?

Crash-- (approved):

Ldoppea:
Tests snapshots have been updated to reflect changes from latest `cozy-ui` version (`51.5.0`) that includes this PR : https://github.com/cozy/cozy-ui/pull/1743

Ldoppea:
This PR is now ready for merging and can be reviewed

Crash-- (approved):

PR #1327, opened 2021-08-02
feat: Harvest ContactPicker not shown if Contacts apps not installed

JF-Cozy:
in EditContract modal

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/Contracts.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/Contracts.jsx
@@ -16,14 +16,12 @@ import NavigationList, {
   NavigationListSection,
   NavigationListHeader
 } from 'cozy-ui/transpiled/react/NavigationList'
-
-import { getAccountLabel } from './bankAccountHelpers'
-import EditContract from './EditContract'
-
-import withLocales from '../../hoc/withLocales'
-
 import WalletIcon from 'cozy-ui/transpiled/react/Icons/Wallet'
 import RightIcon from 'cozy-ui/transpiled/react/Icons/Right'
+
+import withLocales from '../../hoc/withLocales'
+import { getAccountLabel } from './bankAccountHelpers'
+import EditContract from './EditContract'
 
 const makeContractsConn = ({ account }) => {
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/Contracts.jsx
```diff
@@ -95,11 +104,26 @@ const customHeaderPerDoctype = {
 }
 
 const DumbContracts = ({ contracts, account, konnector }) => {
-  const contractData = contracts.data ? contracts.data : contracts
   const { t } = useI18n()
+  const contactsAppQuery = useFetchJSON('GET', '/apps/contacts')
```
Crash--, line 108: Why using useFetchJSON and not an useQuery with this kind of definition? 
```
Q('io.cozy.apps').getById('io.cozy.apps/contacts')`
```

Crash--: Like that, we can make / use the query inside the components without having to pass down a props since the query will not be launched twice

JF-Cozy: > Why using useFetchJSON and not an useQuery with this kind of definition?
> 
> ```
> Q('io.cozy.apps').getById('io.cozy.apps/contacts')`
> ```

Parce que retour de vacances ? :p

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/EditContract.jsx
```diff
@@ -290,7 +308,9 @@ EditContract.propTypes = {
   /** Account id (necessary to toggle contract sync, not present for disconnected accounts) */
   accountId: PropTypes.string,
   /** Konnector that fetched the contract (not present for disconnected accounts) */
-  konnector: PropTypes.object
+  konnector: PropTypes.object,
+  /** Wether the Contacts app is installed */
+  isContactsAppInstalled: PropTypes.bool
```
Crash--, line 313: to remove?

Crash-- (approved):

PR #1340, opened 2021-08-19
Feat/two steps confirmation

Ldoppea:
This PR adds support to `two steps confirmation` mechanisms

This mechanism is an opt-in mechanism and calling app should specify `hasTwoStepsConfirmation:true` in props. So this should not impact existing apps.

This mechanism is needed by `cozy-pass-web` : when Alice shares a folder to Bob, then B can accept this sharing from his emails, then Alice HAS to **confirm** Bob by displaying a fingerprint phrase. This mechanism ensures that Alice and Bob knows each others and that their encryption keys can be shared.

This mechanism has been implemented in a generic way so it can be adapted for future needs.

The only specific component is `CozyPassFingerprintDialogContent` but it should be passed as a prop to ShareDialog to be shown when the user click on `confirm`. So it will be possible to pass another component if needed.

Now when opening the `ShareModal` first thing that is done is to load eligible recipients. This action may not be processed by CozyClient and its realtime mechanisms. In the case of `cozy-pass-web` it is done on Angular side using Bitwarden http client.
This may introduce a latency in `ShareModal` so a loading spinner has been implemented and is called while `getRecipientsToBeConfirmed()` is being called.

This is may also introduce errors as this is a function passed from the exterior (i.e the app). So I implemented an "error" state that show the user that something wrong happened. This message is rudimentary as I don't have graphic specs, but also this should be a rare event.

Then, when fully loaded, user will be able to confirm or reject eligible recipients.

`confirmRecipient` and `rejectRecipient` are passed as props and should be handled by the calling app.

Here is a video of the new behaviour (loading spinner has been force to 3s for this demo but it is nearly instant in real env):

https://user-images.githubusercontent.com/1884255/130109871-4366c878-a1e5-41c5-8549-05e2b2b921f2.mp4

When an error happens then here is the displayed content:
image

Notes :
-  UI is not definitive as @joel-costa still has some work to do on the specs
- Passphrase color is wrong as `cozy-ui` color palettes are not up to date (see: https://github.com/cozy/cozy-ui/issues/1865#issuecomment-894274954)
- Chips backgrounds are grey on hover this should be fixed later when I will finalise the UI
- `reject` Chip is not a perfect circle, I did not investigate yet if this is a bug or just a missing property

### DIFF
--- a/packages/cozy-sharing/locales/fr.json
+++ b/packages/cozy-sharing/locales/fr.json
@@ -1,4 +1,7 @@
 {
+  "loading": {
+    "sharing": "Chargement en cours"
+  },
   "Share": {
     "status": {
       "owner": "Propriétaire",
@@ -39,10 +42,14 @@
       "defaultDisplayName": "Anonyme"
     },
     "twoStepsConfirmation": {
+      "title": "Confirmer le contact",
       "confirm": "Confirmer",
       "reject": "Refuser",
+      "cancel": "Annuler",
       "contactAreWaitingForConfirmation": "%{smart_count} contact en attente de confirmation. Confirmer son identité pour lui permettre d’accéder aux éléments que vous avez partagé avec lui.||||%{smart_count} contacts en attente de confirmation. Confirmer leur identité pour leur permettre d’accéder aux éléments que vous avez partagé avec eux."
     },
+    "loadingError": "Erreur de chargement"
+  },
   "Sharings": {
     "unavailable": {
       "title": "Ravi de vous revoir !",
@@ -151,7 +158,6 @@
       "sharedBy": "Partagé par %{name}",
       "shareByLink": {
         "subtitle": "Partager par lien",
-        "desc": "Chaque personne possédant le lien fourni peut voir et modifier votre note.",
         "create": "Générer un lien",
         "activated": "Lien activé",
         "copy": "Copier le lien",
@@ -232,7 +238,6 @@
       "sharedWithMe": "Partagé avec moi",
       "shareByLink": {
         "subtitle": "Ou partager par lien",
-        "desc": "Chaque personne possédant le lien fourni peut voir et télécharger vos photos.",
         "fetchFailed": "Oups ! Il semblerait que votre connexion internet soit limitée, réessayez plus tard quand le réseau sera meilleur.",
         "create": "Générer un lien",
         "activated": "Lien activé",
--- a/packages/cozy-sharing/locales/fr.json
+++ b/packages/cozy-sharing/locales/fr.json
@@ -360,6 +360,10 @@
         "isInSharedFolder": "se trouve dans un dossier partagé.",
         "hasSharedFolder": "contient un dossier partagé."
       }
+    },
+    "fingerprint": {
+      "instruction": "Chaque Cozy est associé **une phrase d’empreinte** unique. Par sécurité, merci de demander à %{userName} (%{userEmail}) si la phrase d’empreinte de son Cozy correspond bien à :",
+      "instruction2": "La phrase d’empreinte d’un Cozy est accessible **en haut de sa liste de mot de passes dans Cozy Pass**. Cette phrase d'empreinte est également accessible dans **Paramètres/Profil/Phrase d’empreinte** de son client mobile ou de son extension navigateur."
     }
   }
 }
--- a/packages/cozy-sharing/locales/fr.json
+++ b/packages/cozy-sharing/locales/fr.json
@@ -43,10 +43,12 @@
     },
     "twoStepsConfirmation": {
       "title": "Confirmer le contact",
+      "titleReject": "Refuser le contact",
       "confirm": "Confirmer",
       "reject": "Refuser",
       "cancel": "Annuler",
-      "contactAreWaitingForConfirmation": "%{smart_count} contact en attente de confirmation. Confirmer son identité pour lui permettre d’accéder aux éléments que vous avez partagé avec lui.||||%{smart_count} contacts en attente de confirmation. Confirmer leur identité pour leur permettre d’accéder aux éléments que vous avez partagé avec eux."
+      "contactAreWaitingForConfirmation": "%{smart_count} contact en attente de confirmation. Confirmer son identité pour lui permettre d’accéder aux éléments que vous avez partagé avec lui.||||%{smart_count} contacts en attente de confirmation. Confirmer leur identité pour leur permettre d’accéder aux éléments que vous avez partagé avec eux.",
+      "confirmRejection": "Êtes-vous sûr de vouloir refuser le contact %{userName} (%{userEmail})?"
     },
     "loadingError": "Erreur de chargement"
   },
--- a/packages/cozy-sharing/locales/fr.json
+++ b/packages/cozy-sharing/locales/fr.json
@@ -37,8 +37,11 @@
         "success": "L'application a accès à vos contacts"
       },
       "defaultDisplayName": "Anonyme"
-    }
-  },
+    },
+    "twoStepsConfirmation": {
+      "confirm": "Confirmer",
+      "reject": "Refuser"
+    },
   "Sharings": {
     "unavailable": {
       "title": "Ravi de vous revoir !",
--- a/packages/cozy-sharing/locales/fr.json
+++ b/packages/cozy-sharing/locales/fr.json
@@ -40,7 +40,8 @@
     },
     "twoStepsConfirmation": {
       "confirm": "Confirmer",
-      "reject": "Refuser"
+      "reject": "Refuser",
+      "contactAreWaitingForConfirmation": "%{smart_count} contact en attente de confirmation. Confirmer son identité pour lui permettre d’accéder aux éléments que vous avez partagé avec lui.||||%{smart_count} contacts en attente de confirmation. Confirmer leur identité pour leur permettre d’accéder aux éléments que vous avez partagé avec eux."
     },
   "Sharings": {
     "unavailable": {
--- a/packages/cozy-sharing/package.json
+++ b/packages/cozy-sharing/package.json
@@ -32,7 +32,8 @@
     "cozy-doctypes": "^1.82.2",
     "lodash": "^4.17.19",
     "react-autosuggest": "^9.4.3",
-    "react-tooltip": "^3.11.1"
+    "react-tooltip": "^3.11.1",
+    "snarkdown": "^2.0.0"
   },
   "devDependencies": {
     "@babel/core": "7.12.3",
--- a/packages/cozy-sharing/src/components/CozyPassFingerprintDialogContent.jsx
+++ b/packages/cozy-sharing/src/components/CozyPassFingerprintDialogContent.jsx
@@ -1,0 +1,55 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import Paper from 'cozy-ui/transpiled/react/Paper'
+import Typography from 'cozy-ui/transpiled/react/Typography'
+
+import { useI18n } from 'cozy-ui/transpiled/react/I18n'
+import snarkdown from 'snarkdown'
+
+import cx from 'classnames'
+import styles from './CozyPassFingerprintDialogContent.styl'
+
+/**
+ * CozyPassFingerprintDialogContent can be use as a content for ShareModal when
+ * using 2 steps confirmation
+ *
+ * This should be the case when ShareModal is called from `cozy-pass-web`
+ */
+export const CozyPassFingerprintDialogContent = ({
+  recipientConfirmationData
+}) => {
+  const { t } = useI18n()
+
+  const instruction1 = snarkdown(
+    t(`Organizations.fingerprint.instruction`, {
+      userName: recipientConfirmationData.name,
+      userEmail: recipientConfirmationData.email
+    })
+  )
+
+  const instruction2 = snarkdown(t(`Organizations.fingerprint.instruction2`))
+
+  return (
+    <div className={cx(styles['share-modal-content'])}>
+      <p dangerouslySetInnerHTML={{ __html: instruction1 }}></p>
+      <Paper
+        className={cx(styles['cozy-pass-fingerprint-modal__fingerprint'])}
+        elevation={1}
+      >
+        <Typography variant="body1">
+          {recipientConfirmationData.fingerprintPhrase}
+        </Typography>
+      </Paper>
+      <p dangerouslySetInnerHTML={{ __html: instruction2 }}></p>
+    </div>
+  )
+}
+
+CozyPassFingerprintDialogContent.propTypes = {
+  recipientConfirmationData: PropTypes.shape({
+    name: PropTypes.string.isRequired,
+    email: PropTypes.string.isRequired,
+    fingerprintPhrase: PropTypes.string.isRequired
+  })
+}
--- a/packages/cozy-sharing/src/components/Recipient.jsx
+++ b/packages/cozy-sharing/src/components/Recipient.jsx
@@ -16,6 +16,8 @@ import TrashIcon from 'cozy-ui/transpiled/react/Icons/Trash'
 import EyeIcon from 'cozy-ui/transpiled/react/Icons/Eye'
 import PaperplaneIcon from 'cozy-ui/transpiled/react/Icons/Paperplane'
 import ToTheCloudIcon from 'cozy-ui/transpiled/react/Icons/ToTheCloud'
+import CheckCircleIcon from 'cozy-ui/transpiled/react/Icons/CheckCircle'
+import CrossCircleIcon from 'cozy-ui/transpiled/react/Icons/CrossCircle'
 
 import AvatarPlusX from './AvatarPlusX'
 import styles from './recipient.styl'
@@ -26,6 +28,7 @@ import Identity from './Identity'
 import LinkIcon from 'cozy-ui/transpiled/react/Icons/Link'
 
 import Typography from 'cozy-ui/transpiled/react/Typography'
+import Chip from 'cozy-ui/transpiled/react/Chip'
 
 export const MAX_DISPLAYED_RECIPIENTS = 3
 const DEFAULT_DISPLAY_NAME = 'Share.contacts.defaultDisplayName'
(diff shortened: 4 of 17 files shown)


Ldoppea:
`Cannot find module 'cozy-ui/transpiled/react/icons/CheckCircle' from 'Recipient.jsx'` in CI, I suppose this is a yarn cache related problem? 🤔  I'll check that tomorrow.

Crash--:
> Cannot find module 'cozy-ui/transpiled/react/icons/CheckCircle' from 'Recipient.jsx'

Let's replace `icons` by `Icons`. MacOS is not case sensitive. Linux is.

packages/cozy-sharing/src/components/Recipient.jsx
```diff
@@ -295,15 +298,83 @@ const Status = ({ status, isMe, instance }) => {
...
+
+  return (
+    <>
+      <Chip
+        className={cx(styles['recipient-confirm__action--confirm'])}
+        variant="outlined"
+        size="small"
+        onClick={confirm}
+      >
+        <Icon icon={CheckCircleIcon} style={{ marginRight: '0.5rem' }} />
```
Crash--, line 327: prefer the use of classname={'u-mr-half'}

packages/cozy-sharing/src/components/CozyPassFingerprintDialogContent.jsx
```diff
@@ -0,0 +1,55 @@
...
+    })
+  )
+
+  const instruction2 = snarkdown(t(`Organizations.fingerprint.instruction2`))
+
+  return (
+    <div className={cx(styles['share-modal-content'])}>
+      <p dangerouslySetInnerHTML={{ __html: instruction1 }}></p>
+      <Paper
+        className={cx(styles['cozy-pass-fingerprint-modal__fingerprint'])}
```
Crash--, line 37: I think you can use utility class for this kind of stylus. If we can avoid the creation of stylus file, let's do it :)

packages/cozy-sharing/locales/fr.json
```diff
@@ -40,7 +40,9 @@
     },
     "twoStepsConfirmation": {
       "confirm": "Confirmer",
-      "reject": "Refuser"
+      "reject": "Refuser",
+      "oneContactToConfirm": "1 contact en attente de confirmation. Confirmer son identité pour lui permettre d’accéder aux éléments que vous avez partagé avec lui.",
+      "multipleContactsToConfirm": "%{numberOfContacts} contacts en attente de confirmation. Confirmer leur identité pour leur permettre d’accéder aux éléments que vous avez partagé avec eux."
```
Crash--, line 50: t() is pretty good, it's already deal with single / multiple stuff. 

https://github.com/airbnb/polyglot.js#pluralization 

So you can do: 
```jsx
t(`Share.twoStepsConfirmation.contactsToConfirm`, {
      smart_count: recipientsToBeConfirmed.length
    })
... 
```
```json
contactsToConfirm: "%{smart_count} contact en attente de confirmation. Confirmer son identité pour lui permettre d’accéder aux éléments que vous avez partagé avec lui. |||| %{smart_count} contacts en attente de confirmation. Confirmer leur identité pour leur permettre d’accéder aux éléments que vous avez partagé avec eux."

File: packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx [hunk 4]
```diff
@@ -32,76 +159,148 @@ const ShareDialogCozyToCozy = ({
...
   showShareByLink,
   showShareOnlyByLink,
-  showWhoHasAccess
+  showWhoHasAccess,
+  hasTwoStepsConfirmation = false,
+  twoStepsConfirmationMethods = {}
 }) => {
   const { t } = useI18n()
+
+  const [status, setStatus] = React.useState('loading')
```
Crash--, line 168: So by default we're displaying the spinner and then make a re-render to display the modal. It's not so much, but can it be avoided somehow?

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
```diff
@@ -32,76 +159,148 @@ const ShareDialogCozyToCozy = ({
...
+  } = twoStepsConfirmationMethods
+
+  const mountedRef = React.useRef(false)
+  useEffect(() => {
+    mountedRef.current = true
+
+    return () => {
+      mountedRef.current = false
+    }
+  })
```
Crash--, line 187: You do that in order to know if the modal is still opened, right? Wondering if we should not provide a useSafeState hook like https://github.com/tannerlinsley/react-query/blob/79ad5a9ccbce1486c10e7e364d8fc28bef3fa19f/src/devtools/utils.ts#L80 for instance

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
```diff
@@ -32,76 +159,148 @@ const ShareDialogCozyToCozy = ({
...
+
+  const getRecipientsToBeConfirmed = useCallback(() => {
+    if (
+      hasTwoStepsConfirmation &&
+      twoStepsConfirmationMethods?.getRecipientsToBeConfirmed
```
Crash--, line 192: It seems that our babel config doesn't support this stuff by default, no? https://github.com/cozy/cozy-libs/blob/master/packages/babel-preset-cozy-app/index.js 

How it is transpiled ATM?

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
```diff
@@ -32,76 +159,148 @@ const ShareDialogCozyToCozy = ({
...
+    ) {
+      setStatus('loading')
+
+      twoStepsConfirmationMethods
+        .getRecipientsToBeConfirmed(document.id)
+        .then(result => {
```
Crash--, line 198: let's use async / await ;)

packages/cozy-sharing/src/share.styl
```diff
@@ -25,6 +25,9 @@
         padding-top 0
         margin-top  0
 
+.share-modal-loading
+    text-align: center
```
Crash--, line 29: let's use https://docs.cozy.io/cozy-ui/styleguide/section-utilities.html `u-ta-center` ;)

packages/cozy-sharing/src/components/Recipient.spec.jsx
```diff
@@ -95,6 +96,94 @@ describe('Recipient component', () => {
     expect(onRevoke).toBeCalled()
     expect(onRevokeSelf).not.toBeCalled()
   })
+
+  it('should render confirmation actions if recipient is waiting for confirmation', async () => {
```
Merkur39, line 100: Comme il n'y a pas d'`await` dans la fonction, je pense que le `async` n'est pas utile
(Idem pour les autres tests)

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.spec.jsx
```diff
@@ -0,0 +1,382 @@
...
+
+  const setup = props => {
+    return render(
+      <AppLike client={client}>
+        <ShareDialogCozyToCozy {...props} />
+      </AppLike>
+    )
+  }
+
+  it('shoud show sharing dialog directly when hasTwoStepsConfirmation is set to false', async () => {
```
Merkur39, line 23: Idem ici, `async` n'est pas utile.

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
```diff
@@ -32,76 +159,148 @@ const ShareDialogCozyToCozy = ({
   showShareByEmail,
   showShareByLink,
   showShareOnlyByLink,
-  showWhoHasAccess
+  showWhoHasAccess,
+  hasTwoStepsConfirmation = false,
```
Merkur39, line 163: Je me demande si `hasTwoStepsConfirmation` est vraiment essentiel, `twoStepsConfirmationMethods` pourrait suffire non ?

packages/cozy-sharing/src/components/Recipient.spec.jsx
```diff
@@ -3,6 +3,7 @@ import Recipient, { excludeMeAsOwnerFromRecipients } from './Recipient'
 import AppLike from '../../test/AppLike'
 import { createMockClient } from 'cozy-client'
 import { render, fireEvent } from '@testing-library/react'
+import '@testing-library/jest-dom/extend-expect'
```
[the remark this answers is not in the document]
Ldoppea: This is needed to declare `not.toBeInTheDocument()` method : https://github.com/cozy/cozy-libs/pull/1340/files#diff-ca54233711f1c3409867785aa90adb6ae6d61cde2e221b8e5c5ebae4e4c3610fR135

packages/cozy-sharing/src/components/Recipient.spec.jsx
```diff
@@ -95,6 +96,94 @@ describe('Recipient component', () => {
...
+
+    const { getByText } = setup({
+      instance: 'foo.mycozy.cloud',
+      status: 'ready',
+      type: 'two-way',
+      isOwner: false,
+      documentType: 'Organizations',
+      recipientConfirmationData: {
+        name: 'Bob'
+      },
```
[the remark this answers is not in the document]
Ldoppea: Do you mean using the same instance of `recipitientConfirmationData` in both `setup()` and `toBeCalledWith()` calls?
What if the component has a side effect on recipientConfirmationData? The expect would still pass and I think we would missed a bug detection here. 

> Like that if we add informations

> "Are we really confirming by using the name of the user and not its id?"

I don't think this would change in the future in the context of this test.
This object is intentionally declared as `PropTypes.object` because we don't manipulate it in `Recipient`, it is only passed as an argument of `confirmRecipient(recipientConfirmationData)`.
Therefore we don't bother what is inside this object, the only requirement is that it is correctly sent to `confirmRecipient()`.
And that's why the test only declares a name, it is the minimal content to check that the same object is passed as argument.

Do you still think I should use an object reference for the test?

In fact there is another requirement as it should have an `email` member when being manipulated by `WhoAsAccess` (the component that instantiate `Recipient`).

Maybe I should add this proptype to `WhoAsAccess`
```js
recipientsToBeConfirmed: React.PropTypes.arrayOf(React.PropTypes.shape({
     email: React.PropTypes.string.isRequired
}))
```

Also I can change `name: 'Bob'` for `email: 'xxx'` in order to have a test that act more like a documentation.

packages/cozy-sharing/src/components/CozyPassFingerprintDialogContent.jsx
Ldoppea: Do you mean something like `u-mr-half`? I just found documentation on https://docs.cozy.io/cozy-ui/styleguide/section-utilities.html but where are those declared? I don't find any declaration when looking for `u-mr-half` in githug organization nor in material-ui.

Crash--: https://github.com/cozy/cozy-ui/blob/67485ca3a6db28dd4ba64d2c91188c21b00034ce/stylus/utilities/spaces.styl 
https://unpkg.com/browse/cozy-ui@51.5.0/dist/cozy-ui.utils.min.css

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
Ldoppea: Would you prefer if I set `sharing` by default? Or maybe a `blank` state?

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
Ldoppea: You guessed it right.

Maybe this will be an opportunity to initiate the Hook library we mentioned in `Tech Transverse` board.

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
Ldoppea: It seems to be handled correctly, here is the `dist` output : 

```js
    if (hasTwoStepsConfirmation && (twoStepsConfirmationMethods === null || twoStepsConfirmationMethods === void 0 ? void 0 : twoStepsConfirmationMethods.getRecipientsToBeConfirmed)) {
```

Crash--: 🤷‍♂️

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
Crash--: Can't we detect that we have to fetch something? Or 

```
useState(hasTwoStepsConfirmation &&
      twoStepsConfirmationMethods?.getRecipientsToBeConfirmed ? 'loading' : 'loaded') 
      ```

File: packages/cozy-sharing/src/components/CozyPassFingerprintDialogContent.jsx (hunk 2 above)
Comment (Ldoppea) [reply packages/cozy-sharing/src/components/CozyPassFingerprintDialogContent.jsx:37]: Seems like sementic colors are not implemented in utility classes so I cannot replace `color var(--secondaryColor)`. Should I inline this style instead? Or should I keep the stylus?

File: packages/cozy-sharing/src/components/Recipient.jsx (hunk 1 above)
Comment (Ldoppea) [reply packages/cozy-sharing/src/components/Recipient.jsx]: Fixed

File: packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx (hunk 4 above)
Comment (Ldoppea) [reply packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx]: Proposition: https://github.com/cozy/cozy-libs/pull/1340/commits/e6549ee80dfde4a32146129da7b48b0611e1e125

File: packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx (hunk 7 above)
Comment (Ldoppea) [reply packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx]: Fixed

File: packages/cozy-sharing/src/share.styl (hunk 8 above)
Comment (Ldoppea) [reply packages/cozy-sharing/src/share.styl]: Fixed

File: packages/cozy-sharing/src/components/Recipient.spec.jsx (hunk 9 above)
Comment (Ldoppea) [reply packages/cozy-sharing/src/components/Recipient.spec.jsx]: Fixed

File: packages/cozy-sharing/src/components/ShareDialogCozyToCozy.spec.jsx (hunk 10 above)
Comment (Ldoppea) [reply packages/cozy-sharing/src/components/ShareDialogCozyToCozy.spec.jsx]: Fixed

File: packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx (hunk 11 above)
Comment (Ldoppea) [reply packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx:163]: Effectivement c'est un peu redondant, j'avais voulu créer une variable set vraiment explicite et à la base je pensais spécifier les proptypes pour rendre les membres de `twoStepsConfirmationMethods` obligatoires si `hasTwoStepsConfirmation` était à true. Mais ça complexifiait trop la déclaration des proptypes.

Je trouve que ça fait tout de même un bon garde fou. Qu'en penses-tu?

File: packages/cozy-sharing/locales/fr.json (hunk 3 above)
Comment (Ldoppea) [reply packages/cozy-sharing/locales/fr.json]: Fixed, thanks for the tip

File: packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx (hunk 11 above)
Comment (Merkur39) [reply packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx:163]: Ok je comprends sa création, mais du coup elle n'est pas très utile (voir handicapante) en l'état.
Je pense que `twoStepsConfirmationMethods` ce suffit à elle-même :)

File: packages/cozy-sharing/src/components/CozyPassFingerprintDialogContent.jsx (hunk 2 above)
Comment (Crash--) [reply packages/cozy-sharing/src/components/CozyPassFingerprintDialogContent.jsx:37]: This is a good question, but here you should not use `<p>`. You should use the correct variant of `Typography` component and with this component you can use the color props. No?

File: packages/cozy-sharing/package.json [hunk 14]
```diff
@@ -32,7 +32,8 @@
     "cozy-doctypes": "^1.82.2",
     "lodash": "^4.17.19",
     "react-autosuggest": "^9.4.3",
-    "react-tooltip": "^3.11.1"
+    "react-tooltip": "^3.11.1",
+    "snarkdown": "2.0.0"
```
Crash--, line 36: we should not fix the version

packages/cozy-sharing/src/components/Recipient.spec.jsx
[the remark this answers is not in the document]
Crash--: can't we add it to jest's setup file in order to have it everywhere in the project? https://github.com/cozy/cozy-libs/blob/master/packages/cozy-sharing/jestHelpers/setup.js

packages/cozy-sharing/src/components/Recipient.spec.jsx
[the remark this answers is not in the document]
Crash--: > Also I can change name: 'Bob' for email: 'xxx' in order to have a test that act more like a documentation.

It's alway a good thing to have tests acting like documentation and working with the same data as in production.

packages/cozy-sharing/package.json
Ldoppea: Fixed. I copy-pasted that line from `cozy-drive`, should I do the fixe on this project too?

Ldoppea:
Some edits have been made on the `Reject` scenario. Now a confirmation dialog is displayed to give the user a chance to cancel it.

Related commit : https://github.com/cozy/cozy-libs/pull/1340/commits/a73d88239d22b8129cb74ea0063afdb9b72e4b85


https://user-images.githubusercontent.com/1884255/130461645-7d1019ac-95e9-4ea9-8bc1-b40df7037201.mp4

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
Ldoppea: I implemented a first hook in cozy-sharing so I can use it without waiting for the new package to be created.

Related commit : https://github.com/cozy/cozy-libs/pull/1340/commits/86ddb21d7d4ffebb5f322749d225017f6b9a2112

Later I expect to create this package and edit current code to use the new hook from this project.

packages/cozy-sharing/src/components/CozyPassFingerprintDialogContent.jsx
Ldoppea: Not sure understand you, the current code is already based on `Typography` with `body1` variant. The `<p>` is resulting from this.

Or maybe you are talking about the `<p>` containing `instruction1` ? If yes I can convert them. But the `p` targeted by `color var(--secondaryColor)` is the one from the current `Typography` inside of the `Paper`. I just don't know how to reference that semantic variable from the JS code without inlining css.

packages/cozy-sharing/src/components/Recipient.spec.jsx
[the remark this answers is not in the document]
Ldoppea: Fixed, tests are now using email and WhoAsAccess has the correct PropType

packages/cozy-sharing/src/components/Recipient.spec.jsx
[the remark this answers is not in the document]
Ldoppea: Good catch, I did the change

packages/cozy-sharing/src/components/ShareDialogCozyToCozy.jsx
Ldoppea: J'ai supprimé `hasTwoStepsConfirmation` : https://github.com/cozy/cozy-libs/pull/1340/commits/f86bb1599aa5af37717c7d6918188620d87d7a68

packages/cozy-sharing/src/components/Recipient.jsx
```diff
@@ -295,15 +298,83 @@ const Status = ({ status, isMe, instance }) => {
...
+
+  const {
+    instance,
+    isOwner,
+    status,
+    recipientConfirmationData,
+    rejectRecipient,
+    confirmRecipient,
+    ...rest
+  } = props
```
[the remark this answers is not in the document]
Ldoppea: I added them : https://github.com/cozy/cozy-libs/pull/1340/commits/c6b8cb02e00a29224a3a6a8ee9536943fcc259d6

`instance`, `isOwner` and `status` seems to be always provided so I would make them mandatory, but I'm not sure if I checked all scenario. What do you think?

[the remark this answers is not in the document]
Crash--: I don't know 🤷‍♂️

Crash-- (approved):

PR #1342, opened 2021-08-25
feat: Add Sharing Banner as a new component in cozy-sharing

acezard:
Features:

- Take all files needed for this component to work from `cozy-drive`
- Expose `SharingBannerPlugin` component as sole public API
- `SharingBanner` follows loosely a domain-driven design approach, everything that it needs is in a single directory in `cozy-sharing`. This is by design and is inspired by Atlaskit's `editor-core` codebase design.
- Test that SharingBanner works as expected at least in `cozy-notes`
- Try to keep the public API as minimal and simple as possible (currently a success, the `SharingBannerPlugin` does not need props)
- Add some integration/unit tests if deemed valuable
- Localization is internal to the component, the parent application knows nothing about it and doesn't need to pass any parameter
- Display trigger is internal to the component, the parent application must always render `SharingBannerPlugin` where it is needed

Caveats:

- It is very important that `cozy-ui` and `cozy-client` dependencies are not duplicated in `cozy-notes` and `cozy-sharing`. Otherwise, multiple bugs will crash the application
- We need Transifex to work in `cozy-sharing` for French translations. Is that possible?


packages/cozy-sharing/src/SharingBanner/hooks/useShouldDisplay.js
```diff
@@ -0,0 +1,13 @@
+import { useState, useEffect } from 'react'
```
acezard, line 1: This is a brand new file (did not exist in cozy-drive). 

We authorize the banner display by naively looking at the window location. If the location gets a match in an array of valid locations, we just return `true`.

packages/cozy-sharing/src/SharingBanner/index.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
```
acezard, line 1: This is a brand new file (did not exist in cozy-drive).

Not really an HoC but basically has the same use. It supercharges the SharingBanner with display trigger and localization. We used a simple component here instead of an HoC because it was simple enough to solve the problem at hand. Moreover, it is not supposed to be reused.

packages/cozy-sharing/src/SharingBanner/locales/en.json
```diff
@@ -0,0 +1,57 @@
+{
```
acezard, line 1: We extracted the localization from cozy-drive. Problem is that we do not have the French translation which is managed by Transifex in cozy-drive. Should we use Transifex here, or should we dump a fr.json?

packages/cozy-sharing/src/components/ShareByEmail.jsx
```diff
@@ -126,7 +126,9 @@ export class ShareByEmail extends Component {
       })
 
       Alerter.success(
-        ...getSuccessMessage(recipientsBefore, contacts, documentType)
+        this.props.t(
```
acezard, line 129: This never worked properly but cozy-drive used to display its own translation so the bug never was detected.

packages/cozy-sharing/src/index.jsx
```diff
@@ -16,3 +16,4 @@ export { ShareButton } from './ShareButton'
 export { ShareModal } from './ShareModal'
 export { RefreshableSharings } from './RefreshableSharings'
 export { useFetchDocumentPath } from './components/useFetchDocumentPath'
+export { SharingBannerPlugin } from './SharingBanner'
```
acezard, line 19: This is the only API we're giving to users of this plugin. Moreover as seen above, this component doesn't accept any props. This is intended.

Big caveat though, this plugin will not work if cozy-client or cozy-ui are duplicated at runtime (different versions for instance)

packages/cozy-sharing/src/SharingBanner/components/PublicBanner.jsx
```diff
@@ -0,0 +1,166 @@
...
+        action
+      }
+    : {
+        label: t('Share.banner.add_to_mine'),
+        icon: 'to-the-cloud',
+        action
+      }
+  return (
+    <Banner
+      bgcolor={palette['paleGrey']}
```
Crash--, line 88: should use semantic var

packages/cozy-sharing/src/SharingBanner/components/PublicBanner.jsx
```diff
@@ -0,0 +1,166 @@
...
+      href="https://cozy.io"
+      target="_blank"
+      className={'u-link'}
+      rel="noopener noreferrer"
+    >
+      {t('Share.banner.know_more')}
+    </a>
+  )
+  return (
+    <span className={'u-charcoalGrey'}>
```
Crash--, line 129: semantic var

packages/cozy-sharing/src/SharingBanner/components/PublicBanner.jsx
```diff
@@ -0,0 +1,166 @@
...
+      {text} {knowMore}
+    </span>
+  )
+}
+
+const SharingBannerByLink = ({ onClose }) => {
+  const { t } = useI18n()
+  return (
+    <Banner
+      bgcolor={palette['paleGrey']}
```
Crash--, line 139: semantic

packages/cozy-sharing/src/SharingBanner/components/getHomeLinkHref.jsx
```diff
@@ -0,0 +1,6 @@
+const getHomeLinkHref = from =>
+  `https://manager.cozycloud.cc/cozy/create${
+    from ? `?pk_campaign=${encodeURIComponent(from)}` : '?'
+  }&pk_kwd=cozy`
```
Crash--, line 4: @poupotte Est-ce qu'on garde encore un "tracking" particulier pour la création depuis le partage ? Est-ce que tu veux qu'on rajoute l'application ? Là par défaut on va avoir `sharing-drive` on veut peut-être `sharing-$slug`?

packages/cozy-sharing/src/SharingBanner/components/publicBanner.styl
```diff
@@ -0,0 +1,15 @@
+@require 'components/avatar'
+@require 'components/circle'
```
Crash--, line 2: It's clearly what we avoid to do. We should not require components/* anymore. We should use React component from UI instead. I know the code already exists, but let's fix this.

packages/cozy-sharing/src/SharingBanner/hooks/useShouldDisplay.js
```diff
@@ -0,0 +1,13 @@
...
+
+const sharingPathnames = ['/preview']
+
+export const useShouldDisplay = location => {
+  const [shouldDisplay, setDisplay] = useState(false)
+
+  useEffect(() => {
+    setDisplay(sharingPathnames.includes(location.pathname))
+  }, [location])
+
```
Crash--, line 11: This should not be part of this lib. The lib is not the right place to know if we have to display the banner or not. It should be done in the app's land, no?

Crash--: Having harcoded path smells not good

packages/cozy-sharing/src/SharingBanner/locales/en.json
Crash--: No. You should put this content in the already existing files https://github.com/cozy/cozy-libs/tree/master/packages/cozy-sharing/locales

packages/cozy-sharing/src/components/ShareByEmail.jsx
Crash--: It works correctly! `Alerter` can accept a `t` props https://github.com/cozy/cozy-ui/blob/master/react/Alerter/index.jsx#L113 

It was done in Drive https://github.com/cozy/cozy-drive/blob/master/src/drive/web/modules/layout/Layout.jsx#L30

So there is no bug. 

If translation is now handled in cozy-sharing, we'll have an error in the console saying that the key will not exist but will be correctly translated.  

BUT the component should not be linked to translate() anymore but to withLocal

packages/cozy-sharing/src/SharingBanner/hooks/useShouldDisplay.js
acezard: Well, my idea was to plug and play the Sharing Banner so yeah it can be handled in the applications themselves or in the banner. The advantage in the banner means that only one place (in terms of code) handles it instead of multiple apps. The caveat is the level of abstraction can become a black box for the consuming app (cozy notes has no idea when or why the banner will show up)

packages/cozy-sharing/src/SharingBanner/index.jsx
```diff
@@ -0,0 +1,20 @@
...
+import { useShouldDisplay } from './hooks/useShouldDisplay'
+import { SharingBanner } from './components/SharingBanner'
+
+export const SharingBannerPlugin = () => {
+  const sharingInfos = useSharingInfos()
+  const { lang } = useI18n()
+  const shouldDisplay = useShouldDisplay(window.location)
+
+  return shouldDisplay ? (
+    <I18n dictRequire={() => require('./locales/en.json')} lang={lang}>
```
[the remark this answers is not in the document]
acezard: there is still the call to useSharingInfos(), from what I saw there is no gain in calling it from cozy-notes for instance (it's an argument-free function)

packages/cozy-sharing/src/SharingBanner/locales/en.json
acezard: Yes, I hesitated but decided to follow the "everything in one directory". I guess it goes too far when it's about localization

packages/cozy-sharing/src/components/ShareByEmail.jsx
acezard: there must be some problem elsewhere then, in cozy-notes the message was never translated despite receiving the correct parameters (email, locale key)

packages/cozy-sharing/src/SharingBanner/components/publicBanner.styl
acezard: Agreed

packages/cozy-sharing/src/SharingBanner/hooks/useShouldDisplay.js
Crash--: We should not have any black box stuff. Only declarative, so we should not have this kind of behavior harcoded in the lib. Also, /preview is very specific to drive & notes. For instance Pass doesn't have this route.

packages/cozy-sharing/src/SharingBanner/index.jsx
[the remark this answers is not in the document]
Crash--: good catch

packages/cozy-sharing/src/components/ShareByEmail.jsx
Crash--: In Notes, there is no `t` passed to Alerter https://github.com/cozy/cozy-notes/blob/81081c2e5f9c3401e39daba4cf76b68fab615b1b/src/components/app.jsx#L119

packages/cozy-sharing/src/SharingBanner/components/publicBanner.styl
acezard: I tried to use components instead and it's a bit tricky, it's a markdown string compiled to HTML string by an external lib. I'm guessing that's why it was implemented with CSS import instead of component composition. Not sure how to handle this unless by getting rid of the markdown library

packages/cozy-sharing/src/components/__snapshots__/ContactSuggestion.spec.jsx.snap
```diff
@@ -290,7 +290,7 @@ exports[`ContactSuggestion component should display contact suggestion for a gro
         style={
           Object {
             "--circleSize": "32px",
-            "backgroundColor": "#7F6BEE",
+            "backgroundColor": "var(--purpley)",
```
acezard, line 293: Lots of non-semantic colors in those test components (it is their only use here).

packages/cozy-sharing/src/SharingBanner/components/PublicBanner.jsx
```diff
@@ -0,0 +1,165 @@
...
+  return (
+    <Banner
+      bgcolor={'var(--defaultBackgroundColor)'}
+      text={<SharingBannerByLinkText />}
+      buttonOne={
+        <ButtonLink
+          theme="text"
+          label={t('Share.create-cozy')}
+          icon={CozyHomeLinkIcon}
+          href={getHomeLinkHref('sharing-drive')}
```
Crash--, line 145: We should remove `-drive`. Since it'll be used for Drive and Notes now.

Crash-- (approved):

packages/cozy-sharing/src/SharingBanner/components/PublicBanner.jsx
acezard: Done

PR #1349, opened 2021-09-06
chore: Let's exit 1 if there is changes after lint

Crash--:
We previously had issue only when deploying not merging since lerna makes a git diff before deploying. Let's do the same with the CI before merging.

### DIFF
--- a/scripts/travis.sh
+++ b/scripts/travis.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+yarn build
+yarn lint
+yarn lint:md -f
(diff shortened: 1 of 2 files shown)


scripts/travis.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+yarn build
+yarn lint
+yarn lint:md -f
+yarn test
+
+
```
acezard, line 10:
```suggestion

```

acezard (approved):

PR #1353, opened 2021-09-13
feat: Improve two steps confirmation user experience

Ldoppea:
~~⚠️  this PR cannot be merged yet as it depends on https://github.com/cozy/cozy-ui/pull/1905~~

Instead of displaying `confirm` and `reject` buttons, now a `verify`
button is displayed

The opened dialog now allows to accept or reject the user

image
image

### DIFF
--- a/packages/cozy-sharing/locales/en.json
+++ b/packages/cozy-sharing/locales/en.json
@@ -48,6 +48,7 @@
       "confirm": "Confirm",
       "reject": "Reject",
       "cancel": "Cancel",
+      "verify": "Verify",
       "contactAreWaitingForConfirmation": "Some contacts are waiting for your confirmation. Confirm their identities to allow them to access items you shared with them.||||%{smart_count} contacts are waiting for your confirmation. Confirm their identities to allow them to access items you shared with them.",
       "confirmRejection": "Do you really want to reject contact %{userName} (%{userEmail})?"
     },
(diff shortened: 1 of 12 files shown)


Ldoppea:
Note that this test has been disabled as I did not find a way to test the new scenario due to multiple consecutive dialogs and usage of asynchronism. I will investigate how to make it work again

https://github.com/cozy/cozy-libs/pull/1353/files#diff-f49230d3c62af918f82a59ecf554bd230ce4e2a1e951d66f312f305e7710317dR146

packages/cozy-sharing/locales/en.json
```diff
@@ -48,6 +48,7 @@
       "confirm": "Confirm",
       "reject": "Reject",
       "cancel": "Cancel",
```
Crash--, line 50: do reject and confirm still needed?

Crash-- (approved):

packages/cozy-sharing/locales/en.json
Ldoppea: yes, they are still need for the dialogs buttons

https://github.com/cozy/cozy-libs/pull/1353/files#diff-51e44a6c84e55e3a0b74af33011956c42b4c915f06f59a2fc39b6a928d6c167eL122-L133

Crash-- (approved):

PR #1359, opened 2021-09-27
fix: Displaying error message

Merkur39:
Error message should not be displayed if the konnector is on maintenance


Crash--:
can you add a test please?

Crash-- (approved):

PR #1365, opened 2021-09-28
feat: Handle Typescript files in eslint and babel-preset

acezard:
Add an override section that will handle TS files with specific rules

This required an update for eslint. Files were linted due to some changes after that update. Commit history will get squashed before merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '10'
+  - '12'
 env:
   global:
     - NODE_OPTIONS="--max-old-space-size=8192"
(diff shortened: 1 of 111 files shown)


acezard:
@Crash-- I think the build will fail until eslint and babel config are published and consumed by other cozy libs... How should we handle that?

acezard:
> waiting the return of @acezard to talk with him about this issue

@Crash--  I'm available

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '10'
+  - '16'
```
Crash--, line 3: node 12 since our production is on node 12

Crash-- (approved):

PR #1366, opened 2021-09-29
fix: Removed line break on lastSync label

Merkur39:


packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx
```diff
@@ -34,7 +34,12 @@ export const DumbLaunchTriggerCard = ({ flow, className, f, t, disabled }) => {
             >
               {t('card.launchTrigger.lastSync.label')}
             </Typography>
-            <Typography variant="caption" color="textPrimary" component="span">
+            <Typography
+              variant="caption"
+              color="textPrimary"
+              component="span"
+              style={inlineStyle}
```
Ldoppea, line 41: nit: this is homogenous with other Typography from this file, but there seems to be a "MUI" way to do this : 
```jsx
            <Typography
              variant="caption"
              color="textPrimary"
              component="span"
              display="inline-block">
```
https://webcache.googleusercontent.com/search?q=cache:pa4T3YYOgwUJ:https://v4-3-3.material-ui.com/ru/system/display/+&cd=2&hl=fr&ct=clnk&gl=fr

Ldoppea (approved):

PR #1367, opened 2021-09-29
fix: Display Recipient's initials if no avatarPath or status is provided

Ldoppea:
`<Recipient>` may now be used in not sharing related contexts

This is the case on cozy-pass where the user can display a list of
users that are waiting for identity confirmation (this uses
`ConfirmTrustedRecipientsDialog`)

In this scenario, there is no `avatarPath` to provide as the stack only
supports avatars on the
`/sharings/${sharing.id}/recipients/${recipient.id}/avatar` route

So we want to set `null` image in props instead of an url formatted
with missing attributes. By setting `null` image, then the `Avatar`
component will display the text attribute instead

### DIFF
--- a/packages/cozy-sharing/src/components/Recipient.jsx
+++ b/packages/cozy-sharing/src/components/Recipient.jsx
@@ -126,18 +126,24 @@ export const RecipientsAvatars = ({
 
 export const RecipientAvatar = ({ recipient, ...rest }) => {
   const client = useClient()
+
+  /**
+   * avatarPath is always the same for a recipient, but image
+   * can be different since the stack generate it on the fly.
+   * It can be "gray" during the first load depending of the
+   * status' sharing, but can become active (from the realtime)
+   * so we need a way to "refresh" the image. Passing the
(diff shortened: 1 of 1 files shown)


Crash--:
> <Recipient> may now be used in not sharing related contexts

If this is not a `recipient` of a sharing, why using this component and not `Avatar` directly?

Ldoppea:
In the case of CozyPass, those are not "Recipient of a sharing" but "Recipients of one or more sharings".

On every sharing we have the classic recipients view that shows recipients status for that specific sharing.
But when the user open the app, we display a dialog that lists all recipients waiting for a 2nd confirmation step.

So we list all recipients of all sharings in a single dialog (and those are deduplicated). So we can randomly set the avatar from one of available sharings, or just use initials.

image

(saying that I realize that the status is wrong on my screen)

Note: in the future we may implement the concept of "trusted contacts" that will displays the same kind of information (avatar, name, status, instance) but here there will be no linked sharing at all.

Crash-- (approved):

PR #1388, opened 2021-11-18
fix: Informations on readme

Merkur39:
- Import of FlagSwitcher:
  We could re-export from the index, to match the current doc, but that would cause a Breaking Change.
  I don't know if it's really better

- Condition of flag('switcher'):
  The condition does not return `undefined` but rather `null`
  (see: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-flags/src/store.js#L35)

### DIFF
--- a/packages/cozy-flags/README.md
+++ b/packages/cozy-flags/README.md
@@ -45,10 +45,11 @@ The `FlagSwitcher` shows all flags in use and displays all
 flags in use.
 
 ```js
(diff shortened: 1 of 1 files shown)


Crash--:
> We could re-export from the index, to match the current doc, but that would cause a Breaking Change.

I think we don't want to. Since this lib is also used in a node environnement, we don't want any React related stuff

Crash-- (approved):

PR #1389, opened 2021-11-22
feat: Implement cozy webview intent

acezard:
# Cozy-intent

## Concept

This library allows a react-native and a react app running in a webview to communicate in unidirectional and/or bidirectional fashions. The library is following declarative paradigms and is fully strictly typed. An emphasis has been put on low configuration needs and ease of use.

In a nutshell, one end says what it wants to an other end, and the other end handles it any way it wants. For more information on the underlying systems of this library, see https://github.com/alesgenova/post-me.

## Installation

The library needs to be installed both in the parent and the children application.

- `yarn add cozy-webview-intent`

### Nomenclature

- `event`: simple function that sends an event to the listening messenger.
- `intent`: high-level abstraction intended to describe the whole flow of events/methods and the actual result of those calls in the application.
- `messenger`: class implementing `post-me` messenger paradigm. Messengers are environment bound (DOM, React-native, worker, iframe, etc). They send and listen methods and events to other connected messengers.
- `method`: messenger function that accept simple arguments to be handled by the listening messenger.
- `service`: Cozy specific implementation designed to provide decoupling from post-me and easier instanciation/interoperability with React applications.

### Parent

First, we need to provide a React context to our application. This requires no configuration at this moment.

```tsx
import React from 'react'

import { NativeIntentProvider } from 'cozy-webview'
import { ReactNativeApp } from 'react-native-app'

const MyNativeApp = () => (
  <NativeIntentProvider>
    <ReactNativeApp />
  </NativeIntentProvider>
)
```

Later, in a webview, we need to register a messenger and give it access to some methods. We need a ref object to the webview first.

```tsx
import React, { useEffect, useState } from 'react'
import { WebView } from 'react-native-webview'

import { useNativeIntent } from 'cozy-webview'

const MyWebview = () => {
  const [ref, setRef] = useState('')
  const nativeIntent = useNativeIntent()

  useEffect(() => {
    if (ref) {
      nativeIntent.registerWebview(ref, {
        ping: () => console.log('pong'),
      })
    }
  }, [ref, nativeIntent])

  return <Webview ref={(ref) => setRef(ref)} onMessage={nativeIntent.tryEmit} />
}
```

### Children

First, we need to provide a React context to our application. This requires a reference to the local window.

```tsx
import React from 'react'

import { WebviewIntentProvider } from 'cozy-webview'
import { ReactApp } from 'react-app'

const MyWebApp = () => (
  <WebviewIntentProvider windowRef={window}>
    <ReactApp />
  </WebviewIntentProvider>
)
```

Later on, we can use the service in a component and call a registered method. Here we call the `ping()` method injected in the native messenger.

```tsx
import React from 'react'

import { useWebviewIntent } from 'cozy-webview'

const MyComponent = () => {
  const webviewIntent = useWebviewIntent()
  const handleClick = () => webviewIntent.call('ping')

  return <buttton onClick={handleClick} />
}
```

##### `NativeService.call('logout')`

Will logout the user from the react-native application. Will not touch webview state.

##### `NativeService.call('openWebview', href)`

Will open a new webview in a Native app with the href provided.

#### Events

N/A

#### Methods

N/A

#### Events

N/A

### DIFF
--- a/packages/cozy-intent/package.json
+++ b/packages/cozy-intent/package.json
@@ -1,0 +1,51 @@
+{
+  "name": "cozy-intent",
+  "description": "Event-based library allowing interactions between React-Native and React applications",
+  "version": "1.0.0",
+  "author": {
+    "name": "Cozy Cloud",
+    "url": "https://github.com/cozy"
+  },
+  "bugs": {
+    "url": "https://github.com/cozy/cozy-libs/issues"
+  },
+  "dependencies": {
+    "cozy-device-helper": "1.15.0",
+    "post-me": "0.4.5",
+    "typescript": "^4.5.2"
+  },
+  "devDependencies": {
+    "@babel/core": "7.12.3",
+    "babel-preset-cozy-app": "2.0.1",
+    "eslint-config-cozy-app": "3.0.1",
+    "react": "17.0.2"
+  },
+  "files": [
+    "dist"
+  ],
+  "homepage": "https://github.com/cozy/cozy-libs/packages/cozy-intent",
+  "keywords": [
+    "cozy",
+    "cozy-libs",
+    "cozy-react-native"
+  ],
+  "license": "MIT",
+  "main": "./dist/index.js",
+  "peerDependencies": {
+    "react": ">=16"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/cozy/cozy-libs.git",
+    "directory": "packages/cozy-intent"
+  },
+  "scripts": {
+    "build": "rm -rf ./dist && yarn build:types && babel --extensions .ts,.tsx ./src -d ./dist --copy-files",
+    "build:types": "tsc -p tsconfig.json",
+    "build:watch": "yarn build --watch",
+    "lint": "eslint . --ext .ts,.tsx",
+    "lint:fix": "yarn lint --fix",
+    "prepublishOnly": "yarn build"
+  },
+  "types": "./dist/index.d.ts"
+}
--- a/packages/cozy-intent/readme.md
+++ b/packages/cozy-intent/readme.md
@@ -1,0 +1,122 @@
+# Cozy-intent
+
+## Concept
+
+This library allows a react-native and a react app running in a webview to communicate in unidirectional and/or bidirectional fashions. The library is following declarative paradigms and is fully strictly typed. An emphasis has been put on low configuration needs and ease of use.
+
+In a nutshell, one end says what it wants to an other end, and the other end handles it any way it wants. For more information on the underlying systems of this library, see https://github.com/alesgenova/post-me.
+
+## Installation
+
+The library needs to be installed both in the parent and the children application.
+
+- `yarn add cozy-webview-intent`
+
+## API
+
+### Nomenclature
+
+- `event`: function that sends an event to the listening messenger.
+- `intent`: high-level abstraction intended to describe the whole flow of events/methods and the actual result of those calls in the application.
+- `messenger`: class implementing `post-me` messenger paradigm. Messengers are environment bound (DOM, React-native, worker, iframe, etc). They send and listen methods and events to other connected messengers.
+- `method`: messenger function that accept arguments to be handled by the listening messenger.
+- `service`: Cozy specific implementation designed to provide decoupling from `post-me` and easier instanciation/interoperability with React applications.
+
+### Parent
+
+First, we need to provide a React context to our application. This requires no configuration at this moment.
+
+```tsx
+import React from 'react'
+
+import { NativeIntentProvider } from 'cozy-webview'
+import { ReactNativeApp } from 'react-native-app'
+
+const MyNativeApp = () => (
+  <NativeIntentProvider>
+    <ReactNativeApp />
+  </NativeIntentProvider>
+)
+```
+
+Later, in a webview, we need to register a messenger and give it access to some methods. We need a ref object to the webview first.
+
+```tsx
+import React, { useEffect, useState } from 'react'
+import { WebView } from 'react-native-webview'
+
+import { useNativeIntent } from 'cozy-webview'
+
+const MyWebview = () => {
+  const [ref, setRef] = useState('')
+  const nativeIntent = useNativeIntent()
+
+  useEffect(() => {
+    if (ref) {
+      nativeIntent.registerWebview(ref, {
+        ping: () => console.log('pong'),
+      })
+    }
+  }, [ref, nativeIntent])
+
+  return <Webview ref={(ref) => setRef(ref)} onMessage={nativeIntent.tryEmit} />
+}
+```
+
+### Children
+
+First, we need to provide a React context to our application. This requires a reference to the local window.
+
+```tsx
+import React from 'react'
+
+import { WebviewIntentProvider } from 'cozy-webview'
+import { ReactApp } from 'react-app'
+
+const MyWebApp = () => (
+  <WebviewIntentProvider windowRef={window}>
+    <ReactApp />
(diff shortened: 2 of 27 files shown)


packages/cozy-webview/src/api/services/WebviewService.ts
```diff
@@ -0,0 +1,24 @@
...
+    void this.init()
+  }
+
+  public init = async (): Promise<void> => {
+    this.connection = await ChildHandshake(this.messenger)
+    this.remoteHandle = this.connection.remoteHandle()
+  }
+
+  public call = (methodName: keyof NativeMethods): Promise<void> | undefined =>
+    this.remoteHandle?.call(methodName)
```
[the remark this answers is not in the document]
Ldoppea: Also I'm wondering if remoteHandle can be null as init is called from the constructor?

packages/cozy-webview/src/api/services/WebviewService.ts
```diff
@@ -0,0 +1,24 @@
...
+  private connection?: Connection
+  private messenger: WebviewMessenger
+  private remoteHandle?: RemoteHandle<NativeMethods>
+
+  constructor(windowRef: WebviewWindow, messengerService = WebviewMessenger) {
+    this.messenger = new messengerService(windowRef)
+    void this.init()
+  }
+
+  public init = async (): Promise<void> => {
```
Ldoppea, line 17: Shouldn't init be private or at least protected? Are we supposed to call it from outside?

packages/cozy-webview/src/api/services/NativeService.ts
```diff
@@ -0,0 +1,31 @@
...
+    localMethods?: NativeMethods
+  ): Promise<void> => {
+    const messenger = new this.messengerService(webviewRef)
+
+    this.messengerRegister = {
+      ...this.messengerRegister,
+      [webviewRef.props.source.uri]: { messenger }
+    }
+
+    await ParentHandshake(messenger, localMethods, 50, 100)
```
Ldoppea, line 25: Can we extract those numbers into named constants so the reader can understand what they mean without having to read the post-me doc?

Ldoppea (approved):
Looks good to me, only a few question (see comments) are remaining but those are not blocking the approval (except maybe the one about `call` returning `promise` or `undefined`)

Regarding the doc (which is great, I like it), I'm still a bit confused about the relation between Services and Messenger, and which one is responsible of what. Maybe we can add a clarification on the doc.

Also do you think we can add a nomenclature to defines different used terms (messenger, message, service, intent, etc)?
Here is an example of what I did on Cozy client : https://github.com/cozy/cozy-client/blob/849c5f7c772dc3b55c8c7c3ed84fb33ed74cfb0e/packages/cozy-client/src/flagship-certification/README.md#nomenclature
This is not a Cozy normalized practice so feel free to do it or not regarding your preference.

acezard:
> Looks good to me, only a few question (see comments) are remaining but those are not blocking the approval (except maybe the one about `call` returning `promise` or `undefined`)
> 
> Regarding the doc (which is great, I like it), I'm still a bit confused about the relation between Services and Messenger, and which one is responsible of what. Maybe we can add a clarification on the doc.
> 
> Also do you think we can add a nomenclature to defines different used terms (messenger, message, service, intent, etc)? Here is an example of what I did on Cozy client : https://github.com/cozy/cozy-client/blob/849c5f7c772dc3b55c8c7c3ed84fb33ed74cfb0e/packages/cozy-client/src/flagship-certification/README.md#nomenclature This is not a Cozy normalized practice so feel free to do it or not regarding your preference.

will updated readme based on your comment

packages/cozy-intent/readme.md
```diff
@@ -0,0 +1,122 @@
...
+
+The library needs to be installed both in the parent and the children application.
+
+- `yarn add cozy-webview-intent`
+
+## API
+
+### Nomenclature
+
+- `event`: simple function that sends an event to the listening messenger.
```
Ldoppea, line 19: `simple` may have multiple meanings and (_regarding which meaning it is intended to have_) it may not give very useful information. Also we tend to avoid formulations that suggest things are easy/simple.

So I propose to remove this word
```suggestion
- `event`: function that sends an event to the listening messenger.
```

packages/cozy-intent/readme.md
```diff
@@ -0,0 +1,122 @@
...
+- `intent`: high-level abstraction intended to describe the whole flow of events/methods and the actual result of those calls in the application.
+- `messenger`: class implementing `post-me` messenger paradigm. Messengers are environment bound (DOM, React-native, worker, iframe, etc). They send and listen methods and events to other connected messengers.
+- `method`: messenger function that accept simple arguments to be handled by the listening messenger.
```
Ldoppea, line 22: Same comment here. If you want to refer to the type of arg (i.e non complex objects), maybe you can use `primitive` instead. But there seems not to be the case?
```suggestion
- `method`: messenger function that accept arguments to be handled by the listening messenger.
```

packages/cozy-intent/readme.md
```diff
@@ -0,0 +1,122 @@
...
+- `service`: Cozy specific implementation designed to provide decoupling from post-me and easier instanciation/interoperability with React applications.
```
Ldoppea, line 23:
```suggestion
- `service`: Cozy specific implementation designed to provide decoupling from `post-me` and easier instanciation/interoperability with React applications.
```

packages/cozy-intent/readme.md
Ldoppea, line 22: I'm wondering if we can find a "less-generic" name for those `methods` to prevent ambiguity with "language" `methods`? Maybe something like `target methods` or `registered methods` (this one is used on your doc)

Ldoppea (approved):
It is way easier to understand those concepts with the latest doc edits 👍 

Looks good to me now

I added some suggestions but those are kind of `nit` comments.

packages/cozy-intent/readme.md
acezard: I agree, I will also mention that the event is just a string that has to be handled in whatever way on the other side

packages/cozy-intent/readme.md
acezard: Good point because I'm not even sure myself what arguments are valid (obviously string-like is ok, like boolean, numbers.. but objects? functions? I need to test, post-me is not clear on that)

acezard: yeah, I used post-me naming convention, which uses events and methods. But the more I think about it, the less it makes sense. You can have methods with no arguments, and that's functionally the same as sending an event. Maybe we should abstract completely those post-me internals and use a single way of communication. I see your suggestions but I think it should be a single word.

doubleface (approved):

packages/cozy-intent/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "homepage": "https://github.com/cozy/cozy-libs/packages/cozy-intent",
+  "keywords": [
+    "cozy",
+    "cozy-libs",
+    "cozy-react-native"
+  ],
+  "license": "MIT",
+  "main": "./dist/index.js",
+  "peerDependencies": {
+    "react": "^17.0.2"
```
Crash--, line 35: We're not in 17 for our webapps. 16 || 17 ?

packages/eslint-config-cozy-app/tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "compilerOptions": {
     "allowJs": true,
     "allowSyntheticDefaultImports": true,
-    "baseUrl": "./src",
```
Crash--, line 5: why this change?

PR #1392, opened 2021-12-14
feat: Use configured support address in UNKNOWN_ERROR message

doubleface:
This will display the expected support mail in the message according to
the context.

I had to convert the TriggerErrorInfo to a Component class since it now
needs asynchronous information from io.cozy.settings.

image

### DIFF
--- a/packages/cozy-harvest-lib/src/components/infos/TriggerErrorInfo.jsx
+++ b/packages/cozy-harvest-lib/src/components/infos/TriggerErrorInfo.jsx
@@ -4,8 +4,9 @@ import PropTypes from 'prop-types'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
 import Infos from 'cozy-ui/transpiled/react/Infos'
 import Typography from 'cozy-ui/transpiled/react/Typography'
+import { withClient } from 'cozy-client'
 
-import { getErrorLocale } from '../../helpers/konnectors'
+import { getErrorLocale, fetchSupportMail } from '../../helpers/konnectors'
 import withKonnectorLocales from '../hoc/withKonnectorLocales'
 import Markdown from '../Markdown'
 
@@ -18,8 +19,20 @@ import Markdown from '../Markdown'
  * deals mainly with translation concerns.
  */
 export class TriggerErrorInfo extends PureComponent {
+  state = {
+    supportMail: null
+  }
+  async componentDidMount() {
+    await this.loadSupportMail()
+  }
+  async loadSupportMail() {
+    const { client } = this.props
+    const supportMail = await fetchSupportMail(client)
+    this.setState({ supportMail })
+  }
   render() {
     const { className, error, konnector, t, action } = this.props
+    const { supportMail } = this.state
     return (
       <Infos
         className={className}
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/src/components/infos/TriggerErrorInfo.jsx
```diff
@@ -17,9 +18,21 @@ import Markdown from '../Markdown'
  * This component is strongly related with the content of locales files and
  * deals mainly with translation concerns.
  */
-export class TriggerErrorInfo extends PureComponent {
+export class TriggerErrorInfo extends Component {
```
Crash--, line 21: PureComponent has nothing to do with async stuff or else. It's the same thing than a component instead that it has a different logique on the shouldComponentUpdate. The component will not re-render itself if its props do not change. Thats it. 

Here I think we can keep the PureComponent, you can take a look for instance here https://github.com/cozy/orange-customizations/blob/61273d3c6e6498f55076afc72a9275b393423720/overrides/cozy-store/src/ducks/components/App.jsx

packages/cozy-harvest-lib/src/locales/en.json
```diff
@@ -155,7 +155,7 @@
       },
       "UNKNOWN_ERROR": {
         "title": "Connection error",
-        "description": "An unknown error has occurred. You can try to update your data. If the problem persists, please contact us at [claude@cozycloud.cc](mailto:claude@cozycloud.cc)."
+        "description": "An unknown error has occurred. You can try to update your data. If the problem persists, please contact us at [%{supportMail}](mailto:%{supportMail})."
```
Crash--, line 158: just before this sentence we've another mail adresse https://github.com/cozy/cozy-libs/blob/ec14a4ebae901767536e9c3ee7522c995534f1db/packages/cozy-harvest-lib/src/locales/en.json#L154 

we should also fix it

packages/cozy-harvest-lib/src/components/infos/TriggerErrorInfo.jsx
```diff
@@ -30,11 +43,11 @@ export class TriggerErrorInfo extends PureComponent {
             <Typography className="u-error" variant="h6" gutterBottom>
               {getErrorLocale(error, konnector, t, 'title')}
             </Typography>
-            <Typography variant="body1" component="div">
+          {supportMail ? ( <Typography variant="body1" component="div">
```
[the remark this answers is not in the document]
doubleface: Yes, there is a flickering this way but if we do the fetch before mounting the component, we will have a flickering for the whole component. I don't know if this is better.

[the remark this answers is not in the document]
doubleface: @Crash-- The CI finally passes. Are you ok with my response ?

packages/cozy-harvest-lib/src/helpers/konnectors.js
```diff
@@ -156,9 +167,26 @@ export const getErrorLocale = (error, konnector, t, suffixKey) => {
   })
 }
 
-export const getErrorLocaleBound = (error, konnector, lang, suffixKey) => {
+export const fetchSupportMail = async client => {
+  const result = await client.fetchQueryAndGetFromState({
+    definition: Q('io.cozy.settings').getById('context'),
+    options: {
+      as: 'contextSupportMail',
```
Crash--, line 174:
```suggestion
      as: 'io.cozy.settings/context',
```

Crash-- (approved):

PR #1395, opened 2021-12-19
fix: Downgrade babel-core in babel-preset

acezard:
Having a different babel-core version in the babel-preset created a lot of issues when the CI pushed the deploy commit. We now had a difference between packages.json and the lock file. Creating the lock file from scratch broke a lot of libs. Trying to fix every lib one by one has proven to be difficult.

The better way to handle the issue was simply to update the yarn.lock with only one lib (babel-core here). To sum up, never delete and recreate the yarn.lock of a monorepo

### DIFF
--- a/packages/babel-preset-cozy-app/package.json
+++ b/packages/babel-preset-cozy-app/package.json
@@ -17,7 +17,7 @@
     "validate.js"
   ],
   "dependencies": {
-    "@babel/core": "^7.16.0",
+    "@babel/core": "7.12.3",
     "@babel/eslint-parser": "^7.16.3",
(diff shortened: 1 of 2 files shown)


Crash--:
can you set the context in the commit please? In a few weeks we'll not remember why we've done that. It's really important specially for this kind of commit.

acezard:
> can you set the context in the commit please? In a few weeks we'll not remember why we've done that. It's really important specially for this kind of commit.



> can you set the context in the commit please? In a few weeks we'll not remember why we've done that. It's really important specially for this kind of commit.

updated

doubleface (approved):

Crash-- (approved):

PR #1398, opened 2021-12-23
feat: Add a check for flagship app in device-helper

acezard:
Naive check based on the belief that the react-native app will correctly inject a flag in the webview window

### DIFF
--- a/packages/cozy-device-helper/src/index.js
+++ b/packages/cozy-device-helper/src/index.js
@@ -6,7 +6,8 @@ export {
   isMobileApp,
   isAndroid,
   isIOS,
-  isMobile
+  isMobile,
+  isFlagshipApp
 } from './platform'
 export { getDeviceName } from './device'
 export { checkApp, startApp } from './apps'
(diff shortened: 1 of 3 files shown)


doubleface (approved):

trollepierre (approved):

Ldoppea:
Can you add a description in the PR and/or the commit to explain why do we need this?

I'm working on another task that checks if the app is flagship or not using Apple and Android stores validation, but my understanding is that it will impact only the app permissions when generating the connexion token.
I didn't hear about anything to inject in data-cozy-*. But maybe @nono is aware of this?

acezard:
> Can you add a description in the PR and/or the commit to explain why do we need this?
> 
> I'm working on another task that checks if the app is flagship or not using Apple and Android stores validation, but my understanding is that it will impact only the app permissions when generating the connexion token. I didn't hear about anything to inject in data-cozy-*. But maybe @nono is aware of this?

it's not injected in data attributes, but directly in the DOM of the webview (home app, which renders other apps). I'm not sure if what you're working on will be rendered inside. Maybe it's best if we have a chat about this

nono:
> But maybe @nono is aware of this?

Nope, it looks like it is the React Native part that set this DOM attribute to be read from the webview.

PR #1400, opened 2021-12-24
feat: Add new ESLint rule: default-param-last

acezard:
BREAKING CHANGE:
To handle this new rule in your project,
there are three available ways, ranked from best to worst:

- Refactor your code so the rule is respecterd
- Disable the rule on files where refactoring is not possible
- Disable it for the entire project if too many files are concerned

### DIFF
--- a/packages/cozy-scanner/src/ScannerUpload.js
+++ b/packages/cozy-scanner/src/ScannerUpload.js
@@ -12,7 +12,7 @@ import { CozyFile } from 'cozy-doctypes'
 export const doUpload = async (
   imageURI,
   qualification,
-  name = '',
+  name = '', // eslint-disable-line default-param-last
   dirId,
   onConflict,
   contentType
--- a/packages/cozy-sharing/src/SharingBanner/helpers/QueryParameter.js
+++ b/packages/cozy-sharing/src/SharingBanner/helpers/QueryParameter.js
@@ -1,3 +1,4 @@
+/* eslint-disable default-param-last */
 const arrToObj = (obj = {}, [key, val = true]) => {
   obj[key] = decodeURIComponent(val)
   return obj
--- a/packages/cozy-sharing/src/state.js
+++ b/packages/cozy-sharing/src/state.js
@@ -1,3 +1,4 @@
+/* eslint-disable default-param-last */
 import get from 'lodash/get'
 
 const RECEIVE_SHARINGS = 'RECEIVE_SHARINGS'
--- a/packages/eslint-config-cozy-app/basics.js
+++ b/packages/eslint-config-cozy-app/basics.js
@@ -22,6 +22,7 @@ module.exports = {
       }
     ],
     'no-param-reassign': 'warn',
-    'spaced-comment': ['error', 'always']
+    'spaced-comment': ['error', 'always'],
+    'default-param-last': ['error']
   }
 }


packages/cozy-procedures/src/redux/documentsDataSlice.js
```diff
@@ -91,7 +91,7 @@ const getCompletedCount = state =>
   Object.values(getData(state)).reduce((acc, { files }) => {
     files.map(file => {
       if (file !== undefined) {
-        acc++
+        acc++ // eslint-disable-line no-param-reassign
```
trollepierre, line 94: in this case, I guess it we should not disable the line but fix the reassignment.

trollepierre: I get something let this should work
```
const getCompletedCount = state => {
  let count = 0
  Object.values(getData(state)).forEach({ files }) => {
    files.map(file => {
      if (file !== undefined) {
        count++ 
        }
    })
})
return count
}
```

packages/cozy-scanner/src/ScannerUpload.js
```diff
@@ -17,6 +18,7 @@ export const doUpload = async (
   onConflict,
   contentType
 ) => {
+  /* eslint-disable default-param-last */
```
trollepierre, line 21:
```suggestion
  /* eslint-enable default-param-last */
```

packages/cozy-sharing/src/SharingBanner/helpers/QueryParameter.js
```diff
@@ -1,4 +1,6 @@
+/* eslint-disable default-param-last */
 const arrToObj = (obj = {}, [key, val = true]) => {
```
trollepierre, line 2:
```suggestion
const arrToObj = (obj = {}, [key, val = true]) => { // eslint-disable-line default-param-last
```

packages/cozy-sharing/src/state.js
```diff
@@ -89,7 +89,9 @@ const updateByIdItem = (state, id, updater) => {
       }
 }
 
+/* eslint-disable default-param-last */
 const indexSharing = (state = {}, sharing) => {
```
trollepierre, line 93:
```suggestion
const indexSharing = (state = {}, sharing) => { // eslint-disable-line default-param-last
```

trollepierre: I think we should do that in the whole file 🙏 (it's less verbose than he 2 lines)

packages/cozy-procedures/src/redux/documentsDataSlice.js
acezard: As I don't know this lib I preferred to disable each error rather than fix it (fear of unintended side-effects)

packages/cozy-sharing/src/state.js
acezard: @trollepierre I guess that makes sense even if I don't like it much. But the file shouldn't really change so disabling the whole file should be ok

packages/cozy-procedures/src/redux/documentsDataSlice.js
Ldoppea: Without needing to rewrite the entire method, we can keep the reduce and put the variable in the reduce:
```js
Object.values(getData(state)).reduce((acc, { files }) => {
    let newAcc = acc
    files.map(file => {
      if (file !== undefined) {
        newAcc++ // eslint-disable-line no-param-reassign
      }
    })
    return newAcc
  }, 0)
```

If we want to simplify the method, here is another try:
```js
Object.values(getData(state))
  .flatMap(data => data.files)
  .filter(file => filer !== undefined)
  .length
```

But I agree with @acezard, the PR's goal is not to fix the errors yet. We should try to fix them only if we have enough time to test the resulting behavior on project that use this method.

Now the real question is : why do we have this change related to `no-param-reassign` if the PR is named `add default-param-last`? Should we rename the PR/commit?

acezard: @Ldoppea I admit I had no idea why this error popped, I didn't touch it and eslint didn't change version.

packages/cozy-sharing/src/state.js
trollepierre: I am not proposing to disable the whole file. I prefer the current solution rather than disable the whole file.

But I prefer even more using `eslint-disable-line` rather than `eslint-disable` + `eslint-enable`

trollepierre (approved):

trollepierre:
Why did you close the PR? Why didn't you delete the branch?

PR #1402, opened 2021-12-30
chore(nvmrc): add nvmrc with 12 node version

trollepierre:


.nvmrc
```diff
@@ -0,0 +1 @@
+12.22
```
Ldoppea, line 1: Can we just specify the major version ? This would avoid `nvm` to fail if the user only have not exactly this version.
```suggestion
12
```

Ldoppea (approved):

PR #1406, opened 2022-01-04
feat: Do not allow multi accounts features for clientSide konnectors

doubleface:
Our client side konnectors launcher does not handle multiple
sessions for the same konnector yet.

We choose to deactive multi account feature for client side konnectors
at the moment.

Resolves https://trello.com/c/HoTn7eGr/119-%F0%9F%9A%80%F0%9F%8F%A0-e87-d%C3%A9sactiver-le-multi-compte-sur-les-connecteurs-c%C3%B4t%C3%A9-clients-05-jh


packages/cozy-harvest-lib/src/components/AccountModal.jsx
```diff
@@ -118,7 +118,7 @@ export class AccountModal extends Component {
     return (
       <>
         <KonnectorModalHeader konnector={konnector}>
-          {showAccountSelection ? (
+          {showAccountSelection && !konnector.clientSide ? (
```
Crash--, line 121: Can't we use the `showAccountSelection` props directly since it already does that same thing?

Crash--: @doubleface ?

doubleface: Right. Now i define this visibility in the upper level. I removed the unit tests

packages/cozy-harvest-lib/src/components/AccountModal.jsx
```diff
@@ -165,7 +165,7 @@ export class AccountModal extends Component {
               account={account}
               onAccountDeleted={onDismiss}
               addAccount={() => pushHistory('/new')}
-              showNewAccountButton={showNewAccountButton}
+              showNewAccountButton={ showNewAccountButton }
```
Crash--, line 168: lint

Crash-- (approved):
lint issue I think. but 👍

packages/cozy-harvest-lib/src/components/Routes.jsx
```diff
@@ -85,6 +85,8 @@ const Routes = ({ konnectorRoot, konnector, onDismiss, datacardOptions }) => {
                             accountId={match.params.accountId}
                             accountsAndTriggers={accountsAndTriggers}
                             onDismiss={onDismiss}
+                            showNewAccountButton={!konnector.clientSide}
+                            showAccountSelection={!konnector.clientSide}
```
trollepierre, line 89: nit: please add 1 test to test this feature/bug (not mandatory, as Routes.jsx is not tested and looks not so easy to test.)

trollepierre (changes requested):

packages/cozy-harvest-lib/src/components/Routes.jsx
doubleface: sorry I will pass this time since it is indeed not easy to test an this test does not have a lot of value

trollepierre (approved):

PR #1409, opened 2022-01-05
Fix wrong permission cozy sharing

trollepierre:


Crash--:
I think we can close it?

PR #1415, opened 2022-01-11
feat(deps): Eslint is devDependency & peerDependendy of eslint-config

trollepierre:
BREAKING CHANGE: Eslint needs to be installed 
or the project needs to use last version of cozy-script > v7

### DIFF
--- a/packages/eslint-config-cozy-app/package.json
+++ b/packages/eslint-config-cozy-app/package.json
@@ -22,7 +22,6 @@
     "@babel/eslint-parser": "^7.16.3",
     "@typescript-eslint/eslint-plugin": "^5.4.0",
     "@typescript-eslint/parser": "^5.4.0",
-    "eslint": "^8.3.0",
     "eslint-config-prettier": "^8.3.0",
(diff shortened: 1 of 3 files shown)


packages/eslint-config-cozy-app/package.json
```diff
@@ -31,5 +30,11 @@
     "eslint-plugin-vue": "^8.1.1",
     "prettier": "^2.5.0",
     "typescript": "^4.5.2"
```
trollepierre, line 32: @acezard : in my opinion,
- typescript, prettier should be devDependencies
- I wonder if like [eslint config airbnb](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/package.json), we want to set eslint-config-prettier, eslint-plugin-* as peerDependencies + devDependencies
- I don't know what to think about @babel/eslint-parser, @typescript-eslint/*

trollepierre:
@Crash-- : cozy-script v7 is the next version that we are going to release, in order to fix that

PR #1417, opened 2022-01-17
feat: Secure initial handshake

acezard:
Implement secured sync-like handshake initiated by the children when it is ready (this will resolve failed handshake when the Webview takes too long to render)

Implement single localMethods object in ReactNative IntentProvider. All children will use the same method object. This allows all Webviews to call all exposed methods.

### DIFF
--- a/packages/cozy-intent/package.json
+++ b/packages/cozy-intent/package.json
@@ -10,15 +10,15 @@
     "url": "https://github.com/cozy/cozy-libs/issues"
   },
   "dependencies": {
-    "cozy-device-helper": "^1.16.0",
-    "post-me": "0.4.5",
-    "typescript": "^4.5.2"
+    "cozy-device-helper": "1.16.0",
+    "post-me": "0.4.5"
   },
   "devDependencies": {
-    "@babel/core": "7.12.3",
-    "babel-preset-cozy-app": "2.0.1",
-    "eslint-config-cozy-app": "3.0.1",
-    "react": "17.0.2"
+    "@babel/core": "^7.12.3",
+    "babel-preset-cozy-app": "^2.0.1",
+    "eslint-config-cozy-app": "^4.0.0",
+    "react": "17.0.2",
+    "typescript": "^4.5.2"
   },
   "files": [
     "dist"
--- a/packages/cozy-intent/src/api/services/NativeService.ts
+++ b/packages/cozy-intent/src/api/services/NativeService.ts
@@ -1,45 +1,48 @@
-import { ParentHandshake } from 'post-me'
+import { Connection, ParentHandshake } from 'post-me'
 
 import { MessengerRegister } from '../models/messengers'
 import { NativeEvent } from '../models/events'
 import { NativeMessenger } from '../services/NativeMessenger'
 import { NativeMethodsRegister } from '../models/methods'
+import { StaticService } from './StaticService'
+import { Strings } from '../constants'
 import { WebviewRef } from '../models/environments'
-import { Numbers, Strings } from '../constants'
 
 export class NativeService {
   private messengerService = NativeMessenger
 
+  private localMethods: NativeMethodsRegister
   private messengerRegister = {} as MessengerRegister
 
-  public registerWebview = async (
-    webviewRef: WebviewRef,
-    localMethods: NativeMethodsRegister
-  ): Promise<void> => {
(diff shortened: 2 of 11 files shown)


packages/cozy-intent/package.json
```diff
@@ -10,9 +10,9 @@
     "url": "https://github.com/cozy/cozy-libs/issues"
   },
   "dependencies": {
-    "cozy-device-helper": "^1.16.0",
+    "cozy-device-helper": "1.16.0",
     "post-me": "0.4.5",
-    "typescript": "^4.5.2"
+    "typescript": "4.5.4"
```
Crash--, line 15: Is it really a dep? It can't be a devDep? cc @trollepierre

yarn.lock
```diff
@@ -5829,6 +5829,25 @@ babel-polyfill@6.26.0:
     core-js "^2.5.0"
     regenerator-runtime "^0.10.5"
 
+babel-preset-cozy-app@2.0.1, babel-preset-cozy-app@^2.0.1:
```
Crash--, line 5832: What is the change thant brings `babel-preset-cozy-app@2.0.1` in? 🤔

yarn.lock
```diff
@@ -8956,6 +8968,23 @@ eslint-cli@1.1.1:
     debug "^2.6.8"
     resolve "^1.3.3"
 
+eslint-config-cozy-app@3.0.1:
```
Crash--, line 8971: What is the change thant brings `eslint-config-cozy-app` in? 🤔

packages/cozy-intent/package.json
acezard: @Crash-- it was initially, but cozy-libs failed to build. I did not understand why. I guess I should take time to see what's up, but I thought it was not the most important thing right now. I should create an issue

yarn.lock
acezard: yea I can't really say to be honest, I'm going to investigate

yarn.lock
acezard: https://github.com/cozy/cozy-libs/pull/1417#discussion_r785691827

packages/cozy-intent/src/api/services/NativeService.ts
```diff
@@ -1,45 +1,48 @@
...
+
+  public tryEmit = async (event: NativeEvent): Promise<void> => {
+    if (!StaticService.isPostMeMessage(event)) return
+
+    const { message, uri } = StaticService.parseNativeEvent(event)
 
-    this.messengerRegister[
-      new URL(event.nativeEvent.url).hostname
-    ].messenger.onMessage(event)
+    if (message === Strings.webviewIsRendered)
```
Crash--, line 41: I don't understand the PR without the modification of the caller... 

It seems that the caller will send an event "webviewIsRendered" and then we start the handshake process. Right ? 

When are you going to send this message from the caller?

acezard: it is sent here https://github.com/cozy/cozy-libs/pull/1417/files#diff-416061f3de64eb10ebdfdbe98f8e86fcfbcb2f177e58e82e2462f009d64c4f6eR31

unless I misunderstood your question

Crash--: Nice, thank you! This is exactly what I was looking for. 

So we launch the "childHandshake" at the same time than the "ParentHandshake". Is it problematic? Or is it want we exactly want?

acezard: Well, childhandshake promise should be launched first, and Parent some milliseconds after, though I'm not sure if there are some cases where post-me would fail the handshake with this configuration. Post-me documentation isn't very clear on that... 

childhandshake has no timeout, that means for me it should be launched first, because it's waiting for the parent to ping it

Crash--: @doubleface can you share you expertise with us on this subject? :)

acezard: https://github.com/alesgenova/post-me/blob/main/packages/core/src/handshake.ts

for me, childhandshake is waiting forever for the parent ping. So it should be ok if it's always init before parent (which is the case here, since we send a message in a sync way). Which means the parent should find the children at the first ping

Crash-- (approved):

PR #1422, opened 2022-01-20
feat: Add `app` parameter to `cozy-intent`'s `openApp` method

Ldoppea:
Sender can now send app data to the native host in order to open the
native app when available instead of its webview version

---

Related PRs: 
- https://github.com/cozy/cozy-ui/pull/2015
- https://github.com/cozy/cozy-react-native/pull/65

### DIFF
--- a/packages/cozy-intent/src/api/models/applications.ts
+++ b/packages/cozy-intent/src/api/models/applications.ts
@@ -1,0 +1,20 @@
+/** App's description resulting of its manifest.webapp file */
+export interface AppManifest {
+  /** The app's slug */
+  slug: string
+
+  /** The app's mobile information */
+  mobile: AppManifestMobileInfo
+}
+
+/** App's mobile information. Used to describe the app scheme and its store urls */
+export interface AppManifestMobileInfo {
+  /** The app's URL scheme */
+  schema: string
+
+  /** The app's id on Google PlayStore */
+  id_playstore: string
+
+  /** The app's id on Apple AppStore */
+  id_appstore: string
+}
--- a/packages/cozy-intent/src/api/models/methods.ts
+++ b/packages/cozy-intent/src/api/models/methods.ts
@@ -1,4 +1,6 @@
+import { AppManifest } from './applications'
+
 export type NativeMethodsRegister = {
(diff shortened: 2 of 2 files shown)


acezard (approved):

packages/cozy-intent/src/api/models/applications.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface AppManifestMobileInfo {
+  /** The app's URL scheme */
+  schema: string
+
+  /** The app's id on Google PlayStore */
+  id_playstore: string
+
+  /** The app's id on Apple AppStore */
+  id_appstore: string
+}
```
Crash--, line 20: This should definitively be described to https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/manifest.js#L1 

When a lib deal with "Cozy's generic type", the type should not be defined in the lib, but instead in CozyClient, no?

packages/cozy-intent/src/api/models/methods.ts
```diff
@@ -1,4 +1,6 @@
+import { AppManifest } from './applications'
+
 export type NativeMethodsRegister = {
   logout: () => Promise<void>
-  openApp: (href: string) => Promise<void>
+  openApp: (href: string, app: AppManifest) => Promise<void>
```
Crash--, line 5: I prefer to have an object as argument instead of having two args : `openApp: ({href, app})`. Because I don't like to write "openApp(null, {...})". And pretty sure we'll start writing this kind of code pretty soon

packages/cozy-intent/src/api/models/applications.ts
Ldoppea: Oh I was not aware of this file. I'll move the def in cozy-client later today 👍

packages/cozy-intent/src/api/models/methods.ts
Ldoppea: I was not confident enough to remove the href param, but I have the intuition it will be obsolete soon.

I should have a better understanding of possible use cases after I implement the cozy-store connectors link. I propose to do the refactoring after that.

PR #1428, opened 2022-01-22
feat: Bump dependencies to latest minor

acezard:
Avoiding major versions for now as it doesn't look
like it is worth the migration effort at the moment


acezard:
I'm thinking the test will fail until https://github.com/cozy/cozy-libs/pull/1429 is merged

PR #1437, opened 2022-01-27
fix(renovate): use fix as commit type of renovate PR

trollepierre:
We would like to use "fix" as commit type of renovate PR.
According to https://docs.renovatebot.com/semantic-commits/ ,
changing the Semantic Commit type is possible by extending
":semanticCommitTypeAll(fix)" to your extends array.
PR titles and commit messages start with "fix(deps):"

### DIFF
--- a/packages/renovate-config-cozy/package.json
+++ b/packages/renovate-config-cozy/package.json
@@ -5,7 +5,8 @@
   "renovate-config": {
     "default": {
       "extends": [
-        "config:base"
+        "config:base",
+        ":semanticCommitTypeAll(fix)"
       ],
       "rangeStrategy": "pin",
       "packageRules": [


packages/renovate-config-cozy/package.json
```diff
@@ -5,7 +5,8 @@
   "renovate-config": {
     "default": {
       "extends": [
-        "config:base"
+        "config:base",
+        ":semanticCommitTypeAll(fix)"
```
trollepierre, line 9: I am wondering if this needs to be applied only for lib (not for app). What do you think?

Crash--: I don't have strong opinion on it. Let's use it every where and see what happens next?

Crash-- (approved):

trollepierre:
Capture d’écran 2022-02-01 à 11 27 06
it's working 🎉

PR #1440, opened 2022-01-28
feat: Add the defaultFolderPath attribute to the account

doubleface:
This defaultFolderPath attribute will be used by the stack recreate the
connector's destination folder if it has been removed by the user.

### DIFF
--- a/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
+++ b/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
@@ -476,7 +476,7 @@ export class ConnectionFlow {
       t
     })
 
-    await this.addDefaultFolderPath(client, {
+    await this.ensureDefaultFolderPathInAccount(client, {
       trigger: ensuredTrigger,
       account,
       konnector
@@ -489,26 +489,43 @@ export class ConnectionFlow {
   }
 
   /**
-   * Add a default folder path to the account if any folder is needed by the connector
-   * and/or the account does not have it yet.
+   * Ensures there is a defaultFolderPath attribute in the account
+   * if any folder is needed by the connector.
    * The account is saved if any change is done to it.
-   * This defaultFolderPath is needed by the stack to recreate the destination folder.
-   * had beed removed by any mean : drive application, desktop application etc
+   * This defaultFolderPath is needed by the stack to recreate the
+   * destination folder if removed by any mean : drive application,
+   * desktop application etc
    *
    * @param  {CozyClient} client - A cozy client
    * @param  {io.cozy.triggers} options.trigger
    * @param  {io.cozy.accounts} options.account
-   * @param  {io.cozy.konnector} options.konnector
+   * @param  {io.cozy.konnectors} options.konnector
    */
-  async addDefaultFolderPath(client, { trigger, account, konnector }) {
+  async ensureDefaultFolderPathInAccount(
+    client,
+    { trigger, account, konnector }
+  ) {
     const folderId = get(trigger, 'message.folder_to_save')
-    if (folderId && !account.defaultFolderPath) {
-      const { data: folder } = await client.query(
-        Q('io.cozy.files').getById(folderId)
-      )
-      account.defaultFolderPath = folder.path
-      await saveAccount(client, konnector, account)
+    if (!folderId) {
+      return account
     }
+
+    let folder
+    try {
+      const result = await client.query(Q('io.cozy.files').getById(folderId))
+      folder = result.data
+    } catch (err) {
+      logger.warn(
+        `ConnectionFlow.ensureDefaultFolderPath: folder ${folderId} does not exist. Could not ensure defaultFolderPath. ${err.message}`
+      )
+      return account
+    }
+    if (folder.path !== account.defaultFolderPath) {
+      account.defaultFolderPath = folder.path
+      const savedAccount = await saveAccount(client, konnector, account)
+      return savedAccount
+    }
+    return account
   }
 
   /**
--- a/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
+++ b/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
@@ -3,6 +3,7 @@ import get from 'lodash/get'
 
 import Realtime from 'cozy-realtime'
 import flag from 'cozy-flags'
+import { Q } from 'cozy-client'
 
 import {
   fetchReusableAccount,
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -474,12 +475,42 @@ export class ConnectionFlow {
...
     this.trigger = ensuredTrigger
     this.emit(UPDATE_EVENT)
     await this.launch()
   }
 
+  /**
+   * Add a default folder path to the account if any folder is needed by the connector
+   * and/or the account does not have it yet.
+   * The account is saved if any change is done to it.
+   * This defaultFolderPath is needed by the stack te recreate the destination folder this
```
trollepierre, line 495:
```suggestion
   * This defaultFolderPath is needed by the stack to recreate the destination folder.
```

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -474,12 +475,42 @@ export class ConnectionFlow {
...
+   * had beed removed by any mean : drive application, desktop application etc
```
trollepierre, line 496:
```suggestion
   * This has been removed by every Cozy applications: Drive, Desktop, etc
```

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -474,12 +475,42 @@ export class ConnectionFlow {
...
+   *
+   * @param  {CozyClient} client - A cozy client
+   * @param  {io.cozy.triggers} options.trigger
+   * @param  {io.cozy.accounts} options.account
+   * @param  {io.cozy.konnector} options.konnector
+   */
```
trollepierre, line 502: nit: I don't understand why Doctypes are singular and plural:
-  io.cozy.triggers => plural
- io.cozy.accounts => plural
- io.cozy.konnector => singular

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -474,12 +475,42 @@ export class ConnectionFlow {
...
+  async addDefaultFolderPath(client, { trigger, account, konnector }) {
```
trollepierre, line 503: this method should be called "assureDefaultFolderPathExistency" or "addDefaultFolderPathIfNotFound" or "addEventuallyDefaultFolderPath" 

in my opinion, an `add` method MUST add something

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -474,12 +475,42 @@ export class ConnectionFlow {
...
+    const folderId = get(trigger, 'message.folder_to_save')
+    if (folderId && !account.defaultFolderPath) {
+      const { data: folder } = await client.query(
+        Q('io.cozy.files').getById(folderId)
+      )
+      account.defaultFolderPath = folder.path
+      await saveAccount(client, konnector, account)
+    }
+  }
```
trollepierre, line 512: in my experience, the add / post method should return the added object or a promise containing the created object.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -474,12 +475,42 @@ export class ConnectionFlow {
...
+
```
trollepierre, line 513: On other hand, let's return 'null' or 'not necessary' when default folder path exists?

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -388,6 +389,34 @@ describe('ConnectionFlow', () => {
...
+      expect(flow.account).toEqual(fixtures.updatedAccount)
+      expect(saveAccount).toHaveBeenCalledWith(
+        client,
+        fixtures.konnectorWithFolder,
+        {
+          ...fixtures.existingAccount,
+          defaultFolderPath: '/default/folder/path'
+        }
+      )
+    })
```
trollepierre, line 418: let's test also those cases:
- folderId not defined
- account.defaultFolderPath exists
- what if client.query return empty/undefined (I don't know if it's possible)

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -388,6 +389,34 @@ describe('ConnectionFlow', () => {
...
     })
 
+    it('should add the defaultFolderPath to the account', async () => {
+      const { flow, client } = setup()
+      prepareTriggerAccount.mockResolvedValue(fixtures.updatedAccount)
+
+      ensureTrigger.mockImplementation(
+        async () => fixtures.createdTriggerWithFolder.attributes
+      )
+      client.query.mockImplementation(async () => ({
```
trollepierre, line 399:
```suggestion
      client.query.mockResolvedValue({
```

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -388,6 +389,34 @@ describe('ConnectionFlow', () => {
       expect(flow.account).toEqual(fixtures.updatedAccount)
     })
 
+    it('should add the defaultFolderPath to the account', async () => {
+      const { flow, client } = setup()
+      prepareTriggerAccount.mockResolvedValue(fixtures.updatedAccount)
+
+      ensureTrigger.mockImplementation(
```
trollepierre, line 396: ensureTrigger.mockResolvedValue(fixtures.createdTriggerWithFolder.attributes)

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -388,6 +389,34 @@ describe('ConnectionFlow', () => {
...
+          path: '/default/folder/path'
+        }
+      }))
+      await flow.ensureTriggerAndLaunch(client, {
+        account: fixtures.existingAccount,
+        trigger: fixtures.existingTrigger,
+        konnector: fixtures.konnectorWithFolder
+      })
+      expect(flow.account).toEqual(fixtures.updatedAccount)
+      expect(saveAccount).toHaveBeenCalledWith(
```
trollepierre, line 410: let's add
- expect(client.query).toHaveBeenCalledWith('something')

trollepierre (changes requested):
some improvements to update. Feel free to contact me to discuss

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: Simple mistake. The good one is plural

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: You are right. What do you think of ensureDefaultFolderPath ?

doubleface: To be coherent with ensureTrigger

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: Right, let's do this

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: I would prefer to return the unchanged account for consistency

trollepierre: so , this method returns account in both case?

then, the naming of the function should mention account, no?

or the method is a "setter"? 🤔

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
doubleface: - folderId not defined : already tested implicitely in the other tests of the suite
- the stack will ensure itself the destination folder exists and use defaultFolderPath to name it. This is not the responsibility of harvest to check this
- this is not supposed to happend since it is the responsibility of ensureTrigger, before, to also ensure the destination folder exists. But you are right, i will at least catch the exception in this case with a warning log.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: ensureDefaultFolderPathInAccount ?

trollepierre (approved):
👏 🎉

PR #1447, opened 2022-02-02
feat: Ensure isFlagshipApp returns a boolean value

acezard:
It's less confusing to rely on strict boolean retrn value instead of true|falsy value

### DIFF
--- a/packages/cozy-device-helper/src/platform.js
+++ b/packages/cozy-device-helper/src/platform.js
@@ -24,4 +24,4 @@ export const isIOS = () =>
 // isMobile checks if the user is on a smartphone : native app or browser
 export const isMobile = () => isAndroid() || isIOS()
 
-export const isFlagshipApp = () => window.cozy?.isFlagshipApp
+export const isFlagshipApp = () => Boolean(window.cozy?.isFlagshipApp)
--- a/packages/cozy-device-helper/src/platform.spec.js
+++ b/packages/cozy-device-helper/src/platform.spec.js
@@ -1,0 +1,46 @@
+import {
+  isWebApp,
+  isMobileApp,
+  isIOSApp,
+  isAndroidApp,
+  getPlatform,
+  isFlagshipApp
+} from './platform'
+
+describe('platforms', () => {
+  it('should identify is a web application', () => {
+    expect(isWebApp()).toBeTruthy()
+  })
+  it('should identify is a mobile application', () => {
+    window.cordova = true
+    expect(isMobileApp()).toBeTruthy()
+    window.cordova = undefined
+    expect(isMobileApp()).toBeFalsy()
+  })
+  it('should identify is an iOS or Android application', () => {
+    window.cordova = { platformId: 'ios' }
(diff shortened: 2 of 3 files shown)


packages/cozy-device-helper/src/platform.js
```diff
@@ -24,4 +24,4 @@ export const isIOS = () =>
 // isMobile checks if the user is on a smartphone : native app or browser
 export const isMobile = () => isAndroid() || isIOS()
 
-export const isFlagshipApp = () => window.cozy?.isFlagshipApp
+export const isFlagshipApp = () => Boolean(window.cozy?.isFlagshipApp)
```
trollepierre, line 27: let's add a test and/or jsdoc about it 🙏

trollepierre (changes requested):

packages/cozy-device-helper/src/platform.js
acezard: well there is already a test, it's still passing

trollepierre: 1) I think we should update the test to prevent regression (of reverting your commit change for example):

```
  it('should identify as a Flagship app webview', () => {
    window.cozy = undefined
    expect(isFlagshipApp()).toBe(false)
    window.cozy = {}
    expect(isFlagshipApp()).toBe(false)
    window.cozy = { isFlagshipApp: '' }
    expect(isFlagshipApp()).toBe(false)
    window.cozy = { isFlagshipApp: true }
    expect(isFlagshipApp()).toBe(true)
  })

```

trollepierre: 2. I prefer to test platform inside platform.spec.js rather than index.spec.js , it's more straightforward, to find matching test. Could you move the test please?

acezard: updated

packages/cozy-device-helper/src/platform.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    window.cordova = undefined
+    expect(isMobileApp()).toBeFalsy()
+  })
+  it('should identify is an iOS or Android application', () => {
+    window.cordova = { platformId: 'ios' }
+    expect(isIOSApp()).toBeTruthy()
+    expect(isAndroidApp()).toBeFalsy()
+    window.cordova = { platformId: 'android' }
+    expect(isIOSApp()).toBeFalsy()
+    expect(isAndroidApp()).toBeTruthy()
```
trollepierre, line 26: very good! THANKS 🎉 

nit: Just one more point. We just notice that the Jest matcher toBeTruthy is not very adapted to check boolean value.
Using toBe(true) is more strict.

I think we should use it everywhere (especially in the moved tests) 🙏

trollepierre (approved):
well done! 👏

packages/cozy-device-helper/src/platform.spec.js
acezard: I agree but it should be done at a later time, eg when converting the library to TS

PR #1448, opened 2022-02-02
feat: Send context to cozy-bar if it exists

acezard:
That way cozy-bar can use the same IntentService as the parent app

- [x] Implement feature
- [x] Test feature

### DIFF
--- a/packages/cozy-intent/src/view/components/WebviewIntentProvider.spec.tsx
+++ b/packages/cozy-intent/src/view/components/WebviewIntentProvider.spec.tsx
@@ -1,18 +1,29 @@
-import React from 'react'
-import { render, screen } from '@testing-library/react'
 import 'mutationobserver-shim'
+import React from 'react'
+import { render } from '@testing-library/react'
 
+import { CozyBar } from '../../api/models/applications'
 import { WebviewIntentProvider } from './WebviewIntentProvider'
+import { WebviewService } from '../../api/services/WebviewService'
+
 import {
-  mockWebviewWindow,
   mockChildHandshake,
   mockConnection,
+  mockCozyBar,
   mockIsFlagshipApp,
-  mockWebviewService
+  mockSetWebviewContext,
+  mockWebviewService,
+  mockWebviewWindow
 } from '../../tests/mocks'
 
 // eslint-disable-next-line no-global-assign
 window = mockWebviewWindow()
+
+declare const global: {
+  cozy?: CozyBar
+}
+
+global.cozy = mockCozyBar
 
 jest.mock('cozy-device-helper', () => ({
   isFlagshipApp: jest.fn()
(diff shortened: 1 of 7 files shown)


packages/cozy-intent/src/view/components/WebviewIntentProvider.spec.tsx
```diff
@@ -59,4 +71,50 @@ describe('WebviewIntentProvider', () => {
...
+    expect(mockSetWebviewContext).toHaveBeenCalledWith(mockWebviewService)
+    expect(mockSetWebviewContext).toBeCalledTimes(1)
+  })
+
+  it('sets cozy-bar context if cozy-bar does exist when not provided with a context', async () => {
+    global.cozy = mockCozyBar
+
+    render(<WebviewIntentProvider>Hello</WebviewIntentProvider>)
+
+    expect(await screen.findByText('Hello')).toBeTruthy()
```
trollepierre, line 116: Please use this syntax:

it('sets cozy-bar context if cozy-bar does exist when not provided with a context', async () => {
    global.cozy = mockCozyBar

    const { findByText } = render(<WebviewIntentProvider>Hello</WebviewIntentProvider>)

    expect(await findByText('Hello')).toBeTruthy()
    expect(mockSetWebviewContext).toBeCalledWith(expect.any(WebviewService))
    expect(mockSetWebviewContext).toBeCalledTimes(1)
  })****

packages/cozy-intent/src/view/components/WebviewIntentProvider.spec.tsx
```diff
@@ -59,4 +71,50 @@ describe('WebviewIntentProvider', () => {
...
+    expect(mockSetWebviewContext).toBeCalledWith(expect.any(WebviewService))
+    expect(mockSetWebviewContext).toBeCalledTimes(1)
```
trollepierre, line 118: in my opinion, there should be 2 tests:
```
it('sets cozy-bar context if cozy-bar does exist when not provided with a context', async () => {
    global.cozy = mockCozyBar

    const { queryByText } = await render(<WebviewIntentProvider>Hello</WebviewIntentProvider>)

    expect(queryByText('Hello')).toBeTruthy()
  })

  it('sets cozy-bar context if cozy-bar does exist when not provided with a context 2', async () => {
    global.cozy = mockCozyBar

    render(<WebviewIntentProvider>Hello</WebviewIntentProvider>)

    await waitFor(() => {
      expect(mockSetWebviewContext).toBeCalledWith(expect.any(WebviewService))
      expect(mockSetWebviewContext).toBeCalledTimes(1)
    })
  })
```

packages/cozy-intent/src/view/components/WebviewIntentProvider.spec.tsx
acezard: Updated

packages/cozy-intent/src/view/components/WebviewIntentProvider.spec.tsx
acezard: Kinda disagree here, the "hello" query is a smoke test in every test to ensure the component doesn't crash. I think it's a bit redundant to have a test for it everytime (it would double the amount of tests)

trollepierre (approved):

PR #1450, opened 2022-02-09
feat: Improve cozy-intent edge cases handling

acezard:
### This PR contains:

**Handle edge cases:**
* Throw an error if a webview tries to init itself when it's already instantiated
* Implement stricter hooks conditional in WebviewIntentProvider in order to avoid multiple calls to WebviewService or to cozy-bar

**Refactor:**
* Remove StaticService, it had no use
* Put StaticService methods where they should be in the first place

**Add util:**
* Implemented an interpolation util to handle query strings in error messages for better code readability


packages/cozy-intent/src/api/services/StaticService.ts
```diff
@@ -34,6 +38,14 @@ export const StaticService = {
     return nativeEvent.data.includes(Strings.postMeSignature)
   },
 
+  getBarInitAPI(): ((webviewContext: WebviewService) => void) | undefined {
+    try {
+      return cozy?.bar?.setWebviewContext
+    } catch (err) {
+      return undefined
```
Crash--, line 45: I think we should definitely log this error. No?

acezard: I guess so yes, btw this error should only happen in dev mode when webpack for some reason fails to reach the cozy bar module (happened only once to me, and not sure why it happened)

packages/cozy-intent/src/utils/index.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { interpolate } from '.'
+
+describe('interpolate', () => {
+  it('interpolates with one argument', () => {
+    const template = 'This is ${arg} argument.'
+
+    expect(interpolate(template, { arg: 'one' })).toBe('This is one argument.')
+  })
+
+  it('interpolates with five argument', () => {
```
trollepierre, line 10:
```suggestion
  it('interpolates with five arguments', () => {
```

packages/cozy-intent/src/utils/index.ts
```diff
@@ -0,0 +1,20 @@
+export const interpolate = (
+  str: string,
+  params: Record<string, string>
+): string => {
+  if (!params) return str
+
+  const names = Object.keys(params)
+  const vals = Object.values(params)
+
+  // I do not see security risks here, we're not executing user input
```
trollepierre, line 10: yes, you are right, let's remove this comment.

packages/cozy-intent/src/utils/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-implied-eval
+  const makeInterpolate = new Function(...names, `return \`${str}\`;`)
+
+  try {
+    return makeInterpolate(...vals) as string
+  } catch (error) {
+    // String template and params did not match
```
trollepierre, line 17: I don't understand why we return here, instead of throwing error

trollepierre: let's test this, to clarify 🙏

packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
```diff
@@ -1,37 +1,81 @@
...
 }
 
+const getBarInitAPI = ():
+  | ((webviewContext: WebviewService) => void)
+  | undefined => {
+  try {
+    return cozy?.bar?.setWebviewContext
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error((err as Error).stack)
```
trollepierre, line 28: why not using cozy-logger?

packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
```diff
@@ -1,37 +1,81 @@
...
+
+    // If we fail to get the cozy-bar, we want to assume it doesn't exist
+    return undefined
```
trollepierre, line 31: return null then

packages/cozy-intent/src/api/services/NativeService.ts
```diff
@@ -54,23 +65,27 @@ export class NativeService {
   ): Promise<Connection> => await ParentHandshake(messenger, this.localMethods)
 
   public tryEmit = async (event: NativeEvent): Promise<void> => {
-    if (!StaticService.isPostMeMessage(event)) return
+    if (!this.isPostMeMessage(event)) return
```
trollepierre, line 68: should we insert in the return type it can return void directly?

packages/cozy-intent/src/utils/index.ts
acezard: It's already tested actually. It will throw when the interpolate function did not receive adequate params to create the final string. It's better to just return the template string with no variable injection than nothing (especially since it's used for logging purposes and not end-user facing)

acezard: I'm going to expand the comment then

packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
acezard: Yeah, I guess that's a good idea.

packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
acezard: Undefined is more suited to describe the eventual runtime type of a failed cozy-bar fetch

packages/cozy-intent/src/api/services/NativeService.ts
acezard: TS will refuse that since the method is declared as async

trollepierre (approved):

PR #1456, opened 2022-02-11
fix: Temporarily remove console calls

acezard:
Made cozy-drive tests fail


trollepierre:
not needed, let's close 🙏

PR #1458, opened 2022-02-11
chore: Add jest coverage config

acezard:
Add a new test config that will fail if cozy-intent is not test covered below 80%.

Required to add new tests to reach this value. See image below

image

### DIFF
--- a/packages/cozy-intent/package.json
+++ b/packages/cozy-intent/package.json
@@ -15,6 +15,7 @@
   "devDependencies": {
     "@babel/core": "7.16.12",
     "@testing-library/react": "10.4.9",
+    "@testing-library/react-hooks": "^7.0.2",
     "babel-preset-cozy-app": "^2.0.2",
     "cozy-device-helper": "^1.17.0",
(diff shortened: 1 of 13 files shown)


packages/cozy-intent/package.json
```diff
@@ -50,7 +51,8 @@
     "lint": "eslint . --ext .ts,.tsx",
     "lint:fix": "yarn lint --fix",
     "prepublishOnly": "yarn build",
-    "test": "jest"
+    "test": "jest --config=./src/tests/jest.config.js",
+    "test:watch": "yarn test --watchAll"
```
trollepierre, line 55:
```suggestion
    "test:watch": "yarn test --watch"
```

trollepierre: I recommend using https://jestjs.io/docs/cli#--watch

trollepierre (approved):

packages/cozy-intent/package.json
acezard: Well it was required to have the coverage output shown during watch

PR #1464, opened 2022-02-15
feat: Do not allow multiple accounts with same BI connection ID

doubleface:
This avoids to have hidden accounts in bank's accounts view

Also, having multiple triggers associated to the same BI connection can cause
a lot of 409 errors when a BI webhook is triggered for a given BI
connection which triggers multiple triggers

Related to https://trello.com/c/URIhNqZc/205-bugs-identifier-les-diff%C3%A9rences-dans-les-taux-d%C3%A9checs-de-connexions-bancaires-bug-de-409-document-update-conflict

image


packages/cozy-harvest-lib/src/services/budget-insight.spec.js
```diff
@@ -604,6 +605,60 @@ describe('createOrUpdateBIConnection', () => {
...
+    const konnector = {
+      slug: 'bankingconnectortest'
+    }
+
+    jest.spyOn(flow, 'saveAccount').mockImplementation(account => ({
+      _id: 'created-account-id',
+      ...account
+    }))
+
+    jest.spyOn(client, 'query').mockImplementation(async qdef => {
```
trollepierre, line 630:
```suggestion
    jest.spyOn(client, 'query').mockImplementation(async ({doctype}) => {
```

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -166,10 +170,24 @@ export const createOrUpdateBIConnection = async ({
       }
     }
     logger.info('Creating or updating connection...')
-    const connection = await (connId
+    const isUpdate = Boolean(connId)
+    const connection = await (isUpdate
       ? updateBIConnection(config, connId, credsToSend, tempToken)
       : createBIConnection(config, credsToSend, tempToken))
 
+    if (!isUpdate) {
```
trollepierre, line 178: nit: personnally I would have tested the throw would not happen during creation (in case)

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -166,10 +170,24 @@ export const createOrUpdateBIConnection = async ({
...
+      const sameAccount = await fetchAccountWithBiConnection({
```
trollepierre, line 179: actually, we need only a boolean has return type.

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -166,10 +170,24 @@ export const createOrUpdateBIConnection = async ({
...
+        client,
+        konnector,
+        connectionId: connection.id
+      })
+      if (sameAccount) {
+        throw new KonnectorJobError(
+          'ACCOUNT_WITH_SAME_IDENTIFIER_ALREADY_DEFINED'
```
trollepierre, line 186: don't we need to explicit any id?

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -505,6 +523,44 @@ const shouldResumeConnection = error => {
...
+        })
+        .indexFields(['message.konnector'])
+    )
+  ])
+
+  const accountsIndex = keyBy(accountsResult.data, '_id')
+  const trigger = triggersResult.data.find(
+    t => accountsIndex[get(t, 'message.account')]
+  )
+  return trigger ? accountsIndex[trigger.message.account] : null
```
trollepierre, line 561: nit: personnally I would have tested a case where
```
 else if (qdef.doctype === 'io.cozy.triggers') {
        return {
          data: [
            {
              message: { account: 'another_account_id' }
            }
          ]
        }
```
to be sure we go in the null case => not throwing

trollepierre (approved):
upgrading cozy-client to 27.14 🤩

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -505,6 +523,44 @@ const shouldResumeConnection = error => {
...
 }
 
+/**
+ * Tries to find an existing account, associated to an existing trigger
+ * with the given bi connection id
+ *
+ * @param  {CozyClient} options.client - Cozy client
+ * @param  {Object} options.konnector - Konnector manifest
+ * @param  {Integer} options.connectionId - BI connection id
+ * @return {Account|undefined} An account with a trigger with the same identifier if any
```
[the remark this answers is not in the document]
doubleface: no, good catch

packages/cozy-harvest-lib/src/services/budget-insight.js
doubleface: Since, as you ask in the following comment, it would be better to add the corresponding account id. I prefer to keep it like this.

packages/cozy-harvest-lib/src/services/budget-insight.spec.js
```diff
@@ -604,6 +605,127 @@ describe('createOrUpdateBIConnection', () => {
...
+        login: '1234'
+      },
+      data: {
+        auth: {
+          bi: {
+            connId: 'created-bi-connection-id-789'
+          }
+        }
+      }
+    })
```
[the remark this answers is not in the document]
doubleface: ?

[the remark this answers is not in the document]
trollepierre: I am not sure this test is failing. I think we forgot to assert an expectation about the return of the promise here.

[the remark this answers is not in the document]
doubleface: The return of the promise is tested :shrug:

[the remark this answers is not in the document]
trollepierre: oops I didn't see well, my bad 😅

PR #1467, opened 2022-02-22
fix(cozy-intent): Allow undefined hook return

acezard:
When throwing an error, some race conditions could make an app crash which is very unwanted. Allowing the context to be undefined (which is idiomatic React) is much safer. 

Also reduced the amount of warning output and improved an error message. Updated test file to reflect changes made.

### DIFF
--- a/packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
+++ b/packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
@@ -27,14 +27,12 @@ function isWebviewWindow(window: Window): window is WebviewWindow {
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 const getBarInitAPI = (): ((webviewContext: WebviewService) => void) | void => {
   try {
-    if (cozy!.bar!.setWebviewContext === undefined) {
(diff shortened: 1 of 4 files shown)


packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
```diff
@@ -27,14 +27,12 @@ function isWebviewWindow(window: Window): window is WebviewWindow {
...
-    if (cozy!.bar!.setWebviewContext === undefined) {
+    if (cozy!.bar && cozy!.bar.setWebviewContext === undefined) {
       return console.warn(strings.errorCozyBarAPIMissing)
     }
 
     return cozy!.bar!.setWebviewContext
   } catch (err) {
-    console.warn((err as Error).stack)
-
     return console.warn(strings.errorGetCozyBarAPI)
```
trollepierre, line 36: let's use logger.warn

trollepierre (approved):

packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
acezard: Indeed, I will refactor the logging part in another branch
For now a quick publish is needed for Amiral

PR #1470, opened 2022-02-24
feat(cozy-intent): Add new method signatures

acezard:
Will be exposed by the native application to
allow changing interface color


trollepierre (approved):
in my opinion, this is docs or fix, not feat

PR #1473, opened 2022-02-28
chore(dependabot): Add prefix fix dependabot and schedule weekend

trollepierre:


trollepierre:
it works 🎉 
Capture d’écran 2022-03-14 à 09 30 51

PR #1476, opened 2022-03-01
fix(cozy-intent): Allow cozy-bar V8 to inject callback

acezard:
Cozy-bar V8 does not have the same API as V7 after further research.
Added an option to inject manually the cozy-bar method to set WebviewContext

### DIFF
--- a/packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
+++ b/packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
@@ -14,6 +14,7 @@ declare const cozy: CozyBar | undefined
 
 interface Props {
   children?: React.ReactChild
+  setBarContext?: (webviewContext: WebviewService) => void
   webviewService?: WebviewService
 }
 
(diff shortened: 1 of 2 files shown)


packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
```diff
@@ -71,11 +72,12 @@ const isValidEnv = (): boolean => {
 
 export const WebviewIntentProvider = ({
   children,
+  setBarContext,
   webviewService
 }: Props): ReactElement => {
   const [connection, setConnection] = useState<WebviewConnection>()
   const [service, setService] = useState<WebviewService | void>(webviewService)
-  const setBarWebviewContext = getBarInitAPI()
+  const setBarWebviewContext = setBarContext || getBarInitAPI()
```
trollepierre, line 80: I will approuve once this test is added.
```
  it('should allow cozy-bar v8 to inject callback', async () => {
    const mockSetBarContext = jest.fn()
    const { findByText } = render(
      <WebviewIntentProvider setBarContext={mockSetBarContext}>Hello</WebviewIntentProvider>
    )

    expect(await findByText('Hello')).toBeTruthy()
    expect(mockSetBarContext).toBeCalledWith(expect.any(WebviewService))
    expect(mockSetBarContext).toBeCalledTimes(1)
  })
  ```

acezard: added

PR #1478, opened 2022-03-03
feat: Introduce hide/showSplashScreen native methods in expected types

trollepierre:


packages/cozy-intent/src/api/models/methods.ts
```diff
@@ -6,4 +6,5 @@ export type NativeMethodsRegister = {
   openApp: (href: string, app: AppManifest) => Promise<void>
   setNavBarColor: (color: string) => void
   setStatusBarColor: (color: string) => void
+  hideSplashScreen: () => void
```
Crash--, line 9: I think you can also add showSlashScreen. Will be needed soon

Crash-- (approved):
I approve, but since you need to have a new published version with this change, you need to rename your commit to something like feat

PR #1479, opened 2022-03-08
feat: Implement simpler API signature

acezard:
to handle React Native UI manipulations

Instead of having 4 methods to handle statusbar, icons etc, only expose one method with one argument (an object). 
Then react native has to handle all the possible calls but the API side stays simple

### DIFF
--- a/packages/cozy-intent/src/utils/index.ts
+++ b/packages/cozy-intent/src/utils/index.ts
@@ -1,3 +1,5 @@
+import logger from 'cozy-logger'
+
 export const interpolate = (
   str: string,
   params: Record<string, string>
@@ -18,3 +20,5 @@ export const interpolate = (
     return str
   }
 }
+
+export const log = logger.namespace('cozy-intent')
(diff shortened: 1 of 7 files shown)


packages/cozy-intent/src/utils/index.ts
```diff
@@ -18,3 +18,15 @@ export const interpolate = (
...
   }
 }
+
+declare const __DEVELOPMENT__: boolean | void
+
+export const warn = (message: string): void => {
+  try {
+    if (__DEVELOPMENT__) {
+      // eslint-disable-next-line no-console
+      console.warn(message)
```
trollepierre, line 28: let's use logger from cozy-logger please 🙏

trollepierre (approved):

packages/cozy-intent/src/utils/index.ts
acezard: I wanted to keep the 0 dependency constraint here. Would you deem it unwanted? Though cozy-logger is 1.5kb of course, which is not a lot

Crash--: Remember our internal rules: If a cozy lib depends of another cozy-libs, then it's not a direct dep. It's a devDep + peerDep. The consuming app should already have the wanted dep

trollepierre (approved):

PR #1486, opened 2022-03-14
docs: Update FlagshipUI interface

acezard:
Setting the overlay value explicitely is much easier to handle

### DIFF
--- a/packages/cozy-intent/src/api/models/applications.ts
+++ b/packages/cozy-intent/src/api/models/applications.ts
@@ -33,11 +33,14 @@ export enum NativeThemes {
 }
 
 export interface FlagshipUI {
-  /** Has to be a valid React Native color string or will be ignored */
-  statusBarBackground?: string
-  statusBarIcons?: NativeThemes
-
-  /** Has to be a valid React Native color string or will be ignored */
-  navbarBackground?: string
-  navbarIcons?: NativeThemes
+  /** Has to be a <color> CSS data type */
+  bottomBackground?: string
+  bottomColor?: NativeThemes
+  /** Has to be 'rgba(R, G, B[, A])' to enable, 'transparent' to disable */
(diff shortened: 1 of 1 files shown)


packages/cozy-intent/src/api/models/applications.ts
```diff
@@ -33,11 +33,14 @@ export enum NativeThemes {
...
-  statusBarBackground?: string
-  statusBarIcons?: NativeThemes
-
-  /** Has to be a valid React Native color string or will be ignored */
-  navbarBackground?: string
-  navbarIcons?: NativeThemes
+  /** Has to be a <color> CSS data type */
+  bottomBackground?: string
+  bottomColor?: NativeThemes
+  /** Has to be 'rgba(R, G, B[, A])' to enable, 'transparent' to disable */
```
trollepierre, line 39: Did you try to use
```
type RGB<T extends string> = 
    T extends `rgb(${infer R},${infer G},${infer B})` 
        ? '111' extends `${IsColorValue<Trim<R>>}${IsColorValue<Trim<G>>}${IsColorValue<Trim<B>>}`
            ? T
            : never
        : never
        
type RGBA<T extends string> =
    T extends `rgba(${infer R},${infer G},${infer B},${infer O})`
        ? '1111' extends `${IsColorValue<Trim<R>>}${IsColorValue<Trim<G>>}${IsColorValue<Trim<B>>}${Opacity<Trim<O>>}`
            ? T
            : never
        : never
        ```
from
https://stackoverflow.com/questions/68064202/how-to-type-a-color-prop

trollepierre (approved):

packages/cozy-intent/src/api/models/applications.ts
acezard: I'm not sure, seems a bit hacky for me (even if it works).

PR #1515, opened 2022-04-05
fix: Avoid trigger last execution "unknown" while the connector is still running

doubleface:
This is a problem especially visible with client side connectors but you can also reproduce it with node connecteur :

When you display the harvest dialog for a trigger which is still running and which runs for the first time, the trigger is marked as not running and the last execution date is "unknown"

For node connecteur, when the LOGIN_SUCCESS event is triggered, we display a popup which explains that authentication is a success and that the user can do someting else, then this bug is not too much pain

But for client side connectors, on LOGIN_SUCCESS, the harvest is displayed and the connector is displayed as not running, which is wrong and suggests an error : 

image

Now, when a still running trigger is displayed, harvest will mark it a running and watch the corresponding job to get the result of the job and display it

- feat: Do not mark a job as succeeded on login success
- feat: Watch job changes when the trigger is already running

### DIFF
--- a/packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
+++ b/packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
@@ -8,12 +8,13 @@ import {
   launchTrigger
 } from '../connections/triggers'
 import CozyRealtime from 'cozy-realtime'
-import KonnectorJobWatcher from './konnector/KonnectorJobWatcher'
+import KonnectorJobWatcher, { watchKonnectorJob } from './konnector/KonnectorJobWatcher'
 import { konnectorPolicy as biKonnectorPolicy } from '../services/budget-insight'
 import fixtures from '../../test/fixtures'
 import sentryHub from '../sentry'
 import { Q } from 'cozy-client'
 
+jest.mock('./konnector/KonnectorJobWatcher')
 jest.mock('../sentry', () => {
   const mockScope = {
     setTag: jest.fn()
@@ -116,6 +117,17 @@ describe('ConnectionFlow', () => {
     const isSubmitting = flow => {
       return flow.getState().running === true
     }
+    beforeAll(() => {
+      watchKonnectorJob.mockReturnValue({on: () => ({})})
+    })
+
+    afterEach(() => {
+      jest.clearAllMocks()
+    })
+
+    afterAll(() => {
+      jest.restoreAllMocks()
+    })
 
     it('should render as submitting when there is no account', async () => {
       const { flow } = setup()
@@ -213,6 +225,10 @@ describe('ConnectionFlow', () => {
         userCredentials: {
           login: 'foo',
           password: 'bar'
+        },
+        account: {
+          login: 'old',
+          password: 'old'
         }
       })
 
@@ -314,6 +330,7 @@ describe('ConnectionFlow', () => {
   describe('ensureTriggerAndLaunch', () => {
     beforeAll(() => {
       jest.spyOn(cronHelpers, 'fromFrequency').mockReturnValue('0 0 0 * * 0')
+      watchKonnectorJob.mockReturnValue({on: () => ({})})
     })
 
     afterEach(() => {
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -481,6 +498,24 @@ describe('ConnectionFlow', () => {
     delete window.cozy
     delete window.ReactNativeWebView
   })
+
+  describe('contructor', () => {
```
trollepierre, line 502:
```suggestion
  describe('constructor', () => {
```

packages/cozy-harvest-lib/src/models/konnector/KonnectorJobWatcher.js
```diff
@@ -66,6 +66,11 @@ export class KonnectorJobWatcher {
     this.emit('error', new KonnectorJobError(error))
   }
 
+  handleLoginSuccess() {
```
trollepierre, line 69: don't we test that part?

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -481,6 +498,24 @@ describe('ConnectionFlow', () => {
...
+    beforeAll(() => {
+      watchKonnectorJob.mockReturnValue({on: () => ({})})
+    })
+
+    afterEach(() => {
+      jest.clearAllMocks()
+    })
+
+    afterAll(() => {
+      jest.restoreAllMocks()
```
trollepierre, line 512: what is the point to use a clearAllMocks before restoreAllMocks:

https://jestjs.io/docs/mock-function-api#mockfnmockrestore

I read here that mockRestore does everything that mockReset does
And mockReset does everything that mockClear deos

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -116,6 +117,17 @@ describe('ConnectionFlow', () => {
...
```
trollepierre, line 129: what is the point to use a clearAllMocks before restoreAllMocks:

https://jestjs.io/docs/mock-function-api#mockfnmockrestore

I read here that mockRestore does everything that mockReset does
And mockReset does everything that mockClear deos

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
trollepierre: nit: i recommend to remove clearAllMocks

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
trollepierre: nit: i recommend to remove clearAllMocks

trollepierre (approved):

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
doubleface: I prefer to remove restoreAllMocks in this case. ClearAllMocks will erase the history of calls between test which is what I want in the first place.

And removing clearAllMocks makes the tests fail

trollepierre: ok, then.

ps: I need to understand one day why mocks are not reset by default. 🤯

packages/cozy-harvest-lib/src/models/konnector/KonnectorJobWatcher.js
doubleface: This would need to be an integration test. I think there is no point to add a unit test here.

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
doubleface: Agreed

PR #1516, opened 2022-04-05
Feat/add listener acquisition step

Merkur39:
We have added the possibility to validate the acquisition step with the Enter key (to be iso with the following ones)
By the way, fix non-conforming margins


Merkur39:
[UI / UX en version Desktop](https://trello.com/c/u4dBGn8r/841-ui-ux-en-version-desktop)

trollepierre (approved):
nit: maybe we could test this feature

JF-Cozy (commented):
+1 pour le test on a des exemples dans Drive

PR #1517, opened 2022-04-05
fix: Use correct type for define input rules

Merkur39:
In the configuration file, we have attribute types on the Information step `Number::12`, which are no longer valid.


Merkur39:
[Regression type d'attributs des étapes d'Informations](https://trello.com/c/3oel5lqW/875-regression-type-dattributs-des-%C3%A9tapes-dinformations)

trollepierre (approved):

PR #1519, opened 2022-04-06
feat: Change harvest sentry url to errors.cozycloud.cc

doubleface:


packages/cozy-harvest-lib/src/sentry.js
```diff
@@ -6,7 +6,7 @@ import * as Sentry from '@sentry/browser'
  * @param {SentryClient}
  */
 export const client = new Sentry.BrowserClient({
-  dsn: 'https://888abd94993a47a39e751598fc6be803@sentry.cozycloud.cc/145',
+  dsn: 'https://1f8d191bad884bf98ae8cdcb6f6abb72@errors.cozycloud.cc/39',
```
trollepierre, line 9: it is normal to set public key inside repopsitory? 


shouldn't we use at least variable?

sblaisot: all responses to your concerns are in sentry documentation: https://docs.sentry.io/product/sentry-basics/dsn-explainer/#dsn-utilization

TL;DR: No major problem having it there.

trollepierre (approved):

Crash--:
ping @doubleface

PR #1520, opened 2022-04-06
fix: Back to Scan step must be displayed file already selected

Merkur39:
When we have the option "multipage", cancel the last selected image shows directly the previous one.
From there we can continue to undo the images, select a new image or continue the process (if at least one image exists)

Once on the Acquisition page, save the chosen image directly into the FormDataProvider


Merkur39:
[Gestion du back(cozy) pendant acquisition - gestion étapes images](https://trello.com/c/E4OYO36L/762-gestion-du-backcozy-pendant-acquisition-gestion-%C3%A9tapes-images)

JF-Cozy (approved):

PR #1523, opened 2022-04-07
feat: Remove flag condition to display datacard

Crash--:
We don't need to check the flag anymore. We want
this feature for all our users.


Merkur39 (approved):
🙌

packages/cozy-harvest-lib/src/datacards/datacardOptions.js
```diff
@@ -22,8 +21,7 @@ const timeseriesGeoJSONDatacard = {
 
 const filesDatacard = {
   component: FileDataCard,
-  match: ({ trigger }) =>
-    flag('harvest.datacards.files') && get(trigger, 'message.folder_to_save')
+  match: ({ trigger }) => get(trigger, 'message.folder_to_save')
```
trollepierre, line 24: can we create a reminder to delete correctly this feature flag if used in production?

trollepierre: oops, actually this flag is also used: https://github.com/cozy/cozy-libs/blob/18a3ff09d0cbe0d854f484798b445b724179d095/packages/cozy-harvest-lib/src/models/getRelatedAppsSlugs.js#L33

trollepierre: should this cleaning, be done in getRelatedAppSlugs as well? 🤔

Crash--: I forgot to force push earlier, but this is now fixed ;)

trollepierre: can we create a reminder to delete correctly this feature flag if used in production?

trollepierre (approved):

PR #1528, opened 2022-04-10
feat: Fix back navigation

acezard:
Using replace instead of push makes things go easier when going back.
But it also means the connector view is one single route
Going back will always go back to the home

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountModal.jsx
@@ -14,7 +14,7 @@ import * as triggersModel from '../helpers/triggers'
 import KonnectorAccountTabs from './KonnectorConfiguration/KonnectorAccountTabs'
 import AccountSelectBox from './AccountSelectBox/AccountSelectBox'
 import KonnectorModalHeader from './KonnectorModalHeader'
-import { withMountPointPushHistory } from './MountPointContext'
+import { withMountPointHistory } from './MountPointContext'
 import withLocales from './hoc/withLocales'
 import DialogContent from '@material-ui/core/DialogContent'
 
@@ -109,6 +109,7 @@ export class AccountModal extends Component {
       accountsAndTriggers,
       t,
       pushHistory,
+      replaceHistory,
       initialActiveTab,
       breakpoints: { isMobile },
       showAccountSelection,
@@ -164,7 +165,7 @@ export class AccountModal extends Component {
               initialTrigger={trigger}
               account={account}
               onAccountDeleted={onDismiss}
-              addAccount={() => pushHistory('/new')}
+              addAccount={() => replaceHistory('/new')}
               showNewAccountButton={showNewAccountButton}
             />
           </DialogContent>
@@ -197,7 +198,7 @@ AccountModal.propTypes = {
     })
   ).isRequired,
   t: PropTypes.func.isRequired,
-  pushHistory: PropTypes.func.isRequired,
+  replaceHistory: PropTypes.func.isRequired,
   accountId: PropTypes.string.isRequired,
 
   /** @type {string} Can be set to force the initial active tab */
@@ -213,6 +214,6 @@ AccountModal.propTypes = {
 export default flow(
   withClient,
   withLocales,
-  withMountPointPushHistory,
+  withMountPointHistory,
   withBreakpoints()
 )(AccountModal)
--- a/packages/cozy-harvest-lib/src/components/AccountsListModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountsListModal.jsx
@@ -8,7 +8,7 @@ import { MountPointContext } from './MountPointContext'
 import DialogContent from '@material-ui/core/DialogContent'
 
 const AccountsListModal = ({ konnector, accounts, t }) => {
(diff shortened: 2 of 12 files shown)


packages/cozy-harvest-lib/src/components/MountPointContext.jsx
```diff
@@ -23,7 +23,7 @@ export class RawMountPointProvider extends React.Component {
       .split('/')
       .concat(route.split('/'))
       .filter(Boolean)
-    history.push(`/${segments.join('/')}`)
+    history.replace(`/${segments.join('/')}`)
```
Crash--, line 26: We can not have a method called `pushHistory` now that we `replaceHistory`. Can you rename it please?

acezard: > We can not have a method called `pushHistory` now that we `replaceHistory`. Can you rename it please?

yes of course, I should even create pushHistory back, don't you think?
but it will not be used by anything for now..

trollepierre: let's not create unused method. YAGNI.

acezard: Updated

packages/cozy-harvest-lib/src/components/HarvestModalRoot.jsx
```diff
@@ -4,12 +4,12 @@ import AccountsListModal from './AccountsListModal'
...
+  const { replaceHistory } = useContext(MountPointContext)
   if (accounts.length === 0) {
-    pushHistory('/new')
+    replaceHistory('/new')
     return null
   } else if (accounts.length === 1) {
-    pushHistory(`/accounts/${accounts[0].account._id}`)
+    replaceHistory(`/accounts/${accounts[0].account._id}`)
     return null
   } else {
```
Crash--, line 14: I'm totally agreed with this change. Others I don't know

packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
```diff
@@ -164,9 +164,9 @@ export class EditAccountModal extends Component {
   redirectToAccount() {
     const { account } = this.state
     if (account) {
-      this.props.pushHistory(`/accounts/${account._id}`)
+      this.props.replaceHistory(`/accounts/${account._id}`)
     } else {
-      this.props.pushHistory(`/accounts`)
+      this.props.replaceHistory(`/accounts`)
```
Crash--, line 169: I think here we should stay on pushHistory. Not replace. We edited an account, then we go to /accounts or /accounts/id. If I want to click on back, I should go to the edit modale again

packages/cozy-harvest-lib/src/components/AccountsListModal.jsx
```diff
@@ -23,8 +23,8 @@ const AccountsListModal = ({ konnector, accounts, t }) => {
         <AccountsList
           accounts={accounts}
           konnector={konnector}
-          onPick={option => pushHistory(`/accounts/${option.account._id}`)}
-          addAccount={() => pushHistory('/new')}
+          onPick={option => replaceHistory(`/accounts/${option.account._id}`)}
+          addAccount={() => replaceHistory('/new')}
```
Crash--, line 27: I don't think we want a replace here... 

I think pushHistory is the right one

packages/cozy-harvest-lib/src/components/AccountModal.jsx
```diff
@@ -124,10 +124,10 @@ export class AccountModal extends Component {
               selectedAccount={account}
               accountsAndTriggers={accountsAndTriggers}
               onChange={option => {
-                pushHistory(`/accounts/${option.account._id}`)
+                replaceHistory(`/accounts/${option.account._id}`)
```
Crash--, line 127: same here, push is the right thing to do here

packages/cozy-harvest-lib/src/components/AccountModal.jsx
```diff
@@ -124,10 +124,10 @@ export class AccountModal extends Component {
...
               }}
               onCreate={() => {
-                pushHistory('/new')
+                replaceHistory('/new')
```
Crash--, line 130: push

packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
acezard: Updated ✅

packages/cozy-harvest-lib/src/components/AccountsListModal.jsx
acezard: Updated ✅

packages/cozy-harvest-lib/src/components/AccountModal.jsx
acezard: Updated ✅

packages/cozy-harvest-lib/src/components/AccountModal.jsx
acezard: Updated ✅

Crash-- (approved):

PR #1530, opened 2022-04-12
fix(repo-doctor): Moment and Material-UI should not be dependencies

trollepierre:
According to our [guidelines](https://github.com/cozy/cozy-guidelines), material-ui and moment should not be set as dependencies.
This decision has been taken during Front meeting of 12th April

Capture d’écran 2022-04-12 à 14 34 48

### DIFF
--- a/packages/repo-doctor/examples/repo-doctor.json
+++ b/packages/repo-doctor/examples/repo-doctor.json
@@ -50,6 +50,10 @@
           "reason": "Used through cozy-scripts"
         },
         {
+          "name": "material-ui",
+          "reason": "Used through cozy-ui"
+        },
+        {
           "name": "moment",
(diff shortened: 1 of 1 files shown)


Crash-- (changes requested):
I don't understand. Material-ui is declarated as a peerDep of cozy-ui. So the app needs to have material-ui as dep... 

This is inside our libs that material-ui should not be added as dep

trollepierre:
@Crash-- : 
let's fix this LGTM issue : https://github.com/cozy/cozy-ui/issues/1865
if you approve that https://github.com/cozy/cozy-ui/pull/2094
that would unblock your requested change.

acezard (approved):

Crash-- (approved):

PR #1531, opened 2022-04-12
fix: Get correct bi mapping for bnp_es and cic_es

doubleface:
Upgrade cozy-bi-auth package to 0.0.25. 0.0.24 was published without
building the package

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7927,10 +7927,10 @@ cozy-bar@7.16.0:
     redux-thunk "2.3.0"
     semver-compare "^1.0.0"
 
-cozy-bi-auth@0.0.24:
-  version "0.0.24"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -18530,7 +18530,7 @@ react-router@5.2.0:
     tiny-invariant "^1.0.2"
     tiny-warning "^1.0.0"
 
-react-router@6, react-router@6.3.0:
```
Crash--, line 18533: I don't understand why you have this change 🤔

doubleface: Did not see that but I actually get the same change if I take the master and  run `yarn lerna clean && yarn install` with yarn 1.22.18 and node 12.22.7

trollepierre: this is a nit.

This  PR can be approved

trollepierre (approved):

PR #1532, opened 2022-04-13
Feat/viewer route

acezard:
https://trello.com/c/Ge6JSeCT/377-harvest-gestion-dune-route-pour-la-visionneuse-de-fichier

### Features

- [x] Adding a new route in cozy-harvest for the file viewer, allowing better navigation
- [x] It is now possible to navigate between files and back/forward with browser history

### Implementation

- [x] Remove Viewer from FileDataCard
- [x] Create Route
- [x] Create a new hook to get files data
- [x] Create a new component to display this data
- [x] Refactor old FileDataCard to use the new hook

### Caveats

When leaving the viewer with the "goback" arrow, we replace history. But that means we're still 2 back aways from returning to the home from there.

### DIFF
--- a/packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
+++ b/packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
@@ -1,8 +1,6 @@
-import React, { useState, useMemo } from 'react'
+import React, { useContext, useState } from 'react'
 import PropTypes from 'prop-types'
 import get from 'lodash/get'
-import sortBy from 'lodash/sortBy'
-import uniq from 'lodash/uniq'
 import keyBy from 'lodash/keyBy'
 
 import 'leaflet/dist/leaflet.css'
@@ -12,15 +10,13 @@ import List from '@material-ui/core/List'
 import ListItem from '@material-ui/core/ListItem'
 import ListItemIcon from '@material-ui/core/ListItemIcon'
 import Slide from '@material-ui/core/Slide'
-import Modal from '@material-ui/core/Modal'
 
 import ListItemText from 'cozy-ui/transpiled/react/ListItemText'
 
+import { RealTimeQueries } from 'cozy-client'
 import palette from 'cozy-ui/transpiled/react/palette'
 import { Media, Bd, Img } from 'cozy-ui/transpiled/react/Media'
 import Circle from 'cozy-ui/transpiled/react/Circle'
-import Portal from 'cozy-ui/transpiled/react/Portal'
-import Viewer from 'cozy-ui/transpiled/react/Viewer'
 import Card from 'cozy-ui/transpiled/react/Card'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import FileIcon from 'cozy-ui/transpiled/react/Icons/File'
@@ -29,16 +25,9 @@ import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 
 import AppLinkCard, { AppLinkButton } from '../components/cards/AppLinkCard'
 import appLinksProps from '../components/KonnectorConfiguration/DataTab/appLinksProps'
-
-import CozyClient, {
-  Q,
-  queryConnect,
-  isQueryLoading,
-  hasQueryBeenLoaded,
-  RealTimeQueries
-} from 'cozy-client'
-
+import { MountPointContext } from '../components/MountPointContext'
 import { getFileIcon } from './mime-utils'
+import { useDataCardFiles } from './useDataCardFiles'
 
 const LoadingFileListItem = ({ divider }) => {
   return (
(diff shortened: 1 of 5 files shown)


packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
```diff
@@ -169,60 +146,11 @@ const FileCard = ({ files, loading, konnector, trigger }) => {
...
-  const noFiles =
-    hasQueryBeenLoaded(folderToSaveFiles) &&
-    files1.length === 0 &&
-    hasQueryBeenLoaded(sourceAccountFiles) &&
-    files2.length === 0
-  const isLoading =
-    isQueryLoading(folderToSaveFiles) || isQueryLoading(sourceAccountFiles)
+const FileDataCard = ({ accountId, konnector, trigger }) => {
+  const files = useDataCardFiles(accountId, trigger.id)
+  const isLoading = !files
```
Crash--, line 151: This line sounds wrong to me. This is not because we don't have `files` that we're still 'loading'. It creates an indirection between two different things.

packages/cozy-harvest-lib/src/datacards/ViewerModal.jsx
```diff
@@ -0,0 +1,33 @@
+import React, { useContext } from 'react'
+import Modal from '@material-ui/core/Modal'
```
Crash--, line 2: We should not import Modal from m-ui. You should use our CozyDialog from cozy-ui (cc @JF-Cozy )

packages/cozy-harvest-lib/src/datacards/ViewerModal.jsx
```diff
@@ -0,0 +1,33 @@
...
+      <Modal open={true}>
+        <Viewer
+          files={files}
+          currentIndex={Number(fileIndex)}
+          onCloseRequest={handleCloseViewer}
+          onChangeRequest={handleFileChange}
+        />
+      </Modal>
+    </Portal>
+  ) : null
```
Crash--, line 32: Shouldn't we display an error message or something if there is no files to display ? What is the current behavior with this implementation?

packages/cozy-harvest-lib/src/datacards/makeViewerFiles.js
```diff
@@ -0,0 +1,83 @@
...
+
+  return fetchStatus === 'loaded' ? data?.[0]?.message?.folder_to_save : null
+}
+
+const handleReturn = ({ data, fetchStatus }) =>
+  fetchStatus === 'loaded' && data.length === 0
+    ? []
+    : fetchStatus !== 'pending' && data.length > 0
+    ? data
+    : null
```
Crash--, line 22: I don't think we need this method. Cozy-Client already return an empty array if `data.length === 0`. 

I prefer if we avoid to edit the return of cozy-client. 

What is the gain to return null? Why not dealing with fetchStatus within the component?

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
acezard: With my new implementation, null = loading, empty array = loaded but empty. Do you prefer clearer return interface?

Crash--: I prefer to use Cozy-Client mechanism: fetchStatus. How do you know if it's loading or if there is an error somewhere for instance ?

packages/cozy-harvest-lib/src/datacards/ViewerModal.jsx
acezard: It was used in the FileDataCard component I thought it was intended, I'm updating and checking if any error

packages/cozy-harvest-lib/src/datacards/ViewerModal.jsx
acezard: Well, there's none, to be honest, but here it's more a case of I don't know what should I do in terms of behavior. What do we want?

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
acezard: - [ ] Will use `type QueryFetchStatus = "failed" | "loading" | "pending" | "loaded";` along with eventual data

packages/cozy-harvest-lib/src/datacards/ViewerModal.jsx
acezard: - [ ] Replace with cozy-ui

packages/cozy-harvest-lib/src/datacards/makeViewerFiles.js
acezard: - [ ] Will use `type QueryFetchStatus = "failed" | "loading" | "pending" | "loaded";` along with eventual data

packages/cozy-harvest-lib/src/datacards/FileDataCard.jsx
Crash--: Don't forget our internal conversation about the `loading` state ;)

Crash-- (approved):

PR #1536, opened 2022-04-14
Fix/composite header icon prop

Merkur39:
The "icon" property can be undefined.
Therefore, you must make sure that it is defined before any processing.

I also take the opportunity to make a refactor :)

### DIFF
--- a/packages/cozy-mespapiers-lib/src/components/CompositeHeader/CompositeHeaderImage.jsx
+++ b/packages/cozy-mespapiers-lib/src/components/CompositeHeader/CompositeHeaderImage.jsx
@@ -1,0 +1,101 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+import { makeStyles } from '@material-ui/core/styles'
+
+import Icon, { iconPropType } from 'cozy-ui/transpiled/react/Icon'
+
+import IlluCovidVaccineCertificate from '../../assets/images/IlluCovidVaccineCertificate.png'
+import IlluDiploma from '../../assets/images/IlluDiploma.png'
+import IlluDriverLicenseBack from '../../assets/images/IlluDriverLicenseBack.png'
+import IlluDriverLicenseFront from '../../assets/images/IlluDriverLicenseFront.png'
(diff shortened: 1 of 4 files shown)


packages/cozy-mespapiers-lib/src/components/CompositeHeader/CompositeHeaderImage.spec.js
```diff
@@ -0,0 +1,45 @@
...
+      text="Text"
+    />
+  )
+}
+
+describe('CompositeHeaderImage', () => {
+  it('should return nothing if icon & fallbackIcon props is undefined', () => {
+    const { container } = setup()
+
+    expect(container).toMatchInlineSnapshot('<div />')
```
trollepierre, line 22: nit: don't we use https://github.com/testing-library/jest-dom ?

to be honest, that test would be better with a https://github.com/testing-library/jest-dom#tobeemptydomelement=
```suggestion
    expect(container).toBeEmptyDOMElement()
```

packages/cozy-mespapiers-lib/src/components/CompositeHeader/CompositeHeaderImage.jsx
```diff
@@ -0,0 +1,101 @@
...
+  if (!icon && !fallbackIcon) {
+    return null
+  }
+
+  const iconName = icon && icon.split('.')[0]
+  const src = images[iconName] || fallbackIcon
+  const isBitmap = typeof src === 'string' && src.endsWith('.png')
+
+  if (isBitmap) {
+    return <img data-testid={src} src={src} alt="illustration" />
```
trollepierre, line 80: nit: should we test that part?

trollepierre (approved):
good job in LGTM 🎉

packages/cozy-mespapiers-lib/src/components/CompositeHeader/CompositeHeaderImage.spec.js
Merkur39: En effet, merci j'étais passé à côté 👍

Merkur39:
[Certaines qualifications de documents font crash l'app](https://trello.com/c/1qg6lSed/881-certaines-qualifications-de-documents-font-crash-lapp)

PR #1537, opened 2022-04-14
fix: Avoid to have not sanitized fields as identifier fields

doubleface:
When some connectors have unusual fields, like ovh, this can lead to
have "advancedFields" as identifier field

This is just a protection. The best fix beeing the connector choosing
itself the the best identifer field by adding `role: "identifier"` to
the corresponding field


trollepierre:
what does 
> When some connectors havest unusual fields
mean?

havest? harvest? have?

Crash-- (approved):

PR #1538, opened 2022-04-15
fix(eslint-config): Add rule for Windows user

acezard:
Windows machine endOfLine can be "crlf". Git will handle it.
Also ordered alphabetically the configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ before_deploy:
   - git config --global user.name "Cozy Bot"
 deploy:
   - provider: script
-    skip-cleanup: true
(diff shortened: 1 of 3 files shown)


trollepierre (approved):

.travis.yml
```diff
@@ -15,7 +15,7 @@ before_deploy:
   - git config --global user.name "Cozy Bot"
 deploy:
   - provider: script
-    skip-cleanup: true
+    cleanup: true
```
trollepierre, line 18: i don't find, what is cleanup ????

https://docs.travis-ci.com/user/deployment/#uploading-files-and-skip_cleanup

acezard: travis CLI was printing a warning https://app.travis-ci.com/github/cozy/cozy-libs/jobs/566061267/config

cf https://blog.travis-ci.com/2019-08-27-deployment-tooling-dpl-v2-preview-release

skip-cleanup is now default, so we should just delete skip-cleanup and not replace it with cleanup like I did

trollepierre: ok, i prefer last version

PR #1543, opened 2022-04-22
fix: Add skip-cleanup to Travis config

Merkur39:
Deleting skip-cleanup in https://github.com/cozy/cozy-libs/commit/51ae10dd3411ef0ff8c8434d8d5d02d31ac44722 would make the CI fail when executing the deploy script (i.e. when merging on master)
The observed error is that lerna command is not recognized anymore
and suggests that the node_modules folder doesn't exist anymore
and so skip-cleanup's default behavior doesn't seem
to be applied as described in the travis documentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ before_deploy:
   - git config --global user.name "Cozy Bot"
 deploy:
   - provider: script
+  - skip-cleanup: true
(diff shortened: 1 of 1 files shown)


Ldoppea (approved):
Can you reword the commit as this? This gives more details about what error we fix:

> Deleting `skip-cleanup` in https://github.com/cozy/cozy-libs/commit/51ae10dd3411ef0ff8c8434d8d5d02d31ac44722 would make the CI fail when executing the `deploy` script (i.e. when merging on `master`)
> The observed error is that `lerna` command is not recognized anymore and suggests that the `node_modules` folder doesn't exist anymore and so `skip-cleanup`'s default behavior doesn't seem to be applied as described in documentations

PR #1552, opened 2022-04-26
feat: Rework error handling flow

acezard:
- This will handle scenarios that were unknown to this day (url casing)
- Updated tests
- Updated typings
- Replaced throw statements with debug statements
- Added more typeguards
- Removed unused strings

https://trello.com/c/yrgEYA6R/398-%F0%9F%90%9B-android-bloqu%C3%A9-sur-le-nuage

### DIFF
--- a/packages/cozy-intent/src/api/services/NativeMessenger.spec.ts
+++ b/packages/cozy-intent/src/api/services/NativeMessenger.spec.ts
@@ -1,5 +1,15 @@
-import { NativeMessenger, PostMeMessage, strings } from '../../api'
+import { NativeMessenger, PostMeMessage } from '../../api'
 import { mockWebviewRef } from '../../../tests'
+
+const mockDebug = jest.fn()
+
+jest.mock('post-me', () => ({
+  ...jest.requireActual('post-me'),
+  debug:
+    (nameSpace = 'NativeService') =>
+    (): unknown =>
+      mockDebug(nameSpace)
+}))
 
 describe('NativeMessenger', () => {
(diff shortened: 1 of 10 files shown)


packages/cozy-intent/src/api/services/NativeMessenger.spec.ts
```diff
@@ -79,6 +91,8 @@ describe('NativeMessenger', () => {
         type: 'string',
         uri: 'string'
       })
-    ).toThrowError(strings.noListenerFound)
+    ).not.toThrow()
+
+    expect(mockDebug).toHaveBeenCalled()
```
trollepierre, line 96: nit: why not using toHaveBeenCalledWith() ?

packages/cozy-intent/src/utils/index.ts
```diff
@@ -36,3 +36,8 @@ export const isWebDevMode = (): boolean => {
     return false
   }
 }
+
+export const getErrorMessage = (error: unknown): string => {
+  if (error instanceof Error) return error.message
+  return String(error)
```
trollepierre, line 42: nit: let's test this method

trollepierre (approved):

packages/cozy-intent/src/api/services/NativeMessenger.spec.ts
acezard: @trollepierre I preferred to not check the argument in an effort to stay decoupled from the implementation

PR #1556, opened 2022-04-27
fix: Check window existence in cozy-device-helper

doubleface:
Do not suppose we are in browser environment.
Since this library is embedded in cozy-client which is embedded in
cozy-konnector-libs, connectors will fail in node environments.


Crash-- (changes requested):
I don't think this is the right fix. 

Let's discuss it internally.

doubleface:
fix replaced by https://github.com/cozy/cozy-libs/pull/1560

PR #1561, opened 2022-04-28
feat: Open InAppBrowser via cozy-intent when in flaghip app

doubleface:
Now, when in flagship app, for OAuth connectors, harvest will open oauth dance window in an InAppBrowser instead of trying to open a popup

This needs implementation of showInAppBrowser and closeInAppBrowser
methods in flagship app in https://github.com/cozy/cozy-react-native/pull/179


packages/cozy-harvest-lib/package.json
```diff
@@ -55,6 +55,7 @@
     "cozy-client": "27.17.0",
     "cozy-device-helper": "^1.18.0",
     "cozy-flags": "^2.8.7",
+    "cozy-intent": "^1.14.1",
```
Crash--, line 58: Should also be added to peer-dep

packages/cozy-harvest-lib/src/components/InAppBrowser.jsx
```diff
@@ -0,0 +1,33 @@
...
+    return function cleanup() {
+      webviewIntent.call('closeInAppBrowser')
+    }
+  }, [webviewIntent, url, onClose])
+  return null
+}
+
+InAppBrowser.propTypes = {
+  url: PropTypes.string.isRequired,
+  onClose: PropTypes.func
```
Crash--, line 30: this prop is required since we call it without any check

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -344,7 +345,7 @@ export class DumbTriggerManager extends Component {
     }
 
     return (
-      <>
+      <WebviewIntentProvider>
```
Crash--, line 348: Is it mandatory? Consuming App should already have a `WebViewIntentProvider`.

packages/cozy-harvest-lib/src/components/InAppBrowser.jsx
doubleface: let's add a check then

Crash-- (approved):

Crash-- (approved):

Crash--:
@doubleface can you upgrade to the latest cozy-intent version ? https://github.com/cozy/cozy-libs/blob/master/packages/cozy-intent/CHANGELOG.md#1171-2022-04-28

doubleface:
> can you upgrade to the latest cozy-intent version ? [https://github.com/cozy/cozy-libs/blob/master/packages/cozy-intent/CHANGELOG.md#1171-2022-04-28](https://github.com/cozy/cozy-libs/blob/master/packages/cozy-intent/CHANGELOG.md?rgh-link-date=2022-05-02T06%3A51%3A07Z#1171-2022-04-28)

The yarn.lock file already had the right version but here it is in package.json

PR #1576, opened 2022-05-04
fix(cozy-intent): Allow to register webviews with baseUrl and no uri

Ldoppea:
ReactNative webviews can be registered using two forms of sources:
- `source={{ uri }}`
- `source={{ html, baseUrl }}`

Previous code implementation was able to support only the first form
and would crash when a webview is registered with a baseUrl

Not that react-native-webview priorize check on html+baseUrl over uri,
so we chose to implement only `isWebviewSourceBaseUrl` to prevent a
wrong check

More details:
- https://github.com/react-native-webview/react-native-webview/blob/a36394f598944c1ffe288f739b7441830249dc1e/apple/RNCWebView.m#L679-L686
- https://github.com/react-native-webview/react-native-webview/blob/a36394f598944c1ffe288f739b7441830249dc1e/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L513-L518

### DIFF
--- a/packages/cozy-intent/src/api/models/environments.ts
+++ b/packages/cozy-intent/src/api/models/environments.ts
@@ -9,22 +9,19 @@ export interface WebviewSourceUri {
   html?: never
 }
 
-export interface WebiewSourceBaseUrl {
+export interface WebviewSourceBaseUrl {
   baseUrl: string
   html: string
 
   uri?: never
 }
 
-export type WebviewSource = WebviewSourceUri | WebiewSourceBaseUrl
+export type WebviewSource = WebviewSourceUri | WebviewSourceBaseUrl
 
 export function isWebviewSourceBaseUrl(
   webviewSource: WebviewSource
-): webviewSource is WebiewSourceBaseUrl {
-  if ((webviewSource as WebiewSourceBaseUrl).html) {
-    return true
-  }
-  return false
+): webviewSource is WebviewSourceBaseUrl {
+  return Boolean((webviewSource as WebviewSourceBaseUrl).html)
 }
 
 export interface WebviewRef {
--- a/packages/cozy-intent/src/api/models/environments.ts
+++ b/packages/cozy-intent/src/api/models/environments.ts
@@ -2,12 +2,35 @@ import { Connection, EventsType, MethodsType } from 'post-me'
 
 import { NativeMethodsRegister } from '../../api'
 
+export interface WebviewSourceUri {
+  uri: string
+
+  baseUrl?: never
+  html?: never
+}
+
+export interface WebiewSourceBaseUrl {
+  baseUrl: string
+  html: string
+
+  uri?: never
+}
+
+export type WebviewSource = WebviewSourceUri | WebiewSourceBaseUrl
+
+export function isWebviewSourceBaseUrl(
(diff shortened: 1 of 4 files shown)


packages/cozy-intent/src/api/models/environments.ts
```diff
@@ -2,12 +2,35 @@ import { Connection, EventsType, MethodsType } from 'post-me'
 
 import { NativeMethodsRegister } from '../../api'
 
+export interface WebviewSourceUri {
+  uri: string
+
+  baseUrl?: never
+  html?: never
```
trollepierre, line 9: I don't understand that type. 

That is the first time I met the never, and this usage is not documented in the doc
https://www.tutorialsteacher.com/typescript/typescript-never

packages/cozy-intent/src/api/models/environments.ts
```diff
@@ -2,12 +2,35 @@ import { Connection, EventsType, MethodsType } from 'post-me'
...
+}
+
+export interface WebiewSourceBaseUrl {
```
trollepierre, line 12:
```suggestion
export interface WebViewSourceBaseUrl {
```

packages/cozy-intent/src/api/models/environments.ts
```diff
@@ -2,12 +2,35 @@ import { Connection, EventsType, MethodsType } from 'post-me'
 
 import { NativeMethodsRegister } from '../../api'
 
+export interface WebviewSourceUri {
```
trollepierre, line 5:
```suggestion
export interface WebViewSourceUri {
```

packages/cozy-intent/src/api/models/environments.ts
Ldoppea: Woups 😅 

Thanks for the edit, but the correct form should be `Webview` and not `WebView` as the others types from this file are using this form.

packages/cozy-intent/src/api/models/environments.ts
Ldoppea: `Never` means that those props can never have a value.

This notation allows to mutually exclude `WebviewSourceUri` and `WebviewSourceBaseUrl` when doing a union on those types.

You can set
```ts
const source: WebviewSource = { uri: 'https://foo.bar' }
```
or
```ts
const source: WebviewSource = { html: 'SOME_HTML', baseUrl: 'https://foo.bar' }
```
but the compiler will throw if you set

```ts
const source: WebviewSource = { uri: 'https://foo.bar', html: 'SOME_HTML', baseUrl: 'https://foo.bar' }
```
Because no type can have this combination (which is prevented by the `never` notation).

trollepierre: ok i got it!
Capture d’écran 2022-05-05 à 10 37 43

trollepierre (approved):
@Ldoppea : don't forget to review my commit, after yours.

Do you want me to squash the commits?

Ldoppea:
Thanks, looks good to me (I can't add a review for my own PR 😅 )

I'm not sure if we need to squash them. This is the same feature, but your commit also contain some refactoring that would add a bit of noise to the main commit. So my answer is: do as you prefer 😛

PR #1579, opened 2022-05-05
Select current user by default on contact step

Merkur39:
When creating paper, the current user should also be selected by default when multiple selection is enabled at the contact step.


JF-Cozy (approved):

Merkur39:
[10. Attribution multi-profil des papiers (0.75J -> 1.5J)](https://trello.com/c/JXlMDkGd/873-10-attribution-multi-profil-des-papiers-075j-15j)

PR #1582, opened 2022-05-06
chore: Build cozy-device-helper first because of dep in cozy-intent

JF-Cozy:
Since cozy-intent needs a build version of cozy-device-helper to be
built, we update the build command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "lint": "eslint --ext js,jsx,ts,tsx .",
     "lint:md": "remark . -o",
     "test": "lerna run --concurrency 1 test",
-    "build": "lerna run --parallel build",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "lint": "eslint --ext js,jsx,ts,tsx .",
     "lint:md": "remark . -o",
     "test": "lerna run --concurrency 1 test",
-    "build": "lerna run --parallel build",
+    "build": "lerna run --scope cozy-device-helper build && lerna run --parallel --ignore cozy-device-helper build",
```
trollepierre, line 21: I am wondering what have changed inside cozy-intent to get this error. We need to investigate that point one day. Do you think it diserve a Github issue? 🤔

trollepierre (approved):

package.json
JF-Cozy: pourquoi pas oui 👍

PR #1594, opened 2022-05-10
fix: Temp commit

acezard:


packages/cozy-intent/src/api/constants/numbers.ts
```diff
@@ -1,4 +1,3 @@
 export const numbers = {
-  attemptsInterval: 100,
```
acezard, line 2: Was not used

packages/cozy-intent/src/api/services/NativeMessenger.ts
```diff
@@ -19,8 +19,12 @@ export class NativeMessenger implements Messenger {
   }
 
   public postMessage = (message: Record<string, unknown>): void => {
-    const script = `window.postMessage(${JSON.stringify(message)})`
-    this.injectJavaScript?.(script)
+    try {
+      const script = `window.postMessage(${JSON.stringify(message)})`
+      this.injectJavaScript?.(script)
+    } catch (error) {
+      log(strings.noWebviewFound)
```
acezard, line 26: Only happens when going back to home as of now, no breaking impact.
Post-me tries to respond to backToHome call with the cozyapp webview, but that webview doesn't exist anymore when it does that

packages/cozy-intent/src/api/services/NativeService.ts
```diff
@@ -56,14 +45,13 @@ export class NativeService {
   private isInitMessage = (message: PostMeMessage): boolean =>
     message.message === strings.webviewIsRendered
 
-  public registerWebview = (webviewRef: WebviewRef): void => {
-    const uri = this.getUri(webviewRef)
+  public registerWebview = (uri: string, ref: WebviewRef): void => {
```
acezard, line 48: new API

packages/cozy-intent/src/api/services/NativeService.ts
```diff
@@ -73,21 +61,25 @@ export class NativeService {
           : messenger
       }
     }
+
+    log(strings.logging.registered(uri))
   }
 
-  public unregisterWebview = (webviewRef: WebviewRef): void => {
-    const uri = this.getUri(webviewRef)
+  public unregisterWebview = (uri: string): void => {
```
acezard, line 68: new API

packages/cozy-intent/src/api/services/NativeService.ts
trollepierre: this introduces a breaking change 😢 

why not using either

```suggestion
  public registerWebview = ({uri, ref}): void => {
```
or
```suggestion
  public registerWebview = (ref: WebviewRef, uri: string): void => {
  if(!uri) {
    // should do something
  }
```

packages/cozy-intent/src/api/services/NativeService.ts
trollepierre: that is breaking change

trollepierre (changes requested):
we need to specify BREAKING CHANGE in the commit

packages/cozy-intent/src/api/services/NativeService.ts
acezard: I think it's better this way because uri is always the first argument. Also these methods are only called once in the entire cozy code right now so it's ok I think

packages/cozy-intent/src/api/services/NativeService.ts
acezard: I notified this in the PR. Won't this trigger the semver? I thought the PR description was enough to trigger a semver major update

trollepierre: Semantic release uses every commits, especially each commit description to detect breaking change.

The content of PR description is not added inside the git history, that's why setting breaking change in PR won't provoke a breaking change.

acezard: updated

trollepierre (approved):

PR #1602, opened 2022-05-11
fix(harvest-lib): Set cozy libraries as peerDependencies

trollepierre:
BREAKING CHANGE: App needs to install those 4 libraries to run correctly
`yarn add cozy-logger@>=1.9.0 @cozy/minilog@>=1.0.0 cozy-bi-auth@>=0.0.25 cozy-doctypes@>=1.83.7`


JF-Cozy (approved):
On peut peut-être indiquer dans le BC la version pour chaque package, comme indiqué dans les peerDep à savoir `>=x.y.z`

trollepierre:
@Crash-- : do you think it's better to wait (until flagship release for instance) before merging this breaking change?

doubleface:
@trollepierre Why do we have to do this change ?

trollepierre:
> @trollepierre Why do we have to do this change ?

Because it's a guideline we need to follow : 
https://github.com/cozy/cozy-guidelines#dependencies=

When I look to the code of harvest-lib for another issue, I thought, why not follow the boy scout rule.

PR #1605, opened 2022-05-12
fix: Add cozy-device-helper as real dependency

acezard:
https://app.travis-ci.com/github/cozy/cozy-libs/builds/250570179
When it's a dev-dependency, the publish will fail.
Lerna will not be able to access the typings.
As a real dep, it should be able to.

### DIFF
--- a/packages/cozy-intent/package.json
+++ b/packages/cozy-intent/package.json
@@ -10,6 +10,7 @@
     "url": "https://github.com/cozy/cozy-libs/issues"
   },
   "dependencies": {
+    "cozy-device-helper": "^2.1.0",
     "post-me": "0.4.5"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -15855,17 +15855,6 @@ msgpack5@^4.0.2:
     readable-stream "^2.3.6"
     safe-buffer "^5.1.2"
 
-"mui-bottom-sheet@git+https://github.com/cozy/mui-bottom-sheet.git#v1.0.6":
```
acezard, line 15858: I don't know why yarn changed the lockfile..?

packages/cozy-intent/package.json
```diff
@@ -10,6 +10,7 @@
     "url": "https://github.com/cozy/cozy-libs/issues"
   },
   "dependencies": {
+    "cozy-device-helper": "^2.1.0",
```
trollepierre, line 13: could you provide a failing job in the PR description?

if we keep this change, could we remove the devDependencies?

acezard: updated

trollepierre (approved):
I approve, in order not to block. 
however we should investigate why it used to work before.


i recommend `yarn why cozy-device-helper` in travis, yarn install --frozen-lockfile

PR #1637, opened 2022-06-01
Fix/multi selection forwarded file

Merkur39:
The `forwardFile` function is no longer called in the real time callback, but in a function called later (and in another component), so it no longer has the file information (so can't generate a share link)

So we pass the file to the component that will share it


Merkur39:
[5b. Multiselect - construction du Zip et transfert (0.5 => 1.5)](https://trello.com/c/rUnTKjwK/868-5b-multiselect-construction-du-zip-et-transfert-05-15)

JF-Cozy (approved):

PR #1650, opened 2022-06-06
fix: Transfer the file directly if only one

Merkur39:
In multiple selection, if we have only one file selected, no need to create a ZIP file.


Merkur39:
[Transfert multi-sélection](https://trello.com/c/oAHdJoFu/909-transfert-multi-s%C3%A9lection)

JF-Cozy (approved):

PR #1667, opened 2022-06-12
feat: Update french locale

acezard:
For OAUTH apps the wording has to be changed, per the spec
Only account creation is concerned


packages/cozy-harvest-lib/src/locales/fr.json
```diff
@@ -390,7 +390,7 @@
       "label": "Se reconnecter"
     },
     "connect": {
-      "label": "Connecter"
+      "label": "Ajouter un compte"
```
Crash--, line 393: cc @doubleface

PR #1670, opened 2022-06-13
fix: Do not try to display Contract without konnector available

doubleface:
Causes some errors in banks unit tests and in browser console.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/DisconnectedAccountModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/DisconnectedAccountModal.jsx
@@ -49,6 +49,11 @@ const DisconnectedModal = ({ accounts, onClose, initialActiveTab }) => {
   const [activeTab, setActiveTab] = useState(initialActiveTab)
 
   const { dialogProps } = useCozyDialog({ onClose, open: true })
+
+  if (!konnectorRef.current) {
+    return null
+  }
+
   return (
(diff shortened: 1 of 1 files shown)


packages/cozy-harvest-lib/src/components/DisconnectedAccountModal.jsx
```diff
@@ -69,9 +69,12 @@ const DisconnectedModal = ({ accounts, onClose, initialActiveTab }) => {
       ) : null}
       {activeTab === 1 ? (
         <>
-          {accounts.length ? (
+          {accounts.length && konnectorRef.current ? (
```
Crash--, line 72: Do we really want to display "No Contract" if the "konnectorRef" is not there? 

I think we should handle that higher in the component. Here we already pass konnector.current without any check https://github.com/cozy/cozy-libs/blob/1e4e4ccd72af2b33e7a9deb66efbe52006cf6ee7/packages/cozy-harvest-lib/src/components/DisconnectedAccountModal.jsx#L55 

What do you think about having: 
```
render(){
if(!konnectorRef.current){
return null // or <Loader />
}else{
...
}

doubleface: Right that's better

Crash-- (approved):

PR #1695, opened 2022-06-21
fix(mespapiers-lib): Allow masked inputs to be optional

PolariTOON:
Masked inputs can now be optional, and then any value that matches the start of the mask can be entered.

Also for any optional input, the empty string is now always valid, even if the length does not match min and max length constraints.

This matches the behavior of HTML `<input type="text" />` elements.

This also removes the behavior that swapped `minLength` and `maxLength` when set in the wrong order.

### DIFF
--- a/packages/cozy-mespapiers-lib/src/components/ModelSteps/widgets/InputTextAdapter.jsx
+++ b/packages/cozy-mespapiers-lib/src/components/ModelSteps/widgets/InputTextAdapter.jsx
@@ -31,14 +31,7 @@ const InputTextAdapter = ({
   setIsFocus,
   idx
 }) => {
-  const {
-    name,
-    inputLabel,
-    mask,
-    maskPlaceholder = 'ˍ',
-    maxLength,
-    minLength
-  } = attrs
+  const { name, inputLabel, mask, maskPlaceholder = 'ˍ' } = attrs
   const { t } = useI18n()
   const [currentValue, setCurrentValue] = useState(defaultValue || '')
   const [isTooShort, setIsTooShort] = useState(false)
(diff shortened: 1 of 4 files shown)


PolariTOON:
[L'inputMask ne doit pas rendre l'input nécessairement obligatoire (0.25J)](https://trello.com/c/i5X6FBx7/898-linputmask-ne-doit-pas-rendre-linput-n%C3%A9cessairement-obligatoire-025j)

trollepierre:
Why is the TravisCI Pull Request job not executed?

on another PR launched 3 hours before, got the job done, https://github.com/cozy/cozy-libs/pull/1691

idea: `git commit --amend && git push --force-with-lease`

packages/cozy-mespapiers-lib/src/components/ModelSteps/widgets/InputTextAdapter.jsx
```diff
@@ -160,8 +160,8 @@ const InputTextAdapter = ({
       helperText={helperText}
       value={currentValue}
       inputProps={{
-        maxLength: maxLength,
-        minLength: minLength,
+        maxLength: maxLength >= 0 ? maxLength : undefined,
```
trollepierre, line 163:
```suggestion
        maxLength: maxLength >= 0 && maxLength,
```

packages/cozy-mespapiers-lib/src/components/ModelSteps/widgets/InputTextAdapter.jsx
```diff
@@ -160,8 +160,8 @@ const InputTextAdapter = ({
...
+        minLength: minLength >= 0 ? minLength : undefined,
```
trollepierre, line 164:
```suggestion
        minLength: minLength >= 0 && minLength,
```

trollepierre (commented):
quick code syntax improvement ^^

packages/cozy-mespapiers-lib/src/components/ModelSteps/widgets/InputTextAdapter.jsx
Crash--: There is a diff, before we were returning `undefined`, with your suggested change we now return `false`

PR #1701, opened 2022-06-24
fix(mes-papiers): Fix typos in README.md

trollepierre:
This commit should be theorically in a docs scoped commit.
Currently upgrading cozy-mespapiers-lib is required to validate
Renovate behaviour in our application.


Merkur39 (approved):
👌

PR #1702, opened 2022-06-24
Fix/tests cozy harvest lib

Merkur39:
Do not ignore tests on cozy-harvest-lib anymore

### DIFF
--- a/packages/cozy-flags/package.json
+++ b/packages/cozy-flags/package.json
@@ -26,7 +26,7 @@
     "@babel/cli": "7.16.8",
     "@testing-library/react-hooks": "3.2.1",
     "babel-preset-cozy-app": "^2.0.2",
-    "cozy-client": "22.2.0",
+    "cozy-client": "^32.2.0",
     "jest-localstorage-mock": "2.4.21",
     "react": "16.12.0"
   },
--- a/packages/cozy-harvest-lib/src/components/InAppBrowser.jsx
+++ b/packages/cozy-harvest-lib/src/components/InAppBrowser.jsx
@@ -56,7 +56,7 @@ const InAppBrowser = ({ url, onClose, intentsApi = {} }) => {
     return function cleanup() {
       closeInAppBrowser()
     }
-  }, [isReady, url, onClose])
+  }, [isReady, url, onClose, closeInAppBrowser, fetchSessionCode, showInAppBrowser, tokenParamName])
   return null
 }
 
--- a/packages/cozy-harvest-lib/src/components/InAppBrowser.jsx
+++ b/packages/cozy-harvest-lib/src/components/InAppBrowser.jsx
@@ -56,7 +56,15 @@ const InAppBrowser = ({ url, onClose, intentsApi = {} }) => {
     return function cleanup() {
       closeInAppBrowser()
     }
-  }, [isReady, url, onClose, closeInAppBrowser, fetchSessionCode, showInAppBrowser, tokenParamName])
+  }, [
+    isReady,
+    url,
+    onClose,
+    closeInAppBrowser,
+    fetchSessionCode,
+    showInAppBrowser,
+    tokenParamName
+  ])
   return null
 }
 
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/ContractItemSecondaryText.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/ContractItemSecondaryText.jsx
@@ -10,7 +10,6 @@ import UnlinkIcon from 'cozy-ui/transpiled/react/Icons/Unlink'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import SyncIcon from 'cozy-ui/transpiled/react/Icons/Sync'
 
-import { CollectionPropType } from './Contracts'
 import { isDeleted, isDisabled, isErrored, isImported } from './helpers'
 
 const dateFnsLocales = {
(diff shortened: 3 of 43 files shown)


packages/cozy-sharing/src/SharingBanner/test/RouterContext.jsx
```diff
@@ -1,17 +1,18 @@
...
+export const RouterContextProvider = ({ children }) => {
+  const location = useLocation()
+  const params = useParams()
+
+  return (
+    <RouterContext.Provider value={{ params, location }}>
+      {children}
+    </RouterContext.Provider>
+  )
+}
```
Crash--, line 15: Can you explain within the commit message, why? 

withRouter seems to exist: https://v5.reactrouter.com/core/api/withRouter ? 

Also, I don't see react-router to be part as a dev-dep & peerDep https://github.com/cozy/cozy-libs/blob/master/packages/cozy-sharing/package.json 

So I think we have this regression here because one package has removed it's dep to react-router 

Also, some app consuming cozy-sharings have old react-router version: https://github.com/cozy/cozy-drive/blob/master/package.json#L164

Is it still working? After a quick search, I tend to think that no... useParams is not present here https://github.com/remix-run/react-router/blob/v3/docs/API.md

packages/cozy-harvest-lib/src/components/EditAccountModal.jsx
```diff
@@ -23,7 +23,7 @@ import { withMountPointHistory } from './MountPointContext'
 import logger from '../logger'
 import { withTracker } from './hoc/tracking'
 import useTimeout from './hooks/useTimeout'
-import { withRouter } from 'react-router'
+import { withRouter } from 'react-router-dom'
```
Crash--, line 26: This one, why not. https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/package.json#L87 

But I think it'll crash on cozy-drive because it imports harvest without respecting the peerDep...

packages/cozy-harvest-lib/test/AppLike.jsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react'
+import { BrowserRouter as Router } from 'react-router-dom'
```
Crash--, line 2: I'm a bit afraid about importing from react-router-dom

packages/cozy-flags/package.json
```diff
@@ -31,6 +31,9 @@
     "react": "16.12.0"
   },
   "jest": {
+    "globals": {
+      "window": {}
+    },
```
Crash--, line 36: We don't want to have this `window` stuff. cozy-flags is also used in a node env. It should work without the need to rely on window.

packages/cozy-sharing/src/SharingBanner/test/RouterContext.jsx
Merkur39: Je pense que ce fix n'est pas bon en effet.

>So I think we have this regression here because one package has removed it's dep to react-router

Il faudrait clairement ajouter `react-router` aux devDeps/peerDeps et annuler ce changement 👍

packages/cozy-harvest-lib/test/AppLike.jsx
Merkur39: Je vais ajouter `react-router` aux devDeps/peerDeps et annuler ce changement qui ne sera plus nécessaire

packages/cozy-flags/package.json
Merkur39: En effet, la version de `cozy-client` étant assez en retard sur `cozy-flags`, ce derniers ne bénéficiait pas des fix autour de ce sujet.
Une simple maj de `cozy-client` suffit à corriger le problème, et donc à ne plus avoir besoin de ce changement!

packages/cozy-harvest-lib/src/components/InAppBrowser.jsx
```diff
@@ -56,7 +56,15 @@ const InAppBrowser = ({ url, onClose, intentsApi = {} }) => {
...
-  }, [isReady, url, onClose])
+  }, [
+    isReady,
+    url,
+    onClose,
+    closeInAppBrowser,
+    fetchSessionCode,
+    showInAppBrowser,
+    tokenParamName
+  ])
```
trollepierre, line 67: are you sure of this change? have you verified that don't break nothing?

to make a safer change, I would have disabled eslint here.

packages/cozy-intent/src/view/components/WebviewIntentProvider.tsx
```diff
@@ -103,7 +103,7 @@ export const WebviewIntentProvider = ({
       !webviewService &&
       isValidEnv() &&
       getConnection(setConnection, methods)
-  }, [connection, webviewService])
+  }, [connection, webviewService, methods])
```
trollepierre, line 106: is this change safe?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/ContractItemSecondaryText.jsx
```diff
@@ -64,7 +63,7 @@ const ContractItemSecondaryText = ({ contract }) => {
 }
 
 ContractItemSecondaryText.propTypes = {
-  contracts: PropTypes.oneOfType([CollectionPropType, PropTypes.array])
+  contract: PropTypes.object
```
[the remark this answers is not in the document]
doubleface: The thing is that this component expext a unique contract and not a collection of contracts or an array

Crash-- (approved):
🎉

PR #1721, opened 2022-07-04
fix(mespapiers-lib): Invalidate partial but non-empty values in optional masked inputs

PolariTOON:
#1695 allowed masked inputs to be optional: the value could be either empty, partial or full, as long as it matched the start of the mask.
This PR restricts this set by disallowing partial values: now only the empty string (for optional inputs) and values that match and fill the entire mask are valid.


Merkur39 (approved):
LGTM 👌

PR #1723, opened 2022-07-05
feat: Update @babel/runtime to 7.18.6

Merkur39:
After this change #1631, commands like
`yarn start` or `yarn build` on some libs
were causing this error
```
Error: [BABEL]: Cannot find module '@babel/plugin-syntax-jsx'
```
I haven't found out exactly which version fixes this,
but the problem is no longer on version `7.18.6`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "devDependencies": {
     "@babel/cli": "7.16.8",
     "@babel/core": "7.16.12",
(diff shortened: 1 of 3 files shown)


JF-Cozy (approved):
on peut peut-être patcher packages/playgrounds/package.json également du coup ?

PR #1732, opened 2022-07-11
feat: Allow sharemodal to use Location

acezard:
This will allow ShareModal to get itself the props it needs
from the react-router state if they exist


packages/cozy-sharing/src/components/ShareModal.jsx
```diff
@@ -93,7 +93,7 @@ ShareModal.propTypes = {
   hasSharedChild: PropTypes.bool,
   hasSharedParent: PropTypes.bool,
   isOwner: PropTypes.bool,
-  link: PropTypes.string.isRequired,
+  link: PropTypes.string,
```
acezard, line 96: Doesn't seem to be always required, works fine in cozy-notes without it. Removed the required for less console spam

acezard: well, it's null by default as I can see. Not too sure why it works on Notes without the prop, I could investigate

acezard: I'm going to backtrack on that change, I'm not too sure what's supposed to happen with these props. But in any case, Notes never provided it in my use case.

trollepierre (approved):

PR #1736, opened 2022-07-13
fix: Revert ShareByEmail for Notes

acezard:
Reverting to the code used before email implementation
in Notes share modals


Crash--:
Did you revert the commit? 🤔

Ldoppea:
Just a comment to specify that this revert should be temporary until we fix the `error.generic` bug when sharing a note that has no title.

Ldoppea (approved):

Crash--:
> Just a comment to specify that this revert should be temporary until we fix the error.generic bug when sharing a note that has no title (we need the why to ease future us to understand what happened).

This is not for that reason that we revert the sharing by link. This is because the link sent by mail is not working.

PR #1737, opened 2022-07-13
feat: Add a cache to BI temporary token

doubleface:
Since this token can last 30 min and can be used many times, and since
    there are multiple calls to createTemporaryToken with BI webviews now,
    the return value of createTemporaryToken is now cached (29min for
    safety)
    
    This make BI webviews interfaces a lot faster
    
    Notes :
    
    * This cache is activated only for bi webview (with harvest.bi.webview
    flag set to true
    * This cache needs last version of banking connectors which return a
    `biMapping` attribute, which allow harvest to translate cozy banks ids
    to bi bank ids itself. This value will be cached

### DIFF
--- a/packages/cozy-harvest-lib/src/services/biWebView.js
+++ b/packages/cozy-harvest-lib/src/services/biWebView.js
@@ -7,6 +7,7 @@
 import { getBIConnection, getBIConnectionAccountsList } from './bi-http'
 import assert from '../assert'
 import logger from '../logger'
+import { Q } from 'cozy-client'
 import flag from 'cozy-flags'
 import {
   setBIConnectionId,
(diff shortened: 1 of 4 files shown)


trollepierre (approved):

packages/cozy-harvest-lib/src/services/biWebView.js
```diff
@@ -336,28 +380,29 @@ export const createTemporaryToken = async ({ client, konnector, account }) => {
     konnector.slug,
     'createTemporaryToken: konnector passed in options has no slug'
   )
+
+  let cache = await getTokenFromCache({ client })
```
Crash--, line 384: cachedToken or something?

packages/cozy-harvest-lib/src/services/biWebView.js
```diff
@@ -322,6 +323,49 @@ function convertBIDateToStandardDate(biDate) {
   return biDate?.replace(' ', 'T')
 }
 
+async function getTokenFromCache({ client }) {
+  const queryResult = await client.query(
+    Q('io.cozy.bank.settings').getById('bi')
```
Crash--, line 328: Since this is specific to the process of creating a temporary token, then it should be named that way. `bi` is too generic. `temporaryBiToken` `biTmpToken` or else, but something specific please

packages/cozy-harvest-lib/src/services/biWebView.js
```diff
@@ -322,6 +323,49 @@ function convertBIDateToStandardDate(biDate) {
...
+  )
+  return queryResult?.data || {}
+}
+
+function isCacheExpired({ cache }) {
```
Crash--, line 333: js doc

packages/cozy-harvest-lib/src/services/biWebView.js
```diff
@@ -322,6 +323,49 @@ function convertBIDateToStandardDate(biDate) {
...
+  const cacheAge = Date.now() - Number(cache?.timestamp)
+  logger.debug('cache age', cacheAge / 1000 / 60, 'minutes')
+  const MAX_CACHE_AGE = 29 * 60 * 1000
+  if (cache && cacheAge < MAX_CACHE_AGE) {
+    return false
+  }
+  return true
+}
+
+async function updateCache({ client, konnector, cache }) {
```
Crash--, line 343: js doc

PR #1740, opened 2022-07-15
feat: Fetch connector from cozy-stack if undefined

Ldoppea:
If no connector is given by calling app, the Route.jsx wouldn't be able
to display anything and would throw when trying to access connector
props

To allow `cozy-harvest-lib` to display the connector anyway, we can
fetch connector data from cozy-stack based on the new `konnectorSlug`
prop

This scenario may happen in flagship app if RealTime is not active on
`cozy-home` and the app asks `cozy-home` to display a newly installed
connector

However this does not allow to display a not installed connector. If so
`cozy-harvest-lib` would throw as before

___

Implementation is based on https://github.com/cozy/cozy-harvest-app/blob/f4ee79865a9e74bf67f12d1ca9d85f9905e65f4a/src/hooks/useKonnectorWithTriggers.js

### DIFF
--- a/packages/cozy-harvest-lib/src/components/Routes.jsx
+++ b/packages/cozy-harvest-lib/src/components/Routes.jsx
@@ -7,6 +7,7 @@ import {
   DialogCloseButton,
   useCozyDialog
 } from 'cozy-ui/transpiled/react/CozyDialogs'
+import Spinner from 'cozy-ui/transpiled/react/Spinner'
 import {
   useVaultUnlockContext,
   VaultUnlockProvider,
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/components/Routes.jsx
```diff
@@ -48,107 +49,127 @@ const HarvestDialog = withStyles({
...
-                            <KonnectorSuccess
-                              konnector={konnector}
+
+  const { konnectorWithTriggers, fetching } = useKonnectorWithTriggers(
+    konnectorSlug,
+    konnector
+  )
+
+  if (fetching) {
+    return <p></p>
```
acezard, line 71: This doesn't look so great to me, can't we show a spinner

Crash-- (approved):

Ldoppea:
Here is a proposal for displaying a spinner while the connector is fetched from the stack: https://github.com/cozy/cozy-libs/pull/1740/commits/013b3a25542837fc940eff8229f8c70e46ebfc7b

Here is the result when the home tries to display a connector that is not defined in props:

https://user-images.githubusercontent.com/1884255/181073411-ff4528b7-f925-4a34-96e5-0bb65f52c3d7.mov

For reference here is the behavior when displaying a connector that is defined in props but with no cached data in store:

https://user-images.githubusercontent.com/1884255/181073626-fe1d9aea-ac3a-4c2d-a09c-183d9ea669d1.mov

And here is the behavior when displaying a connector that is defined in props and is cached data in store:

https://user-images.githubusercontent.com/1884255/181073759-3d72d5d9-0a0b-4336-9fba-3198e95c8021.mov

Crash-- (approved):

PR #1742, opened 2022-07-18
feat: Upgrade to react-router6

acezard:


Crash-- (approved):

packages/cozy-sharing/src/ShareModal.jsx
```diff
@@ -30,20 +29,19 @@ const SharingModal = ({ document, ...rest }) => (
 )
 
 export const ShareModal = withLocales(props => {
```
acezard, line 31: Investigate on the feasibility of fetching data from current route param

PR #1743, opened 2022-07-18
fix: Disallow email sharing for io.cozy.notes

acezard:
The old documentType did not work anymore.
Using private property to identify note file.

Also simplified render logic


acezard:
> It should work. We still pass it there: https://github.com/cozy/cozy-notes/blob/0a3f10581f2e7a9e3dc8249ed91c727a52bdc313/src/targets/browser/index.jsx#L78
> 
> So let's understand why it doesn't work anymore?

It returns "File" as documentType now. Isn't that linked to the fact Note is an archive now? I'm not sure cc @nono

acezard:
> I still don't buy the answer. documentType is declared by the front app (cf the link I posted earlier), so it must go down to this component. Nothing related to the stack here.
> 
> So if the documentType is not the good one, something in cozy-sharings must change it. But why? This is the question.

Well I don't know I will do further research then

PR #1744, opened 2022-07-18
fix: Do not use biBankId from cache

doubleface:
When creating multiple connections on multiple banks, we could have a
webview for the first bank because we did not use the
createTemporaryToken properly.

Since with BI webviews, biBankId has no use. We now only use biBankIds
value, which is also updated even when using cached token.

### DIFF
--- a/packages/cozy-harvest-lib/src/services/biWebView.js
+++ b/packages/cozy-harvest-lib/src/services/biWebView.js
@@ -201,20 +201,18 @@ export const onBIAccountCreation = async ({
 /**
  * Create OAuth extra parameters specific to reconnect webview
  *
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/services/biWebView.js
```diff
@@ -208,13 +208,12 @@ export const onBIAccountCreation = async ({
  * @return {Object}
  */
 const getReconnectExtraOAuthUrlParams = async ({
```
Crash--, line 210: js doc to update...

Crash-- (approved):
update the jsdoc please.

PR #1746, opened 2022-07-19
feat: Now allow disabled bank accounts to be clickable

doubleface:
This allows users to remove bank accounts even when they are disabled.

image


packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/ContractItem.jsx
```diff
@@ -17,6 +17,10 @@ import ContractItemSecondaryText from './ContractItemSecondaryText'
 const ContractItem = ({ contract, konnector, accountId, divider }) => {
   const [showingEditModal, setShowingEditModal] = useState(false)
 
+  const disabledColorWhenNeeded = isDisabled(contract)
+    ? 'var(--disabledTextColor)'
+    : null
```
Crash--, line 22:
```suggestion
    : var(--iconTextColor)
```

No?

Crash-- (approved):

PR #1748, opened 2022-07-19
fix: Do not fail on temporary token cache save error

doubleface:
In some concurrent cases, we could get 409 error while saving
temporary token cache.
The error is now logged as warning and ignored since at least one of
the two ore more saving request was a success

Any of the two token will be saved and useable.

### DIFF
--- a/packages/cozy-harvest-lib/src/services/biWebView.js
+++ b/packages/cozy-harvest-lib/src/services/biWebView.js
@@ -369,14 +369,20 @@ async function updateCache({ client, konnector, tokenCache }) {
     'result',
     TEMP_TOKEN_TIMOUT_S * 1000
   )
-  const saveResult = await client.save({
-    _type: 'io.cozy.bank.settings',
-    _id: 'bi',
-    ...(tokenCache?._rev ? { _rev: tokenCache._rev } : {}),
-    timestamp: Date.now(),
-    ...event.data.result
-  })
-  return saveResult.data
+  try {
+    const saveOptions = {
+      _type: 'io.cozy.bank.settings',
+      _id: 'bi',
+      ...(tokenCache?._rev ? { _rev: tokenCache._rev } : {}),
+      timestamp: Date.now(),
+      ...event.data.result
+    }
+    const saveResult = await client.save(saveOptions)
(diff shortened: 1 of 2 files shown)


Crash--:
> Any of the two token will be saved and useable. 

this is not true. Only the first one will be saved and useable. And we have generated a new temporary token with an expire date ++ that we'll not use... 

I kinda dislike this solution.

doubleface:
The only way I found to thrigger the conflict was to trigger two calls to createTemporaryToken while the cache is not set (or too old). Here is how I do it :

- I remove the cached token from db
- Go to the configuration screen (the one with the "Gérer mes connexions" button) of an existing connection
- Click on "Ajouter un compte" before the "Gérer mes synchronisations" button is enabled

This way we have 2 concurrent createTemporaryToken jobs and create the conflict

doubleface:
@Crash-- Actually there are some cases where the request : 

```
  await client.query(
    Q('io.cozy.bank.settings').getById('bi')
  )
```

Always returns `null` which is then the cause of many conflicts. Maybe a problem of cozy-client policy. I'm investigating

Crash--:
`null`is weird. Also, there is no fetchPolicy here is this request so it's not linked to that. 

Even if it's linked to fetchPolicy you'll have an undefined instead of null https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L916 

But having `null` I don't understand yet. I imagine we pass here https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/CozyClient.js#L919-L926. When I look the tests I wrote for this feature https://github.com/cozy/cozy-client/commit/f3f1f0f0136555a43c4fd157d96be68b11a818f8 I see that I don't have any test that check the behavior for a "loading query". But, if I go deeper, when we init a query the data is set to `[]` and not `null`. https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/store/queries.js#L58 

Does the request is dispatched to the stack ? 

So it means that we

doubleface:
On n'a plus ce problème désormais depuis la mise à jour de cozy-client en v33. Donc je ferme la PR

PR #1753, opened 2022-07-21
fix: Add ts-check on biWebview

Crash--:
The idea is to be careful about our type
and be confident about the change we make.

### DIFF
--- a/packages/cozy-harvest-lib/src/services/biWebView.js
+++ b/packages/cozy-harvest-lib/src/services/biWebView.js
@@ -1,3 +1,4 @@
+// @ts-check
 /**
  * Interface between Budget Insight and Cozy using BI's webview
  *
@@ -8,7 +9,9 @@ import { getBIConnection, getBIConnectionAccountsList } from './bi-http'
 import assert from '../assert'
 import logger from '../logger'
 import { Q } from 'cozy-client'
+// @ts-ignore (its a peerDep and I don't know how to configure ts for that)
 import flag from 'cozy-flags'
+
 import {
   setBIConnectionId,
   saveBIConfig,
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/helpers/konnectors.js
```diff
@@ -76,6 +76,11 @@ export class KonnectorJobError extends Error {
      */
     const type = this.code.split('.')[0]
     this.type = KNOWN_ERRORS.includes(type) ? type : UNKNOWN_ERROR
+
+    // I don't know where this is used, but in biWebview.js or budget-insight.js
+    // we assign the account id to the error. Should we remove this definition?
+    // or should we remove the assignation?
```
doubleface, line 82: This accountId is actually not used. I just added this as a debugging information. We can remove it here and in budget-insight.js and biWebview.js

packages/cozy-harvest-lib/src/services/biWebView.js
```diff
@@ -22,20 +25,26 @@ import { waitForRealtimeEvent } from './jobUtils'
 import '../types'
 import { LOGIN_SUCCESS_EVENT } from '../models/flowEvents'
 
+/**
+ * @typedef {import("../models/ConnectionFlow").default} ConnectionFlow
```
doubleface, line 29: Nice, I did not now this way to import type definitions

packages/cozy-harvest-lib/src/services/biWebView.js
```diff
@@ -8,7 +9,9 @@ import { getBIConnection, getBIConnectionAccountsList } from './bi-http'
 import assert from '../assert'
 import logger from '../logger'
 import { Q } from 'cozy-client'
+// @ts-ignore (its a peerDep and I don't know how to configure ts for that)
```
acezard, line 12: you can add ambient declarations with d.ts files https://www.typescriptlang.org/docs/handbook/modules.html

acezard (approved):

PR #1755, opened 2022-07-22
fix: Close the multi-selector returns to the homepage

Merkur39:
When we select a paper from the Viewer, then we close the multi-selection, we return to the Viewer, but the Viewer does a `goBack` if a history exists (useful to see a selected paper).
But in this present case, we fall into a loop of navigation.
We choose to return to the homepage when closing the multi-selection.


Merkur39:
[Viewer retour arrière en boucle](https://trello.com/c/0EwUMPQu/921-viewer-retour-arri%C3%A8re-en-boucle)

JF-Cozy (approved):

PR #1761, opened 2022-07-25
Fix Oauth Flow in non intentsApi apps

Crash--:
We had a few regressions in our flagship app for our OAuth Konnectors because of the work done for the BI webview: The InAppBrowser was opened but we tried to open it several times resulting an error (because the IAB can be only open once) since fetchSessionCode() was returning an error. 

This PR fix these issues.


Crash--:
@trollepierre @doubleface can you review this PR? 🙏

trollepierre (approved):

PR #1762, opened 2022-07-26
chore: Mui/lab has to be in peerDep

Crash--:
We rely on /lab for some Harvest components. So let's add it to the peerDep section.

### DIFF
--- a/packages/cozy-harvest-lib/package.json
+++ b/packages/cozy-harvest-lib/package.json
@@ -77,6 +77,7 @@
   "peerDependencies": {
     "@babel/runtime": ">=7.12.5",
     "@material-ui/core": ">=4",
+    "@material-ui/lab": "4.0.0-alpha.60",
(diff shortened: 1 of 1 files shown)


JF-Cozy (approved):
la dernière est `4.0.0-alpha.61`. Il faut spécifiquement la 60 ici ?

Crash--:
I don't know this is what I have here https://github.com/cozy/cozy-home/blob/master/package.json#L38 

I don't want to broke something 🤸‍♂️

trollepierre:
> la dernière est `4.0.0-alpha.61`. Il faut spécifiquement la 60 ici ?

I agree with @JF-Cozy , we should have used:
"@material-ui/lab": ">4.0.0-alpha.60",

Crash--:
I'm not sure. This is an "alpha" version. There is nothing telling us that in alpha 61, 62 or else this component will be still there.

PR #1794, opened 2022-08-16
Feat/add actions search result

Merkur39:


PolariTOON (approved):
At lot of code that could likely be refactored with `PaperLine` and `PapersList`, but that can come later.

Merkur39:
[Le menu des actions par fichier disparait lors de la recherche/filtres](https://trello.com/c/UrLhqkcM/940-le-menu-des-actions-par-fichier-disparait-lors-de-la-recherche-filtres)

PR #1804, opened 2022-08-23
feat: Reconnection with BI webviews & webhooks

taratatach:
When a banking connector fails with a LOGIN_ERROR status, we offer
users to reconnect their connector via the BI webview.
Once the user closes the webview, we trigger a connector launch to try
the new credentials (we can still encounter an error) and fetch the
new data.
The user sees a synchronization is in progress within the connector's
modal in Harvest.

However, we want to use BI webhooks to receive data pushed from BI
rather than pulling data when we trigger the connector and the current
process cannot handle this because:
- we won't trigger the connector ourselves (it will be triggered by a
  webhook call) so the user wouldn't see a synchronization is in
  progress
- the user might not close the webview or the webhook call could be
  received before the webview is closed so we need to expect it ahead
  of time

Since we don't have much control over the webview, we'll update the
in-memory connection flow state to expect the connector trigger launch
(via the BI webhook call). The update will be done once the user opens
the webview by clicking the "Reconnect" button displayed in the
connector error card.
From the user perspective, this trigger launch expectation will be
seen as a synchronization in progress.
Once the connector trigger is launched, we'll remove this in-memory
state and start watching for changes on the connector job.


packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/BiContractActivationWindow.jsx
```diff
@@ -49,7 +49,7 @@ const BIContractActivationWindow = ({
     if (konnectorPolicy.fetchExtraOAuthUrlParams) {
       handleLinkFetch()
     }
-  }, [konnector.slug, account, client, konnectorPolicy])
+  }, [konnector, account, client, konnectorPolicy])
```
doubleface, line 52: This konnector.slug was here on purpose. Or else you get multiple renders and multiple calls  to createTemporaryToken.

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.spec.jsx
```diff
@@ -30,3 +48,66 @@ describe('Konnector account tabs', () => {
     expect(root.getByText('Configuration')).toBeTruthy()
   })
 })
+
+describe.only('Konnector account tabs content', () => {
```
doubleface: .only to remove

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/BiContractActivationWindow.jsx
taratatach: Ah. The temporary token is stored in the konnector?

doubleface (approved):

packages/cozy-harvest-lib/src/components/OAuthForm.jsx
```diff
@@ -16,142 +16,127 @@ import { KonnectorJobError } from '../helpers/konnectors'
...
+          onCancel={handleOAuthCancel}
+          account={account}
+          intentsApi={intentsApi}
+        />
+      )}
+    </>
+  )
 }
 
 OAuthForm.propTypes = {
```
Crash--, line 138: @taratatach nit FYI this should not be tagged as `chore`. `Chore` is only about stuff that are not deployed like CI stuff or things like that. Here you are editing what we consume so it's more a `refactor` commit https://github.com/cozy/cozy-guidelines/#type

packages/cozy-harvest-lib/src/components/__snapshots__/OAuthForm.spec.js.snap
```diff
@@ -14,9 +14,9 @@ exports[`OAuthForm should handle oauth cancelation 1`] = `
     }
   />
   <DefaultButton
-    busy={true}
+    busy={false}
```
Crash--: @doubleface @taratatach is it really wanted to change this value? Is it because before we didn't wait the update of the component?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx
```diff
@@ -91,6 +100,23 @@ const DumbKonnectorAccountTabs = props => {
...
+              <RedirectToAccountFormButton
+                konnector={konnector}
+                trigger={initialTrigger}
+              />
+            ) : null
+          }
+          className="u-mt-1"
+        />
+      )}
+
```
Crash--, line 164: By doing this, you are changing the way we display the `TriggerErrorInfo` component, right? Before it was inside the `SwipeableViews` so I can imagine inside a container, now it's outside. 

Can you share a before / after screenshot to see the difference ?

packages/cozy-harvest-lib/src/components/__snapshots__/OAuthForm.spec.js.snap
```diff
@@ -1,38 +1,54 @@
...
+      </div>
+    </div>
+  </div>
+  <button
+    class="styles__c-btn___-2Vnj styles__c-btn--full___2VmR1 styles__c-btn--center___16_Xh u-mt-1"
+    type="submit"
+  >
+    <span>
+      <span>
+        oauth.connect.label
```
Crash--, line 34: It seems that here https://github.com/cozy/cozy-libs/pull/1804/commits/5b7db192cb641a47aa806e04080543722f557674#diff-3eb34f56944764b1f3833c7ab9d70f5c41244b75c5f024d4d0019e2461c35e3eR22 the string is translated, but not for `oauth.connect.label`. Is it a bug?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx
```diff
@@ -8,13 +8,17 @@ import Divider from 'cozy-ui/transpiled/react/MuiCozyTheme/Divider'
 import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 import useBreakpoints from 'cozy-ui/transpiled/react/hooks/useBreakpoints'
 import SwipeableViews from 'react-swipeable-views'
+import Button from 'cozy-ui/transpiled/react/Button'
```
Crash--, line 12: https://docs.cozy.io/cozy-ui/react/#!/Button

> ⚠️ This <Button /> is going to be deprecated, please use [Buttons](https://docs.cozy.io/cozy-ui/react/#/Buttons) instead.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -184,6 +185,7 @@ export class ConnectionFlow {
   }
 
   getKonnectorSlug() {
+    // TODO: Use this.konnector.slug?
```
Crash--, line 189: is it still needed?

packages/cozy-harvest-lib/src/components/OAuthForm.jsx
```diff
@@ -59,9 +60,13 @@ export const OAuthForm = props => {
...
 
-  const handleConnect = () => {
+  const handleConnect = useCallback(() => {
     setShowOAuthWindow(true)
-  }
+    const konnectorPolicy = findKonnectorPolicy(konnector)
+    if (konnectorPolicy.isBIWebView) {
+      flow.expectTriggerLaunch({ konnector })
+    }
+  }, [client, flow, konnector.slug])
```
Crash--, line 70: Why `konnector.slug`? You are using the whole object `konnector`. So this should be this object within the dep array. 

Your IDE should have tell you that.

packages/cozy-harvest-lib/src/services/jobUtils.js
```diff
@@ -30,3 +30,8 @@ export const waitForRealtimeEvent = (
       reject(new Error('Timeout for waitForRealtimeResult'))
     }, timeout)
   })
+
+export const sendRealtimeNotification = ({ client, data }) => {
+  const rt = new CozyRealtime({ client })
+  rt.sendNotification(JOBS_DOCTYPE, 'test', data)
+}
```
Crash--, line 37: Now that you know you can use client.plugins.realtime can you use it for the new code please?

packages/cozy-harvest-lib/src/components/hooks/useOAuthExtraParams.js
```diff
@@ -27,7 +27,14 @@ const useOAuthExtraParams = ({ account, client, konnector, reconnect }) => {
     if (needsExtraParams) {
       load()
     }
-  })
+  }, [
+    account,
+    client,
+    konnector.slug,
```
Crash--, line 33: Same there, it should be konnector, not konnector.slug

packages/cozy-harvest-lib/src/services/jobUtils.js
doubleface: @Crash-- Done but the cozy-client realtime plugin does not expose the sendNotification method. Fixed with https://github.com/cozy/cozy-libs/pull/1846

Will need to upgrade cozy-realtime in home and banks after that and in peerDep of harvest after next cozy-realtime release.

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx
```diff
@@ -144,4 +216,4 @@ KonnectorAccountTabs.propTypes = {
   innerAccountModalOverrides: innerAccountModalOverridesProptype
 }
 
-export default KonnectorAccountTabs
+export default withClient(KonnectorAccountTabs)
```
Crash--, line 219: Why are we adding an HOC here? Can't we use useClient() inside the function?

packages/cozy-harvest-lib/src/services/biWebView.js
```diff
@@ -77,7 +77,7 @@ export const checkBIConnection = async ({ connId, client, konnector }) => {
  * @param {CozyClient} options.client CozyClient object
  * @param {Boolean} options.reconnect If this is a reconnection
  * @param {Function} options.t Translation fonction
- * @return {Promise<Number|null>} Connection Id
+ * @return {Promise<Boolean>} Connection Id
```
Crash--, line 80: This is not `Connection Id` then. Can you update the comment

packages/cozy-harvest-lib/src/services/budget-insight.js
```diff
@@ -220,7 +220,7 @@ export const setBIConnectionId = (originalAccount, biConnectionId) => {
  * @param {Object} options.konnector connector manifest content
  * @param {Object} options.client CozyClient object
  *
- * @return {Integer} Connection Id
+ * @return {Promise<Boolean>} Connection Id
```
Crash--, line 223: same here

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx
doubleface: Now I use useClient

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: No, sometimes connectionFlow can be initialized without any konnector

packages/cozy-harvest-lib/src/components/__snapshots__/OAuthForm.spec.js.snap
doubleface: No this is not a bug. The t function is mocked to return the translation key itself. We still have translation in TriggerErrorInfo component which uses I18n from cozy-ui directly.

packages/cozy-harvest-lib/src/components/__snapshots__/OAuthForm.spec.js.snap
doubleface: The test now waits for the fetch of extra params to be done, which explains why the button is not busy anymore. And also, the test has been rewritten since the move of the OAuthForm component to a functionnal component.

I also checked manually and the buton still displays the busy state when loading extra parameters.

Crash--: Is it a good idea to have : 
- first test that check if the button is busy while the hook is / will be called 
- Then resolve the hook then check that the button is not busy anymore 

?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/KonnectorAccountTabs.jsx
doubleface: before

after

I just see that there is a difference with the button color, related to the use of the buttons component I suppose

Crash--: So we need to fix this color issue.

doubleface: I cannot reproduce this issue which came after the use of Butttons instead of Button. It looks like it was a problem related to watch and rlink I think.

packages/cozy-harvest-lib/src/components/__snapshots__/OAuthForm.spec.js.snap
doubleface: Ok I added a test for busy and not busy state

doubleface:
@Crash-- J'ai mis à jour cette PR avec les dernières remarques

Crash-- (approved):

PR #1809, opened 2022-08-25
fix: Misleading example konnectors request in readme file

doubleface:
Konnectors doctype is io.cozy.konnectors and not io.cozy.apps


Crash-- (approved):

Crash--:
let's rename this commit as `docs:`. there is no need to deploy a new version of the lib.

PR #1812, opened 2022-08-29
fix: Add prepublish script

Crash--:
We add issue on a few packages since we have
added a new travis' job. Indeed, the
"repo doctor"'s job is the first one to be run.

So it's also the one to handle the lerna
process to publish. But since this job doesn't
build the package, we can have published package
without all the build stuff in it.

So let's add a prepublish script to all the
repo in order to not have issue and dep with
the job's order.


doubleface (approved):
Ok for me but you have to remove the .DS_Store file

PR #1815, opened 2022-08-30
feat: Update eslint-config-cozy-app

acezard:
We were a bit behind in versions, bugs started to leak in our lints.
These version upgrades should fix them,
but they also have a small chance to detect new lint errors.

### DIFF
--- a/packages/eslint-config-cozy-app/package.json
+++ b/packages/eslint-config-cozy-app/package.json
@@ -19,17 +19,16 @@
     "vue.js"
   ],
   "dependencies": {
-    "@babel/eslint-parser": "^7.17.0",
-    "@typescript-eslint/eslint-plugin": "^5.19.0",
-    "@typescript-eslint/parser": "^5.19.0",
-    "eslint": "^8.13.0",
-    "eslint-config-prettier": "^8.5.0",
-    "eslint-plugin-prettier": "^4.0.0",
-    "eslint-plugin-promise": "^6.0.0",
-    "eslint-plugin-react": "^7.29.4",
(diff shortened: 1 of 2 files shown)


packages/eslint-config-cozy-app/package.json
```diff
@@ -19,17 +19,16 @@
...
-    "@babel/eslint-parser": "^7.17.0",
-    "@typescript-eslint/eslint-plugin": "^5.19.0",
-    "@typescript-eslint/parser": "^5.19.0",
-    "eslint": "^8.13.0",
-    "eslint-config-prettier": "^8.5.0",
-    "eslint-plugin-prettier": "^4.0.0",
-    "eslint-plugin-promise": "^6.0.0",
-    "eslint-plugin-react": "^7.29.4",
-    "eslint-plugin-react-hooks": "^4.5.0",
-    "eslint-plugin-vue": "^8.6.0",
```
trollepierre, line 31: isn't it a breaking change to remove plugin vue ?

acezard: @trollepierre I guess it is, but do we even have a Vue app? It seems legacy

acezard: @trollepierre added breaking change in commit description

trollepierre (approved):

packages/eslint-config-cozy-app/package.json
trollepierre: I wonder if we should ask to front team about their opinion. 
With a breaking change, there is no way we are faulty

PR #1821, opened 2022-09-06
fix: Open oauth window for oauth connectors

doubleface:
This was a regression from BI webview evolution to open oauth window
from reconnect button


doubleface:
@Crash-- Since Erwan reworked this component + unit tests in https://github.com/cozy/cozy-libs/pull/1804 , I would prefer to write those unit tests after the merge of this PR.

PR #1825, opened 2022-09-12
feat(cozy-intent): Update cozy-settings switch API

acezard:
One method instead of n methods for n possible setting

### DIFF
--- a/packages/cozy-intent/src/api/models/methods.ts
+++ b/packages/cozy-intent/src/api/models/methods.ts
@@ -1,6 +1,8 @@
 import { AppManifest, FlagshipUI } from '../../api'
 
-export type NativeMethodsRegister = {
+type PostMeDefault = Record<string, (...args: unknown[]) => Promise<null>>
+
+interface _NativeMethodsRegister {
   backToHome: () => Promise<null>
   hideSplashScreen: () => Promise<null>
   logout: () => Promise<null>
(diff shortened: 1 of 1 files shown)


packages/cozy-intent/src/api/models/methods.ts
```diff
@@ -12,6 +12,9 @@ export type NativeMethodsRegister = {
   openSettingBiometry: () => Promise<boolean>
   setFlagshipUI: (flagshipUI: FlagshipUI, caller?: string) => Promise<null>
   showSplashScreen: () => Promise<null>
+  toggleSetting: (
+    settingName: 'biometryLock' | 'PINLock' | 'autoLock'
+  ) => Promise<boolean | null>
```
Ldoppea, line 17: Why do we need `boolean | null` as return type? Shouldn't it be always `boolean`?

packages/cozy-intent/src/api/models/methods.ts
```diff
@@ -1,6 +1,6 @@
 import { AppManifest, FlagshipUI } from '../../api'
 
-export type NativeMethodsRegister = {
+export interface NativeMethodsRegister {
```
Ldoppea, line 3: Did you encounter some limitations with the previous `type` notation?

Ldoppea (approved):

packages/cozy-intent/src/api/models/methods.ts
acezard: It should return a boolean to confirm everything went well (even if we don't need the value since we have `flagshipMetadata` as a source of truth, and not an awaited value). But in case of error, null is the default return in async post-me.
You could make an argument that the boolean is useless and we should always use null return, but maybe we want to add an error handler web side, sometime down the road

packages/cozy-intent/src/api/models/methods.ts
acezard: Actually I didn't even notice that change. It seems it was made automatically with the new config I added to cozy eslint config for TS files: `'plugin:@typescript-eslint/strict'`

PR #1837, opened 2022-09-24
fix: Bind this automatically with arrow function

Crash--:
In drive, when moving a file, I got a "this.props" is undefined. It seems that fetchAllsharing was
already binded to this, but by converting it to
an arrow function, I don't have the issue anymore.


acezard (commented):
@Crash--  is this still up to date please? (I'm reviewing some older PRs I'm assigned to)

acezard (approved):

tylerteresa706-hub (approved):

PR #1838, opened 2022-09-26
feat(mespapiers): We can now edit a metadata even  if it didn't exist

JF-Cozy:


packages/cozy-mespapiers-lib/src/components/ModelSteps/Edit/helpers.spec.js
```diff
@@ -129,15 +129,15 @@ describe('isInformationEditPermitted', () => {
 
     expect(res).toBe(false)
   })
-  it('should return False if tries to edit a metadata that does not exist on current file', () => {
+  it('should return True if tries to edit a metadata that does not exist on current file', () => {
```
Merkur39, line 132:
```suggestion
  it('should return True if tries to edit a metadata that is not filled in the current file', () => {
```

Merkur39 (approved):

PR #1842, opened 2022-09-30
feat: Update device-helper interface

acezard:
Used in flagshipapp and consumers


packages/cozy-device-helper/src/flagship.ts
```diff
@@ -8,46 +8,34 @@ export enum FlagshipRoutes {
...
   platform?: Record<string, unknown>
   route?: FlagshipRoutes
-  settings?: {
-    PINEnabled?: boolean
-    autoLockDelay?: number
-    autoLockEnabled?: boolean
-    biometryEnabled?: boolean
-  }
+  settings_PINEnabled?: boolean
+  settings_autoLockDelay?: number
```
Ldoppea, line 21: nit: I should have commented that in previous PRs, but it would be nice to have the unit in the variable name (i.e `settings_autoLockDelayMs`)

Ldoppea (approved):

packages/cozy-device-helper/src/flagship.ts
acezard: There will most likely be another release soon

PR #1848, opened 2022-10-04
feat: Allow to pass custom Properties to OAuthForm wrapper

doubleface:
via OAuthFormWrapperCompProps and without unmounting/remounting
OauthForm on each rerender

This avoids the disparition of the BI webview in banks when a new job is
expected


Crash-- (changes requested):
Can you update the propTypes please? https://github.com/cozy/cozy-libs/blob/500b2f72275315c10cf1dd18cb0a1028562ce8f9/packages/cozy-harvest-lib/src/components/TriggerManager.jsx#L411

Crash-- (approved):

PR #1852, opened 2022-10-05
feat: Use BI manage url to update contract synchronization

doubleface:
This PR needs some update in cozy-stack regarding manage route before being submitted

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/BiContractActivationWindow.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/BiContractActivationWindow.jsx
@@ -1,3 +1,4 @@
+// @ts-check
 import React, { useState, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import { useClient } from 'cozy-client'
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/BiContractActivationWindow.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/BiContractActivationWindow.jsx
@@ -42,7 +42,8 @@ const BIContractActivationWindow = ({
       const result = await konnectorPolicy.fetchExtraOAuthUrlParams({
         client,
         account,
-        konnector
+        konnector,
+        manage: true
       })
       setExtraParams(result)
     }
@@ -77,6 +78,7 @@ const BIContractActivationWindow = ({
           intentsApi={intentsApi}
           onSuccess={onPopupClosed}
           onCancel={onPopupClosed}
+          manage={true}
         />
       )}
     </ListItem>
--- a/packages/cozy-harvest-lib/src/components/OAuthWindow.jsx
+++ b/packages/cozy-harvest-lib/src/components/OAuthWindow.jsx
@@ -47,8 +47,15 @@ export class OAuthWindow extends PureComponent {
   }
 
   componentDidMount() {
-    const { client, konnector, redirectSlug, extraParams, reconnect, account } =
-      this.props
+    const {
+      client,
+      konnector,
+      redirectSlug,
+      extraParams,
+      reconnect = false,
+      manage = false,
+      account
+    } = this.props
     this.realtime = new CozyRealtime({ client })
     this.realtime.subscribe(
       'notified',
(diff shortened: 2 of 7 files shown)


doubleface:
Works for me

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/BiContractActivationWindow.jsx
```diff
@@ -77,6 +78,7 @@ const BIContractActivationWindow = ({
           intentsApi={intentsApi}
           onSuccess={onPopupClosed}
           onCancel={onPopupClosed}
+          manage={true}
         />
       )}
     </ListItem>
```
Crash--, line 84: Ne devrait-on pas réserver ce manage true que quand les webhooks sont actifs ?

packages/cozy-harvest-lib/src/components/OAuthWindow.jsx
```diff
@@ -195,6 +203,8 @@ OAuthWindow.propTypes = {
   redirectSlug: PropTypes.string,
   /** Is it a reconnection or not */
   reconnect: PropTypes.bool,
+  /** Is it a manage or not */
+  manage: PropTypes.bool,
   /** Existing account */
   account: PropTypes.object,
   /** custom intents api. Can have fetchSessionCode, showInAppBrowser, closeInAppBrowser at the moment */
```
Crash--, line 210: I think that we should have a  "lower" `OAuthWindow` component. And create a new `BIOAuthWindow` for instance because if I need to develop an OAuth Konnector and then I want to "edit" the account, reading "Is it a manage or not" I will set it to `true`... 

But manage & reconnect are very specific to BI not standard OAuth. Same for `getOauthData`

packages/cozy-harvest-lib/src/helpers/oauth.js
```diff
@@ -160,6 +167,7 @@ const getAppSlug = client => {
  * @param  {string} redirectSlug The app we want to redirect the user on after the end of the flow
  * @param  {Object} extraParams some extra parameters to add to the query string
  * @param  {Boolean} reconnect Are we trying to reconnect an existing account ?
+ * @param  {Boolean} manage Are we trying to manage an existing account ?
```
Crash--, line 170: can you add 'for bi account' or something like that please ?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/ConfigurationTab/BiContractActivationWindow.jsx
doubleface: Non parce que BIContractActivationWindow n'est affichée que quand les webviews sont actives et ça n'est pas directement lié au full webhook

packages/cozy-harvest-lib/src/components/OAuthWindow.jsx
doubleface: Reconnect is actually not specific to BI and could be used for the Google connector for example to renew an OAuth token when refresh token is expired. 
And from the stack point of view, the manage route could also be used outside of BI (but yes, I don't have a real use case here for the moment). That is why I did not mention BI in this component.

I will try to make my descriptions better

packages/cozy-harvest-lib/src/helpers/oauth.js
doubleface: I would prefer not to do that regarding https://github.com/cozy/cozy-libs/pull/1852#discussion_r1001466358

doubleface:
@Crash-- Added ts-check and modification depending on it in BIContractActivationWindow

packages/cozy-harvest-lib/src/services/biWebView.spec.js
```diff
@@ -378,23 +378,27 @@ describe('fetchExtraOAuthUrlParams', () => {
...
-      account
+      account: { ...account, data: { auth: { bi: { connId: 15 } } } },
+      manage: true
+    })
+    expect(result).toMatchObject({
+      connection_id: 15,
+      code: 'bi-temporary-access-token-12-manage'
     })
-    expect(result).not.toHaveProperty('connection_id')
   })
```
Crash--, line 402: Le test qui testait le fait de n'avoir ni reconnect ni manage n'est plus utile ? celui qui regardait que connection_id n'était pas set?

Crash-- (approved):

packages/cozy-harvest-lib/src/services/biWebView.spec.js
Crash--: @doubleface Un avis ? Une réponse ? 🙏

PR #1863, opened 2022-10-27
docs: Document publishing beta versions

sblaisot:
This PR documents how to publish beta versions of an app in `cozy-apps-registry` using `cozy-app-publish`

### DIFF
--- a/packages/cozy-app-publish/README.md
+++ b/packages/cozy-app-publish/README.md
@@ -42,6 +42,11 @@ yarn cozy-app-publish \
 --build-commit $BUILD_COMMIT
 ```
 
+Published version is inferred from the tag set on current commit or the version defined in the manifest `version` property:
+
+- If the current commit has no tag associated, then this is a dev version in the form `<manifest_version>-dev.<commit_sha><date>`
+- If the current commit has a beta tag `x.y.z-beta.n` or a stable tag `x.y.z`, then this is a beta or stable version and the tag value is used as the version
+
 ### Manual usage (not recommended)
 
 First of all, don't forget to build the application:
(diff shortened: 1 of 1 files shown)


packages/cozy-app-publish/README.md
```diff
@@ -42,6 +42,11 @@ yarn cozy-app-publish \
 --build-commit $BUILD_COMMIT
 ```
 
+Published version is inferred from the tag set on current commit or the version defined in the manifest `version` property:
+
+- If the current commit has no tag associated, then this is a dev version in the form `<manifest_version>-dev.<commit_sha><date>`
+- If the current commit has a beta tag `x.y.z-beta.n` or a stable tag ` x.y.z`, then this is a beta or stable version and the tag value is used as the version
```
Crash--, line 48:
```suggestion
- If the current commit has a beta tag `x.y.z-beta.n` or a stable tag `x.y.z`, then this is a beta or stable version and the tag value is used as the version
```

packages/cozy-app-publish/README.md
```diff
@@ -60,6 +65,43 @@ yarn cozy-app-publish \
 --manual-version 1.0.2-dev.042cef26d9d33ea604fe4364eaab569980b500c9
 ```
 
+#### Publishing a beta version
+
+Beta versions are only available through beta channel of the registry and are not automatically deployable on production instances. However a beta tester can
+force deployment of beta versions of a given (installed) app by clicking trhee times in a row on the app version in the app store.
```
Crash--, line 71:
```suggestion
force deployment of beta versions of a given (installed) app by clicking three times in a row on the app version in the Cozy store.
```

Are we sure we want to document publicly that stuff? I've nothing against it. :)

Crash-- (approved):
Thank you @sblaisot !

packages/cozy-app-publish/README.md
sblaisot: Resolved in  4587fd0

PR #1866, opened 2022-11-04
fix: Close Harvest modal with the close button in error dialog

doubleface:
image

This close button did not do anything on screen. I think the expected action here is to really close the Harvest dialog.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountModal.jsx
@@ -147,7 +147,7 @@ export class AccountModal extends Component {
                 actionButton={
                   <Button
                     theme="danger"
-                    onClick={this.loadSelectedAccountId.bind(this)}
+                    onClick={onDismiss}
                     label={t('modal.konnector.error.button')}
                   />
                 }


Crash--:
> I think the expected action here is to really close the Harvest dialog.

Did you check with @joel-costa & the PO?

doubleface:
@Crash-- I did check with the PO. But I don't think I need to check with @joel-costa if a close button needs to close the current dialog

Crash--:
Also, there is an issue with your screenshot: some messages are in french, but not the other?

Crash--:
> But I don't think I need to check with @joel-costa if a close button needs to close the current dialog

Because you assume that the button is a close button. But what if the real issue is to have to wrong translation here? And that we should try to load again the information? 

Also, if this error is displayed, how many buttons do we have to close the modal? Three? ... Really, I think that having @joel-costa point of view on this subject can be useful.

Crash--:
ping @doubleface

doubleface:
@Crash-- Est-ce qu'on peut merger cette PR et je peux en faire une autre pour localiser le message [no matching trigger found](https://github.com/cozy/cozy-libs/blob/33270d3a9050f5754c337f7352b597238ca8d0f2/packages/cozy-harvest-lib/src/components/AccountModal.jsx#L82) ?

PR #1868, opened 2022-11-07
refactor: Remove unneeded parameter from getCozyBankIds

doubleface:
The bank id from account is not needed anymore since all bank ids
from the manifest are sent anyway
And moreover, the next version of the banking connectors will not use
the bankids for sentry but directly the slug of the connector


Crash-- (approved):

Crash--:
ping @doubleface

Crash--:
@doubleface still needed?

doubleface:
I checked. This PR is not needed anymore since the affected code has already been removed

PR #1872, opened 2022-11-09
fix: Paginated permission were not normalized

zatteo:
Fix a bug where we could not change permission after refreshing the modal with a lot of files.

We need this PR from cozy-client https://github.com/cozy/cozy-client/pull/1262


Crash-- (approved):
🎉

PR #1874, opened 2022-11-10
Feat: Add share button with recipients

cballevre:
In the cozy-notes with a lot of notes and sharing, there are some late requests. The ShareButton button doesn't wait for everything to be finished to appear. The user may be confused because an already shared note appears unshared first. I take this opportunity to pool two components in order to manage their state together.

**Related PRs:**
- https://github.com/cozy/cozy-ui/pull/2286

### DIFF
--- a/packages/cozy-sharing/package.json
+++ b/packages/cozy-sharing/package.json
@@ -37,12 +37,11 @@
   },
   "devDependencies": {
     "@babel/core": "7.16.12",
-    "@material-ui/core": "4.12.3",
     "babel-jest": "26.6.3",
     "babel-plugin-css-modules-transform": "1.6.2",
     "babel-plugin-inline-react-svg": "1.1.2",
     "cozy-client": "34.1.0",
-    "cozy-ui": "62.1.2",
+    "cozy-ui": "77.6.0",
     "enzyme": "3.11.0",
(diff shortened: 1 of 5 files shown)


packages/cozy-sharing/package.json
```diff
@@ -52,10 +51,9 @@
     "react-router": "^5.0.1"
   },
   "peerDependencies": {
-    "@material-ui/core": "4",
     "cozy-client": "^34.1.0",
     "cozy-realtime": "^3.11.0",
-    "cozy-ui": "^45.1.0",
+    "cozy-ui": "^77.6.0",
```
Crash--, line 56:
```suggestion
    "cozy-ui": ">=77.6.0",
```

(yes others are not right, but hey...)

packages/cozy-sharing/package.json
```diff
@@ -37,7 +37,6 @@
   },
   "devDependencies": {
     "@babel/core": "7.16.12",
-    "@material-ui/core": "4.12.3",
     "babel-jest": "26.6.3",
```
Crash--, line 40: You need to upgrade cozy-ui inside the devDep too 🙏

Crash-- (approved):

PR #1877, opened 2022-11-14
fix: Make cozy-harvest-lib works with cozy-keys-lib V4

Ldoppea:
In cozy/cozy-keys-lib@1e715624654bcd7c66f6fe7d6558d782f336d0fd we
refactored `<VaultUnlockProvider />` so we can inject
`checkShouldUnlock` into it

This method should allow the component to check if the Cozy's vault
need to be unlock before accessing Harvest

We choose to inject it from the parent component as we expect some
scenario to implement their own verification process (i.e: cozy-drive
for folder encryption)

Here we don't need a custom implementation, so we use the one provided
by CozyUtils


Crash-- (approved):
Nice <3

(just an import to remove in order to make the linter happy)

doubleface:
I tested it on home application and it fixes the infinite rerender problem when no vault is set. Please  just fix the lint problem

doubleface (approved):

Crash--:
After seeing why the CI is red now, I'm re-requesting a review after the fix ;).

Crash-- (approved):
Nice job! 👏

PR #1878, opened 2022-11-15
fix: Dispatch account deletion after BI connection deletion

doubleface:
This store update will espacially allow banks to update the
connected/disconnected status of it's accounts without adding realtime
on accounts, which causes more troubles for OAuthWindow component which
will unmount.

This is a quick fix. A better approach is in progress to make
OAuthWindow independent of Harvest components tree


Crash-- (approved):
Je ne comprends plus rien à ce composant, à ce qu'il fait ni comment. Mais si tu es confiant et que ça fait ce qu'on demande...

PR #1888, opened 2022-11-28
fix: Prevent createTemporaryToken to updateCache twice at the same time

Ldoppea:
`createTemporaryToken` is responsible for retrieving a temporary token from the Cozy, and if this token does not exist or is expired then it will create a new one

This method may be called from multiple different components

In some scenario, the call may be simultaneous. This is problematic because the cozy-stack would reject the 2nd update query as it would detect simultaneous database writes

To prevent this we wrap the method into a PromiseCache mechanism that would deduplicate calls

___

> **Warning**
> This PR fixes the bug that prevented the `HANDLE MY SYNCHRONISATION` button to be enabled
> This PR does not pre-load the token on dialog's opening
> This PR does not automatically refresh the token after 30min (which is partially mitigated by the first bullet point)

image

___

> **Note**
> Our first approach was to wrap those component into a React Provider that would be responsible to mutualise the same `temporaryToken` between components
> This would add too much complexity to the code due to the many Harvest entry points
> Eligible components that have enough information to declare the provider are `<TriggerManager>/<LegacyTriggerManager>`, `<KonnectorAccountTabs>` and `<Contracts>/<ContractsForAccount>`.
> But `<Contracts>/<ContractsForAccount>` is problematic here as `Contracts` it is top level from `<DisconnectedAccountModal>` but `<ContractsForAccount` is also embeded inside of `<KonnectorAccountTabs>`. So it would result into nested Providers
> Also we have some no-react entry points (i.e: Harvest CLI), intent or HOC. So we would have to check if the Provider exist or not every time.
> Finally, `<AccountModal>` seems to be the only place where the simultaneous call is possible (due to parallel hook + components)

Here is a map of all components/methods that may result to a `createTemporaryToken` call
- Harvest Routes
  - `<NewAccountModal>` (`${konnectorRoot}/new`)
    - `<TriggerManager>`
      - `<OAuthForm>`
        - `useOAuthExtraParams` (hook)
          - `biWebView.js` ou `budget-insight.js`
            - `fetchExtraOAuthUrlParams()`
              - 🎯 **createTemporaryToken**
  - `<EditAccountModal>` (`${konnectorRoot}/accounts/:accountId/edit`)
    - `<TriggerManager>`
      - `<OAuthForm>`
        - `useOAuthExtraParams` (hook)
          - `biWebView.js` ou `budget-insight.js`
            - `fetchExtraOAuthUrlParams()`
              - 🎯 **createTemporaryToken**     
      - `biWebview.handleOAuthAccount()` ou `budget-insight.handleOAuthAccount()`
        - `ConnectionFlow.createOrUpdateAccount()`
          - `onAccountCreation()`
            - `onBIAccountCreation()`
              - `createOrUpdateBIConnection()`
                - 🎯 **createTemporaryToken**
  - `<AccountModal>` (`${konnectorRoot}/accounts/:accountId`)
    - `<KonnectorAccountTabs>`
       - `useOAuthExtraParams` (hook)
          - `biWebView.js` ou `budget-insight.js`
            - `fetchExtraOAuthUrlParams()`
              - 🎯 **createTemporaryToken**     
      - `<ConfigurationTab>`
        - `<ContractsForAccount>`
          - `<BIContractActivationWindow>`
            - `biWebView.js`
              - `refreshContracts()`
                - 🎯 **createTemporaryToken**
              - `fetchExtraOAuthUrlParams()`
                - 🎯 **createTemporaryToken**
- CozyBanks (cozy-banks/AccountsListSettings)
  - `<DisconnectedAccountModal>`
    - `<Contracts>`
      - `<BIContractActivationWindow>`
        - `biWebView.js`
          - `refreshContracts()`
            - 🎯 **createTemporaryToken**
          - `fetchExtraOAuthUrlParams()`
            - 🎯 **createTemporaryToken**
- cli.js
  - `ConnectionFlow.createOrUpdateAccount()`
    - `budget-insight.js`
      - `onAccountCreation()`
        - `onBIAccountCreation()`
          - `createOrUpdateBIConnection()`
            - 🎯 **createTemporaryToken**

It may also be called in those unused functions (those will be deleted in another PR):
- `withKonnectorModal()`
  - `<KonnectorModal>`
    - `<TriggerManager>`
      - `<OAuthForm>`
        - `useOAuthExtraParams` (hook)
          - `biWebView.js` ou `budget-insight.js`
            - `fetchExtraOAuthUrlParams()`
              - 🎯 **createTemporaryToken**     
      - `biWebview.handleOAuthAccount()` ou `budget-insight.handleOAuthAccount()`
        - `ConnectionFlow.createOrUpdateAccount()`
          - `onAccountCreation()`
            - `onBIAccountCreation()`
              - `createOrUpdateBIConnection()`
                - 🎯 **createTemporaryToken**
    - `<KonnectorAccountTabs>`
       - `useOAuthExtraParams` (hook)
          - `biWebView.js` ou `budget-insight.js`
            - `fetchExtraOAuthUrlParams()`
              - 🎯 **createTemporaryToken**     
      - `<ConfigurationTab>`
        - `<ContractsForAccount>`
          - `<BIContractActivationWindow>`
            - `biWebView.js`
              - `refreshContracts()`
                - 🎯 **createTemporaryToken**
              - `fetchExtraOAuthUrlParams()`
                - 🎯 **createTemporaryToken**
- `budget-insight.js`
  - `getUserConfig`
    - 🎯 **createTemporaryToken**
- `budget-insight.js`
  - `updateUserConfig`
    - 🎯 **createTemporaryToken**

### DIFF
--- a/packages/cozy-harvest-lib/src/services/biWebView.js
+++ b/packages/cozy-harvest-lib/src/services/biWebView.js
@@ -9,6 +9,7 @@ import { getBIConnectionAccountsList, getBIConnection } from './bi-http'
 import assert from '../assert'
 import logger from '../logger'
 import { Q } from 'cozy-client'
+import PromiseCache from 'cozy-client/dist/promise-cache'
 // @ts-ignore (its a peerDep and I don't know how to configure ts for that)
 import flag from 'cozy-flags'
 
@@ -22,6 +23,8 @@ import { KonnectorJobError } from '../helpers/konnectors'
 import { waitForRealtimeEvent } from './jobUtils'
 import '../types'
 import { LOGIN_SUCCESS_EVENT } from '../models/flowEvents'
+
+const promiseCache = new PromiseCache()
 
 /**
  * @typedef {import("../models/ConnectionFlow").default} ConnectionFlow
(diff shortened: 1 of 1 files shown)


Crash--:
> In some scenario, the call may be simultaneous. 

This is typically what we'll need to know in a few months: What are the concerned scenarios? 🙏

Crash-- (approved):

doubleface (approved):

PR #1889, opened 2022-11-28
Feat: Follow current trigger jobs event after the first load

doubleface:
Now follow the future updates of the current trigger by following any
job related to it with realtime

Anytime a new job related to the current trigger is detected, the
ConnectionFlow will detect it, refetch the current trigger, to let the
stack determine the current job (if multiple jobs are running at the same
time), and follow the new current job.

This will show the current trigger is running to the user and update the
result at the end of the job

- refactor: Use cozy-client plugin instead of cozy-realtime
- fix: Use flowEvent events
- refactor: Do not use lodash.get anymore
- feat: Follow current trigger jobs even after the first load
- docs: Add @ts-check and fix jsdoc and code accordingly


packages/cozy-harvest-lib/src/models/konnector/KonnectorJobWatcher.js
```diff
@@ -85,7 +86,7 @@ export class KonnectorJobWatcher {
     this.disableSuccessTimer()
     if (this._error || this._succeed) return
 
-    this.emit('loginSuccess', this.job)
+    this.emit(LOGIN_SUCCESS_EVENT, this.job)
   }
 
   disableSuccessTimer() {
```
Crash--, line 92: 👏

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -483,9 +486,16 @@ export class ConnectionFlow {
...
   }
 
-  async handleJobUpdated() {
+  handleCurrentJobUpdated() {
     logger.debug('ConnectionFlow: Handling update from job')
-    await this.refetchTrigger()
+    this.refetchTrigger()
+  }
+
+  handleTriggerJobUpdated(job) {
```
Crash--, line 512: nit: I'll have created two methods: 

- listenJobsEvent 
- And this listenJobsEvent will have passed down the `job` to `handleJobsUpdateForCurrentTrigger` or something similar.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -483,9 +486,16 @@ export class ConnectionFlow {
...
+    if (job.trigger_id !== this.trigger?._id) return
+    if (job._id === this.job?._id) return
```
Crash--, line 514: having a comment to explain why we need to do that can be pretty useful

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -624,12 +637,32 @@ export class ConnectionFlow {
     }
   }
 
+  watchTriggerJobs() {
+    // can happen when the current trigger comes from realtime
```
Crash--: to not have `this.trigger` set? 

I don't understand clearly this comment

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -624,12 +637,32 @@ export class ConnectionFlow {
...
+    }
+  }
+
   watchJob(options = { autoSuccessTimer: false }) {
+    if (this.job?._id === this.jobWatcher?.job?._id) {
+      return
+    }
+    if (this.jobWatcher) {
+      this.jobWatcher.unsubscribeAll()
+    }
```
Crash--, line 689: Can you explain those both changes please?

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -637,9 +663,12 @@ export class ConnectionFlow {
     }
   }
 
+  /**
+   * Watch all jobs related to the current trigger
+   */
   watchTriggerJobs() {
-    // can happen when the current trigger comes from realtime
     if (this.trigger) {
+      // can happen when the current trigger comes from realtime
```
Crash--, line 671: You can say `When trigger comes from realtime, cozy-stack doesn't add the current_state to the object. So we need to request the stack to get it`

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: The comment finished in the wrong place, sorry. It is `this.trigger?.current_state` which can be empty

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -296,6 +303,13 @@ export class ConnectionFlow {
     })
   }
 
+  /**
+   * Set a state to the flow to show that the current trigger is running while the job is not
+   * created yed (the job will be created with a webhook)
```
Crash--, line 308:
```suggestion
   * created yed (the job will be created by a webhook)
```

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
Crash--: What is the purpose of `watchJob`, `watchKonnectorJob` and `watchTriggerJobs`? 

Why don't we subscribe directly to realtime event inside the constructor?

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -617,20 +655,45 @@ export class ConnectionFlow {
...
 
+  /**
+   * Watch all jobs related to the current trigger
+   */
+  watchTriggerJobs() {
+    if (this.trigger) {
+      // When the trigger comes from realtime, cozy-stack does not add the current_state to the object. So we need to request the stack to get it
+      if (!this.trigger?.current_state) {
+        this.refetchTrigger()
+      }
```
Crash--, line 674: It seems that `watchTriggerJobs` is only called from the constructor: https://github.com/cozy/cozy-libs/blob/5e25cd6da63b93558bf3609d4230153097476ca8/packages/cozy-harvest-lib/src/models/ConnectionFlow.js#LL185C18-L185C18 

How can we have a trigger without `current_state` at this point?  Can you explain how it can happen?

Crash--: Is it because when it's the first run of the trigger, then we don't have yet a `current_state`?

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -497,6 +527,10 @@ export class ConnectionFlow {
     const trigger = await fetchTrigger(this.client, this.trigger._id)
     logger.debug(`Refetched trigger`, trigger)
     this.trigger = trigger
+
+    this.watchCurrentJobIfTriggerIsAlreadyRunning()
```
Crash--, line 531: Do we really want to only `watchJob` if the trigger is running? Do we need to subscribe here to the realtime? Don't we have already subscribe in a way or other?

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -617,20 +655,45 @@ export class ConnectionFlow {
...
+        this.handleTriggerJobUpdated
+      )
+    }
+  }
+
   watchJob(options = { autoSuccessTimer: false }) {
+    if (this.job?._id === this.jobWatcher?.job?._id) {
+      // no need to rewatch a job we are already watching
+      return
+    }
```
Crash--, line 687: We should not call `watchJob` if the job is already watched. This should be the real fix.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -617,20 +655,45 @@ export class ConnectionFlow {
...
+    if (this.jobWatcher) {
+      // if no job has been watched yet, there is not jobWatcher yet.
+      this.jobWatcher.unsubscribeAll()
```
Crash--, line 690: I still don't understand why we need to `unsubscribeAll()` here?

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: This trigger instance comes from the host application of harvest (home for example) which can get it from a realtime event, which does not harvest current_state

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: The role of the job watcher is to watch a specific job with a specified id. Since in this case the id of the job to monitor has changed, we tell to  je job watcher to unsubscribe from events associated to the previous job id.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: `Watchjob` is called from `watchCurrentJobIfTriggerIsAlreadyRunning` and from `launch`. In the case of launch, we call `watchjob` directly with the job we received directly from the trigger launch which in this case is in double with `watchCurrentjobIfTriggerIsAlreadyRunning`.

I could remove the call to `watchjob` in `launch` but that would make `watchjob` totally dependent of realtime which I think is not a good solution.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: I added jsdoc to watchJob now but here it is :

* watchJob uses watchKonnectorJob to watch a specific job, it also relays events from KonnectorJobWatcher instance to the current ConnectionFlow instance (like LOGIN_SUCCESS after 8 seconds)
* watchKonnectorJob creates a new instance of KonnectorJobWatcher to watch a specified job with a given id. It also handle the autoSucessFullTimer option which triggers LOGIN_SUCCESS after 8 second when true
* watchTriggerJobs : detects new jobs associated to the current trigger. If any, it refetches the current trigger which will set the new job as the current job to watch with watchJob

Crash-- (approved):

PR #1896, opened 2022-11-30
fix: Add konnector slug in fake trigger passed to onLoginSuccess

doubleface:
This information is needed by a b2c app

### DIFF
--- a/packages/cozy-harvest-lib/src/components/FlowProvider.jsx
+++ b/packages/cozy-harvest-lib/src/components/FlowProvider.jsx
@@ -141,7 +141,8 @@ export class FlowProvider extends Component {
     // created, this is why a trigger like object is created here
     const triggerLike = {
       message: {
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/components/FlowProvider.jsx
```diff
@@ -141,7 +141,8 @@ export class FlowProvider extends Component {
     // created, this is why a trigger like object is created here
     const triggerLike = {
       message: {
-        account: flow?.account?._id
+        account: flow?.account?._id,
+        konnector: flow?.konnector?.slug
```
Crash--, line 145: What is the real trigger.message? Should we add others keys?

doubleface: This is the real trigger message keys : account and konnector

PR #1928, opened 2022-12-19
fix: Reactive share modal for all apps

zatteo:
The previous check was too strong and the Share modal disappeared in some apps like Drive.

I also added unit test to check the good behavior because it can make crazy.

### DIFF
--- a/packages/cozy-sharing/src/components/ShareModal.jsx
+++ b/packages/cozy-sharing/src/components/ShareModal.jsx
@@ -30,64 +30,54 @@ export const ShareModal = ({
   sharingDesc,
   twoStepsConfirmationMethods
 }) => {
-  const showShareCozyToCozyForNotes =
-    flag('notes.sharing-cozy-to-cozy') !== null &&
-    flag('notes.sharing-cozy-to-cozy') &&
-    documentType === 'Notes'
+  const shareDialogOnlyByLink =
+    (documentType === 'Notes' && !flag('notes.sharing-cozy-to-cozy')) ||
+    documentType === 'Albums'
 
-  const showShareByEmail =
-    showShareCozyToCozyForNotes &&
-    documentType !== 'Albums' &&
-    !hasSharedParent &&
-    !hasSharedChild
+  const showShareByEmail = !hasSharedParent && !hasSharedChild
   const showShareByLink = documentType !== 'Organizations'
   const showShareOnlyByLink = hasSharedParent || hasSharedChild
   const showWhoHasAccess = documentType !== 'Albums'
 
-  return (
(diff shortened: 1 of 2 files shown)


packages/cozy-sharing/src/components/ShareModal.jsx
```diff
@@ -32,12 +32,11 @@ export const ShareModal = ({
...
   const showShareCozyToCozyForNotes =
     flag('notes.sharing-cozy-to-cozy') !== null &&
-    flag('notes.sharing-cozy-to-cozy') &&
-    documentType === 'Notes'
+    flag('notes.sharing-cozy-to-cozy')
 
   const showShareByEmail =
-    showShareCozyToCozyForNotes &&
-    documentType !== 'Albums' &&
+    ((documentType === 'Notes' && showShareCozyToCozyForNotes) ||
```
acezard, line 38: I think we should definitely make the conditional more readable from a business point of view

zatteo: @acezard we wrote a new version with @Ldoppea and @cballevre

packages/cozy-sharing/src/components/ShareModal.jsx
```diff
@@ -30,64 +30,54 @@ export const ShareModal = ({
   sharingDesc,
   twoStepsConfirmationMethods
 }) => {
-  const showShareCozyToCozyForNotes =
-    flag('notes.sharing-cozy-to-cozy') !== null &&
-    flag('notes.sharing-cozy-to-cozy') &&
-    documentType === 'Notes'
+  const shareDialogOnlyByLink =
```
acezard, line 33: I'm pretty sure I worked on this, so I'm not blaming anyone, but it's still too obscure to me.

Ideally, we should have some config object we act upon when we receive data.

Something like that for instance: 
```
const shareableOnlyByLink = ['notes', 'albums']

export const isShareableOnlyByLink = (type: DocumentType): boolean =>
  shareableOnlyByLink.includes(type)

isShareableOnlyByLink('notes') // true
```

cballevre (approved):

packages/cozy-sharing/src/components/ShareModal.jsx
zatteo: It would be more complicated to have a generic config object because the 'notes' app need a flag and the 'albums' app not. So I prefer to keep it like this. But it will definitively need some refactoring because there is some "onlyByLink" content in "cozyToCozy" part.

acezard (approved):

Ldoppea (approved):

PR #1939, opened 2022-12-29
Refactor/prepare update sharing modal

zatteo:
Migrating class components to functional components and updating query method.


packages/cozy-sharing/src/queries/queries.js
```diff
@@ -0,0 +1,45 @@
...
+              $size: 0
+            }
+          }
+        }
+      ]
+    })
+    .indexFields(['_id'])
+    .limitBy(1000),
+  options: {
+    as: 'contacts'
```
Crash--, line 50: sharings-io.cozy.contacts-query (or something similar, but not just `contacts`)

packages/cozy-sharing/src/queries/queries.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+})
+
+export const buildGroupsQuery = () => ({
+  definition: Q(Group.doctype),
+  options: {
+    as: 'groups'
```
Crash--, line 57: `io.cozy.contacts.groups` cf https://github.com/cozy/cozy-guidelines/#naming-of-queries

packages/cozy-sharing/src/SharingBanner/hooks/useSharingInfos.jsx
```diff
@@ -4,7 +4,7 @@ import { useClient, models } from 'cozy-client'
 
 import getQueryParameter from '../helpers/QueryParameter'
 import logger from '../../logger'
-import { buildSharingsByIdQuery } from '../helpers/queries'
+import { buildSharingsByIdQuery } from '../../queries/queries'
```
paultranvan, line 7: I think we want to avoid relative imports

packages/cozy-sharing/src/queries/queries.js
```diff
@@ -1,7 +1,21 @@
-import { Q } from 'cozy-client'
+import { Q, fetchPolicies } from 'cozy-client'
 
 import { Contact, Group } from '../models'
 
+const DEFAULT_CACHE_TIMEOUT_QUERIES = 9 * 60 * 1000
```
paultranvan, line 5: Why this value? I assume the `9 * 60` computation means something?
I'm also wondering if we should not expose such default value from cozy-client, so apps/libs don't have to come up with a new value every time

packages/cozy-sharing/src/components/ShareByLink.jsx
```diff
@@ -35,67 +35,56 @@ const checkIsReadOnlyPermissions = permissions => {
...
+  onEnable,
+  onDisable,
+  checked,
+  popperOptions
+}) => {
+  const { t } = useI18n()
+  const [loading, setLoading] = useState(false)
+  const [menuIsOpen, setMenuIsOpen] = useState(false)
+
+  const containerRef = React.createRef()
```
Ldoppea, line 53:
```suggestion
  const containerRef = useRef()
```

packages/cozy-sharing/src/SharingBanner/hooks/useSharingInfos.jsx
Crash--: > I think we want to avoid relative imports

This is not really true and this is not really false neither. 

We need relative import if the component needs to be overridden. For now, this lib is not concerned, yet. 

Also, we're within a lib, so in order to be able to do that, we need to transform the path only at the transpilation/publish time. (see https://github.com/cozy/cozy-ui/blob/master/babel.config.js#L35-L36 for instance)

Since this is not the scope of this PR, let's move on ;)

packages/cozy-sharing/src/queries/queries.js
Crash--: Same, this is not really the scope of this PR so we can move on without doing anything but: 

I assume that this value is copy pasted from Drive because Drive has to manipule `io.cozy.files` with `thumbnails` links only valid for 10 minutes. So we cache the value only for 9 minutes (9 * 60 * 1000). 

Maybe cozy-client should expose default cache value for a few doctypes + unlimited cache. https://github.com/cozy/cozy-client/issues/1253

packages/cozy-sharing/src/SharingBanner/hooks/useSharingInfos.jsx
zatteo: Thank you for the answer!

packages/cozy-sharing/src/components/ShareByEmail.jsx
```diff
@@ -16,38 +16,38 @@ import styles from '../share.styl'
...
+    setSelectedOption(value)
   }
 
-  onRecipientPick = recipient => {
+  const onRecipientPick = recipient => {
     let contactsToAdd
     if (recipient._type === Group.doctype) {
       const groupId = recipient.id
-      contactsToAdd = this.props.contacts.data.filter(contact => {
+      contacts.data.filter(contact => {
```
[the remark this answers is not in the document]
zatteo: Very good catch!

Crash-- (approved):

PR #1942, opened 2023-01-02
Update share by email buttons

zatteo:
- Hide ShareByEmail controls when input not focused or empty
- Add "Who has access?" title in Dialog

### DIFF
--- a/packages/cozy-sharing/src/components/ShareAutosuggest.jsx
+++ b/packages/cozy-sharing/src/components/ShareAutosuggest.jsx
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React, { useState, useRef } from 'react'
 import PropTypes from 'prop-types'
 import Autosuggest from 'react-autosuggest'
 
@@ -18,24 +18,22 @@ import {
 import ContactSuggestion from './ContactSuggestion'
 import { extractEmails, validateEmail } from '../helpers/email'
 
-export default class ShareAutocomplete extends Component {
-  state = {
-    inputValue: '',
-    suggestions: []
-  }
+const ShareAutocomplete = ({
+  loading,
+  contactsAndGroups,
+  recipients,
+  onFocus,
+  onPick,
+  onRemove,
+  placeholder
+}) => {
+  const [inputValue, setInputValue] = useState('')
+  const [suggestions, setSuggestions] = useState([])
+  const [isPasted, setIsPasted] = useState(false)
 
-  componentDidUpdate(prevProps) {
-    if (
-      this.props.contactsAndGroups.length !== prevProps.contactsAndGroups.length
-    ) {
-      this.onSuggestionsFetchRequested({
-        value: this.state.inputValue
-      })
-    }
-  }
+  const autosuggestRef = useRef(null)
 
-  computeSuggestions(value) {
-    const { contactsAndGroups } = this.props
+  const computeSuggestions = value => {
     const inputValue = value.trim().toLowerCase()
 
     return inputValue.length === 0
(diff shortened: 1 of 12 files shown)


packages/cozy-sharing/src/components/ShareAutosuggest.jsx
```diff
@@ -49,81 +45,71 @@ export default class ShareAutocomplete extends Component {
...
+      onAutosuggestPick(highlightedSuggestion)
+    } else if (inputValue !== '' && inputValue.match(/\S+@\S+/)) {
+      onAutosuggestPick({ email: inputValue })
     }
   }
 
-  onPick = value => {
-    this.props.onPick(value)
-    this.setState(state => ({ ...state, inputValue: '' }))
-    requestAnimationFrame(() => this.input && this.input.focus(), 1) // don't ask...
```
zatteo, line 117: If someone can explain me what was the purpose of this line ? I did not see a difference without it (first added here https://github.com/cozy/cozy-drive/commit/c7f95d9d68a66042c87e5acc3cb91e54b9f9e6a4#diff-deb41b822a10ac8f8f645674a1188a96a8ad503e9bf192c09d63c462d94480bcR53)

zatteo: It was about the `requestAnimationFrame` line. The `// don't ask` comment was already before the "CI purpose" commit. CI still passes so it is a good start.

packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
```diff
@@ -49,25 +47,19 @@ const ShareRecipientsInput = ({
   }
 
   return (
-    <div>
-      <label className={styles['coz-form-label']} htmlFor="email">
-        {label}
-      </label>
```
Crash--, line 55: Can you explain why you removed that one?

yarn.lock
```diff
@@ -17215,7 +17215,7 @@ rc@^1.2.7:
     minimist "^1.2.0"
     strip-json-comments "~2.0.1"
 
-react-autosuggest@9.4.3, react-autosuggest@^9.4.3:
+react-autosuggest@9.4.3:
```
Crash--, line 17218: Can you send the result of `yarn why react-autosuggest` ? I'm surprised to always see this `9.4.3` version of react-autosuggest within the lock file.

packages/cozy-sharing/src/components/ShareAutosuggest.jsx
Ldoppea: In general when you see a `setTimeout(..., 1)` in modern frameworks code, this is used to execute some code on the next rendering loop, so you are sure that every props have been propagated and are up to date. Maybe this is the case here?
Is the selection cursor at the same place with or without this line of code after the event?

yarn.lock
zatteo: It is from cozy-bar in mespapiers-lib

```
yarn why v1.22.19
[1/4] 🤔  Why do we have the module "react-autosuggest"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "react-autosuggest@10.1.0"
info Reasons this module exists
   - "_project_#cozy-sharing" depends on it
   - Hoisted from "_project_#cozy-sharing#react-autosuggest"
info Disk size without dependencies: "248KB"
info Disk size with unique dependencies: "972KB"
info Disk size with transitive dependencies: "1.07MB"
info Number of shared dependencies: 9
=> Found "cozy-bar#react-autosuggest@9.4.3"
info This module exists because "_project_#cozy-mespapiers-lib#cozy-bar" depends on it.
info Disk size without dependencies: "380KB"
info Disk size with unique dependencies: "728KB"
info Disk size with transitive dependencies: "1008KB"
info Number of shared dependencies: 9
✨  Done in 0.63s.
```

packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
zatteo: UI changed and there is no label here anymore. Do I need to explain it in the commit message?

yarn.lock
Crash--: OK thanks for the information 👍

packages/cozy-sharing/src/components/ShareAutosuggest.jsx
zatteo: Thank you for your answer!

- I understand better the setTimeout x next rendering loop thing
- I discovered one case (and maybe there are more) where it was useful to focus (when you remove a recipient, focus stays on the input)

Ldoppea (approved):

PR #1943, opened 2023-01-02
New sharing modal

zatteo:
New sharing modal with : 
- updated UI/UX
- a native sharing button on mobile
- an animation when a new recipient is added


Crash-- (approved):
This PR has already been reviewed branch per branch previously, so let's merge it when CI is green ;)

PR #1959, opened 2023-01-09
Feat/polish sharing modal

zatteo:
Improve spacing between Who has access title and components
Update wording when modifying file or folder
Propose in autosuggest unknown email


packages/cozy-sharing/src/components/EditLinkPermissionDialog.jsx
```diff
@@ -62,12 +62,20 @@ const EditLinkPermissionDialog = ({
         >
           <FormControlLabel
             value="true"
-            label={t('Share.permissionLink.read')}
+            label={
+              document.type === 'file'
```
Crash--, line 66: Can't you have a document without `type`? `type` is only there for `io.cozy.files` but we can share others data.

zatteo: Right! And now file text is default version.

Crash-- (approved):

PR #1967, opened 2023-01-11
fix(mes-papiers-lib): Check file's `lastModified` field on file upload

Ldoppea:
On iOS, taking a photo with `FileInput capture="environment" />` will always return a file named `image.jpg`

In previous implementation we compared the image's name to prevent duplication which would not work on iOS as all files would have the same name. Result is that we wouldn't be able to upload multiple photo

To prevent this we also want to compare the image's `lastModified` field that should be unique for each photo capture

### DIFF
--- a/packages/cozy-mespapiers-lib/src/components/ModelSteps/AcquisitionResult.jsx
+++ b/packages/cozy-mespapiers-lib/src/components/ModelSteps/AcquisitionResult.jsx
@@ -17,6 +17,7 @@ import { useStepperDialog } from '../Hooks/useStepperDialog'
 import { useFormData } from '../Hooks/useFormData'
 import { PaperDefinitionsStepPropTypes } from '../../constants/PaperDefinitionsPropTypes'
 import { KEYS } from '../../constants/const'
+import { isSameFile } from './helpers'
 
 const isPDF = file => file.type === 'application/pdf'
 
@@ -42,8 +43,7 @@ const AcquisitionResult = ({ currentFile, setCurrentFile, currentStep }) => {
 
   const changeSelectedFile = () => {
     const newData = formData.data.filter(
-      data =>
-        data.file.name !== currentFile.name || data.file.id !== currentFile.id
+      data => !isSameFile(currentFile, data.file)
     )
 
     setFormData(prev => ({
--- a/packages/cozy-mespapiers-lib/src/components/ModelSteps/AcquisitionResult.jsx
+++ b/packages/cozy-mespapiers-lib/src/components/ModelSteps/AcquisitionResult.jsx
@@ -42,7 +42,8 @@ const AcquisitionResult = ({ currentFile, setCurrentFile, currentStep }) => {
 
   const changeSelectedFile = () => {
     const newData = formData.data.filter(
-      data => data.file.name !== currentFile.name
+      data =>
+        data.file.name !== currentFile.name || data.file.id !== currentFile.id
     )
 
     setFormData(prev => ({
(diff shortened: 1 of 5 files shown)


Merkur39 (approved):

packages/cozy-mespapiers-lib/src/components/ModelSteps/helpers.js
```diff
@@ -0,0 +1,20 @@
...
+  for (const data of formData.data) {
+    if (data.stepIndex === stepIndex) {
+      if (currentFile.constructor.name === 'File') {
+        return (
+          data.file.name === currentFile.name &&
+          data.file.lastModified === currentFile.lastModified
+        )
+      }
+    }
+    return false
```
PolariTOON, line 21: Shouldn't this `return` statement be outside of the `for` loop?

packages/cozy-mespapiers-lib/src/components/ModelSteps/helpers.js
```diff
@@ -0,0 +1,20 @@
...
+ * @returns
+ */
+export const isFileAlreadySelected = (formData, stepIndex, currentFile) => {
+  for (const data of formData.data) {
+    if (data.stepIndex === stepIndex) {
+      if (currentFile.constructor.name === 'File') {
+        return (
+          data.file.name === currentFile.name &&
+          data.file.lastModified === currentFile.lastModified
+        )
```
PolariTOON, line 18: The next entries in `formData` will be skipped no matter if the name and the date match. Is this intended?

packages/cozy-mespapiers-lib/src/components/ModelSteps/helpers.js
```diff
@@ -8,6 +8,9 @@
 export const isFileAlreadySelected = (formData, stepIndex, currentFile) => {
   for (const data of formData.data) {
     if (data.stepIndex === stepIndex) {
+      if (currentFile.constructor.name === 'Blob') {
+        return data.file.id === currentFile.id
```
PolariTOON, line 12: Same question here: is returning `false` early intended if the `id`s are different?

packages/cozy-mespapiers-lib/src/components/ModelSteps/AcquisitionResult.jsx
```diff
@@ -42,7 +42,8 @@ const AcquisitionResult = ({ currentFile, setCurrentFile, currentStep }) => {
 
   const changeSelectedFile = () => {
     const newData = formData.data.filter(
-      data => data.file.name !== currentFile.name
+      data =>
+        data.file.name !== currentFile.name || data.file.id !== currentFile.id
```
PolariTOON, line 46: Shouldn't this also test the `lastModified` date? In case multiple files have the same name?

packages/cozy-mespapiers-lib/src/components/ModelSteps/helpers.js
Merkur39: Non, la logique ici n'a pas changée de l'ancienne implémentation. Si nous ne rentrons pas dans le cas du `stepIndex`, alors il ne faut pas aller plus loin.
Ce contrôle est utile pour les quelques cas (permis/CIN/..) où nous avons 2 étapes de scan

packages/cozy-mespapiers-lib/src/components/ModelSteps/helpers.js
Merkur39: Vu ensemble

packages/cozy-mespapiers-lib/src/components/ModelSteps/helpers.js
Merkur39: Vu ensemble

PolariTOON (approved):
Approuvé, sous réserve d'avoir les changements vus ensemble, à savoir revoir les helpers `isFileAlreadySelected` et `changeSelectedFile` pour avoir la même logique et éviter les faux négatifs / faux positifs.

PR #1970, opened 2023-01-13
Feat/handle sharing by link read only

zatteo:
feat: Prevent apps with only read only link to create read write link
feat: Add and export DOCUMENT_TYPE constant


packages/cozy-sharing/src/components/ShareByLink.jsx
```diff
@@ -60,7 +61,11 @@ const ShareByLink = ({ link, document, documentType, onEnable }) => {
   }
 
   const onCreate = async () => {
-    setIsEditDialogOpen(true)
+    if (isOnlyReadOnlyLinkAllowed({ documentType })) {
+      onPermissionsSelected({ verbs: ['GET'] })
+    } else {
+      setIsEditDialogOpen(true)
+    }
```
zatteo, line 68: => if only read only, we don't open the dialog but create directly the link

Crash-- (approved):

PR #1971, opened 2023-01-13
feat: Use navigator.share in mobile to share link

zatteo:


Crash--:
> Warning : Web Share API works only in HTTPS

This is not 100% accurate. The real definition is "needs a secure context". So let's take a look at that definition: https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts 

> Locally-delivered resources such as those with http://127.0.0.1 URLs, http://localhost and http://*.localhost URLs (e.g. http://dev.whatever.localhost/), and file:// URLs are also considered to have been delivered securely.
> Resources that are not local, to be considered secure, must meet the following criteria:
    - must be served over https:// or wss:// URLs
    - the security properties of the network channel used to deliver the resource must not be considered deprecated

packages/cozy-sharing/src/components/ShareByLink.jsx
```diff
@@ -78,6 +78,28 @@ const ShareByLink = ({ link, document, documentType, onEnable }) => {
     setAlert({ ...alert, open: false })
   }
 
+  const showMobileShareButtons = link && isMobile && navigator.share
+  const showDesktopShareButtons =
```
Crash--, line 82: Always confusing to have a var named `Desktop` but can be apply to a `mobile`.

Crash-- (approved):

packages/cozy-sharing/src/components/ShareByLink.jsx
zatteo: Right! And yesterday I read the "Naming" chapter of the Clean Code book...

PR #1972, opened 2023-01-16
Feat/fade in new recipients

zatteo:


Crash-- (approved):
I think there is something to improve on how we handle the `hadBeenJustAdded` because we should be able to leverage cozy-client cache instead of React ref, but for a first naive implementation, it should be ok 👍

zatteo:
> I think there is something to improve on how we handle the `hadBeenJustAdded` because we should be able to leverage cozy-client cache instead of React ref, but for a first naive implementation, it should be ok 👍

I tried first to integrate directly in cozy-sharing Redux store but it was more complicated. But it could be interesting with cozy-client cache. I merge it and we will see how it perform.

PR #2000, opened 2023-02-02
feat: Extract cccPolicy and do not create ccc triggers and accounts

doubleface:
Now all ccc specific code is gathered in cccPolicy.

Then ConnectionFlow can create an account, create a trigger, launch a trigger or not according to current konnector's policy

ccc specific code is still in `src/helpers/triggers.js` code since it will be used by flagship application to create accounts and triggers itself.


packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
```diff
@@ -59,7 +59,8 @@ export const LaunchTriggerAlert = ({
   const styles = useStyles({ block })
 
   const lastSuccessDate = getLastSuccessDate(trigger)
-  const isKonnectorRunnable = isRunnable({ win: window, konnector })
+  const anchorRef = useRef()
+  const [showOptions, setShowOptions] = useState(false)
```
Crash--, line 63: What is that?

packages/cozy-harvest-lib/src/konnector-policies.js
```diff
@@ -1,6 +1,7 @@
 import logger from './logger'
 import { konnectorPolicy as biKonnectorPolicy } from './services/budget-insight'
 import { konnectorPolicy as biWebViewPolicy } from './services/biWebView'
+import { konnectorPolicy as cccPolicy } from './services/cccPolicy'
```
Crash--, line 4: ccc means connecteur côté client, this is not the word we used. Name it clisk 🙏 . 

Also, other policies are namde `budget-insight` or `biWebview`. There is no `Policy` within the name.

packages/cozy-harvest-lib/src/konnector-policies.js
```diff
@@ -26,8 +32,5 @@ logger.info('Available konnector policies', policies)
  * @param {KonnectorManifest} konnector
  * @returns {KonnectorPolicy}
  */
-export const findKonnectorPolicy = konnector => {
-  const policy = policies.find(policy => policy.match(konnector))
-  logger.info(`Using ${policy.name} konnector policy for ${konnector.slug}`)
-  return policy
-}
+export const findKonnectorPolicy = konnector =>
```
Crash--, line 35: Why did you remove the log?

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -440,31 +440,51 @@ export class ConnectionFlow {
...
-      this.account = account
+        this.account = account
 
-      logger.info(`ConnectionFlow: Saved account ${account._id}`)
+        logger.info(`ConnectionFlow: Saved account ${account._id}`)
 
-      if (!konnectorPolicy.needsTriggerCreation) {
-        await this.ensureTriggerAndLaunch(client, {
+        logger.debug('ConnectionFlow: Ensuring trigger...')
+        const ensuredTrigger = await ensureTrigger(client, {
```
Crash--, line 460: why this change in this commit?

Crash--: Also, it should be called... `trigger` no?

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -369,138 +369,67 @@ describe('ConnectionFlow', () => {
...
         trigger: fixtures.existingTrigger,
         userCredentials: fixtures.credentials
       })
 
-      expect(flow.ensureTriggerAndLaunch).toHaveBeenCalledWith(
+      expect(ensureTrigger).toHaveBeenCalledWith(
         client,
         expect.objectContaining({ account: fixtures.updatedAccount })
       )
     })
```
Crash--, line 385: Since now this `ensureTiggerAndLaunch` has been split into several call, you should also test that `ensureDefaultFolderPathInAccount` has been called for instance. 

Once it's done, I'm interested in what is the expected behavior if `ensureTrigger` resole, but `ensureDefaultFolderPathInAccount` throw.

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -369,138 +369,67 @@ describe('ConnectionFlow', () => {
...
-      await flow.ensureTriggerAndLaunch(client, {
-        trigger: fixtures.createdTrigger,
-        konnector: fixtures.konnector
-      })
-      expect(launchTrigger).toHaveBeenCalledTimes(1)
-      expect(launchTrigger).toHaveBeenCalledWith(
-        client,
-        fixtures.createdTrigger
       )
     })
```
Crash--, line 424: You should also test the call to launch where this is not the : 
```
window.cozy = {
        ClientConnectorLauncher: 'react-native'
      }
```

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -571,64 +501,91 @@ describe('ConnectionFlow', () => {
...
-      expect(saveAccount).not.toHaveBeenCalled()
+      expect(saveAccount).toHaveBeenCalledTimes(1)
       expect(client.query).toHaveBeenCalledWith(
         Q('io.cozy.files').getById(
           fixtures.createdTriggerWithFolder.attributes.message.folder_to_save
         )
       )
     })
+    delete window.cozy
+    delete window.ReactNativeWebView
```
Crash--, line 526: Where are these variables defined? Is it because of https://github.com/cozy/cozy-libs/pull/2000/commits/9f251f802b0dd046d1e99768b3b402350baed1b8#diff-57ee38fdf45b61888d99d54f38f4dfe8bc42c88c8e3af8e1026d120a1b9af17fR407-R411 ? 

If yes, this is a red flag. Each test must be autonomous and have no side effect

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -571,64 +501,91 @@ describe('ConnectionFlow', () => {
...
+  })
+})
+
+describe('ensureTrigger', () => {
+  beforeAll(() => {
+    jest.spyOn(cronHelpers, 'fromFrequency').mockReturnValue('0 0 0 * * 0')
+  })
+
+  afterEach(() => {
+    jest.clearAllMocks()
```
Crash--, line 588: What is the goal of those lines?

packages/cozy-harvest-lib/src/services/cccPolicy.js
```diff
@@ -0,0 +1,64 @@
...
+const isCCC = konnector => Boolean(konnector.clientSide)
+
+/**
+ * Get's the launcher in the current environment if any
+ *
+ * @param {Object} win The window object in the current environment
+ * @returns {Object|null}
+ */
+export const getLauncher = ({ win }) =>
+  win?.cozy?.ClientConnectorLauncher || null
```
Crash--, line 22: Why do we need to pass the object in params? Can't we just read it from `window` directly?

packages/cozy-harvest-lib/src/konnector-policies.js
```diff
@@ -9,10 +10,15 @@ const defaultKonnectorPolicy = {
   match: () => true,
   name: 'default',
   fetchExtraOAuthUrlParams: null,
-  refreshContracts: null
+  refreshContracts: null,
+  needsTriggerLaunch: true,
+  needsAccountAndTriggerCreation: true,
+  onLaunch: null,
+  isRunnable: () => true
```
Crash--, line 17: `isRunnable` seems to take the `konnector` as arg. I prefer to see it there, same for `match` and else.

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
doubleface: A mistake in rebase

packages/cozy-harvest-lib/src/konnector-policies.js
doubleface: This generates too many lines of logs in a simple connector execution. I now moved this removal in its own commit with explanations

packages/cozy-harvest-lib/src/konnector-policies.js
doubleface: Ok for clisk.

For the second part, I would prefer then to move the policies in their own directory named policies to make it clearer

packages/cozy-harvest-lib/src/konnector-policies.js
Crash--: Maybe it's a hint that there are too many call / re-renders ? Not sure that hiding it is the right solution. 

Is it intentional to have so many calls to this method?

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: I think it is clearer to separate account creation, trigger creation and trigger launch. And this makes it possible to policies to activated/deactivate any of them.

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
doubleface: These lines are not actually needed anymore

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
doubleface: Right, I put it back inside the same test now

packages/cozy-harvest-lib/src/services/cccPolicy.js
doubleface: Right, this is not actually needed. No reason to keep it

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -571,64 +501,91 @@ describe('ConnectionFlow', () => {
...
       client.query.mockRejectedValue(new Error('404'))
-      await flow.ensureTriggerAndLaunch(client, {
+      await setupSubmit(flow, {
         account: fixtures.existingAccount,
         trigger: fixtures.existingTrigger,
         konnector: fixtures.konnectorWithFolder
       })
       expect(flow.account).toEqual(fixtures.updatedAccount)
-      expect(saveAccount).not.toHaveBeenCalled()
+      expect(saveAccount).toHaveBeenCalledTimes(1)
```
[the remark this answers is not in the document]
doubleface: The change is that we call flow.handleFormSubmit in this test instead of ensureTriggerAndLaunch. flow.handleFormSubmit calls createOrUpdateAccount, which is why a call to saveAccount is expected.

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
doubleface: Done

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
doubleface: Done.

Concerning `ensureDefaultFolderPathInAccount`, it depends on saveAccount, which retries on 409 errors and it does not throw (I fixed the saveAccount outside the try catch), errors are converted to warning messages. 

In this case (no default folder path defined), it can be defined once again on next account  update or the stack has a mecanism of defining a default folder with the name as the slug of the connector.

packages/cozy-harvest-lib/src/konnector-policies.js
doubleface: Since this method can be called with different konnectors  in the same session, it looks like it's mandatory to call it everytime we need a policy.

At first I thought we could define a an association object on load and use it for the next calls, but since the user can install, update konnectors in the interval, we cannot be sure that the same policy will be used for the same konnector everytime.

doubleface:
@Crash-- A priori je suis à jour avec tes commentaires

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
```diff
@@ -57,7 +58,9 @@ export const LaunchTriggerAlert = ({
 
   const isInError = !!error
   const block = withDescription && (isInError || isInMaintenance)
-  const isKonnectorRunnable = isRunnable({ win: window, konnector })
+  const isKonnectorRunnable = findKonnectorPolicy(konnector).isRunnable({
```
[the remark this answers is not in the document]
Crash--: @doubleface is it wanted to still have the args?

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -585,16 +582,15 @@ export class ConnectionFlow {
     try {
       const result = await client.query(Q('io.cozy.files').getById(folderId))
       folder = result.data
+      if (folder.path !== account.defaultFolderPath) {
+        account.defaultFolderPath = folder.path
+        const savedAccount = await saveAccount(client, konnector, account)
+        return savedAccount
+      }
```
Crash--, line 589: can you explain why we do that inside the try / catch (aka why did you move the call)?

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -369,138 +368,92 @@ describe('ConnectionFlow', () => {
...
-  describe('ensureTriggerAndLaunch', () => {
-    beforeAll(() => {
-      jest.spyOn(cronHelpers, 'fromFrequency').mockReturnValue('0 0 0 * * 0')
-      watchKonnectorJob.mockReturnValue({ on: () => ({}) })
-    })
-
-    afterEach(() => {
-      jest.clearAllMocks()
+      delete window.cozy
+      delete window.ReactNativeWebView
```
Crash--, line 432: can you test that `launch()` is not called?

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -369,138 +368,92 @@ describe('ConnectionFlow', () => {
...
-        trigger: fixtures.existingTrigger,
-        konnector: fixtures.konnector
-      })
-      expect(launchTrigger).toHaveBeenCalledTimes(1)
-      expect(launchTrigger).toHaveBeenCalledWith(client, {
-        ...fixtures.existingTrigger
+      const promise = flow.handleFormSubmit({
+        konnector: fixtures.clientKonnector
       })
+      expect(async () => await promise).not.toThrow()
```
Crash--, line 446: but this time, should we test that `launch` is called?

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -571,66 +525,83 @@ describe('ConnectionFlow', () => {
...
+        flow.konnector,
+        onAccountCreationResult
       )
     })
-    delete window.cozy
-    delete window.ReactNativeWebView
   })
+})
 
+describe('ensureTrigger', () => {
```
Crash--, line 604: is it wanted?

packages/cozy-harvest-lib/src/services/cccPolicy.js
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+import logger from '../logger'
+
+/**
+ * Check if the given konnector is a client side connector
+ * @param {import('cozy-client/types/types').KonnectorsDoctype} konnector - konnector object
+ * @returns {Boolean}
+ */
+const isCCC = konnector => Boolean(konnector.clientSide)
```
Crash--: nit: should be inside the konnector model

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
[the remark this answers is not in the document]
doubleface: no. This is now fixed

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: Just a better practice. To avoid to have uncaught errors and saveAccount can be a source of errors.

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
Crash--: where?

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
Crash--: cc @doubleface

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
doubleface: If no launcher is available, the trigger is not launched

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
doubleface: no, artifact of previous tests. Removed

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
Crash--: I don't understand. If there is no launcher, then the policy will be the default one, right? And then, `needsTriggerLaunch` will be true, no?

doubleface: We are in case like this : in a web browser, in the home application, on a clisk.

- Since this is a web browser, there is no launcher available
- Since this is a clisk, the clisk policy is chosen

-> The trigger is not launched. Actually, the button to launch the konnector is not even enabled

I will even add a test to check that the trigger is not launched

Crash--: OK I got it, this is because of `konnector: fixtures.clientKonnector` 🤦

Crash-- (approved):

PR #2005, opened 2023-02-02
Fix computation of `LaunchTriggerAlert` styles

PolariTOON:


PolariTOON:
Closed in favor of #2007 which fixes more things.

PR #2007, opened 2023-02-02
Fix/launch trigger alert

Merkur39:
Quelques corrections suite au dernières avancées sur le nouveau bandeau `LaunchTriggerAlert`

### DIFF
--- a/packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
+++ b/packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
@@ -11,7 +11,7 @@ import Snackbar from 'cozy-ui/transpiled/react/Snackbar'
 import WrenchCircleIcon from 'cozy-ui/transpiled/react/Icons/WrenchCircle'
 import { makeStyles } from 'cozy-ui/transpiled/react/styles'
 
-import { getKonnectorSlug } from '../../helpers/triggers'
+import { getAccountId, getKonnectorSlug } from '../../helpers/triggers'
 import { isRunnable, isDisconnected } from '../../helpers/konnectors'
 import { useFlowState } from '../../models/withConnectionFlow'
 import { SUCCESS } from '../../models/flowEvents'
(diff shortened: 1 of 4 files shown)


packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
```diff
@@ -42,8 +42,7 @@ export const LaunchTriggerAlert = ({
   flow,
   t,
   konnectorRoot,
-  historyAction,
-  withDescription
```
PolariTOON, line 46: This prop is actually needed because we want to control whether the description is shown depending on the context. In particular, we want to see the description in Harvest dialog (where it is set), but not in the paper list in Mes Papiers (where it isn't set).

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
```diff
@@ -27,9 +27,11 @@ const useStyles = makeStyles({
   root: {
     padding: '.5rem 1rem'
   },
-  message: {
-    maxWidth: ({ block }) => block && 'calc(100% - 16px - .5rem)' // 16px is the size of the icon
-  },
+  message: ({ block }) =>
+    block && {
+      maxWidth: 'calc(100% - 16px - .5rem)', // 16px is the size of the icon
+      width: '100%'
```
PolariTOON, line 33: I don't think this change is needed as long as Cozy UI is up to date?
This was the point of the `flex: auto;` added in cozy/cozy-ui#2333.

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
```diff
@@ -52,10 +54,11 @@ export const LaunchTriggerAlert = ({
   const {
     data: { isInMaintenance, messages: maintenanceMessages }
   } = useMaintenanceStatus(client, konnector)
-  const styles = useStyles({ block })
 
   const isInError = !!error
   const block = isInError || isInMaintenance
+  const styles = useStyles({ block })
+
```
PolariTOON, line 61: This change is indeed needed though.

PolariTOON:
(merging as we need it for a build)

PR #2017, opened 2023-02-08
feat: Update label-slugs.js file

doubleface:
This avoids a degraded banking accounts reconciliation when account
institution label is unknown.

We should find a better way of reconciliation not dependent on
institutionLabels but this will probably avoid many unwanted accounts
merge at the moment.

The previous associations were also made to recognize old linxo institutionLabels.
but this is not in the scope anymore/


Crash-- (approved):

packages/cozy-doctypes/src/banking/slug-account.spec.js
```diff
@@ -1,11 +1,9 @@
 const { getSlugFromInstitutionLabel } = require('./slug-account')
 
 it('should find the slug from the institutionLabel', () => {
-  expect(getSlugFromInstitutionLabel("caisse d'Épargne (particuliers)")).toBe(
```
Crash--, line 4: Est-ce qu'en base chez nous on n'a pas du `caisse d'Épargne (particuliers)`?

packages/cozy-doctypes/src/banking/label-slugs.js
```diff
@@ -1,33 +1,105 @@
...
-  BNPP: 'bnpparibas82',
-  '/Boursorama.*/': 'boursorama83',
-  casden: 'casden173',
-  '/Hello bank!.*/': 'hellobank145',
-  Bred: 'bred',
-  CA: 'caatlantica3',
+  Bolden: 'bolden',
+  Boursorama: 'boursorama83',
+  BRED: 'bred',
+  "Caisse d'Épargne Particuliers": 'caissedepargne1',
```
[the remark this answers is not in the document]
doubleface: Since BI, we have only 3 institutionLabels associated to cic45 and they are listed here, for example. And I see that institutionLabel is updated when an account is updated

packages/cozy-doctypes/src/banking/slug-account.spec.js
doubleface: Il reste cette possibilité pour de très vieux comptes provenant le linxo et n'ayant pas été mis à jour depuis. Dans ce cas là, on passera alors par le réconciliation par score ce qui me semble acceptable

Crash-- (approved):

PR #2024, opened 2023-02-17
feat: New informations card

acezard:
## Screenshots
image
image


packages/cozy-harvest-lib/tsconfig.json
```diff
@@ -0,0 +1,35 @@
...
+    ],
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noFallthroughCasesInSwitch": true,
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "target": "es5"
+  }
+}
```
Crash--, line 35: should you remove it since there is cozy-tsconfig?

packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
```diff
@@ -34,7 +36,7 @@ const NewAccountModal = ({ konnector, onSuccess, onDismiss }) => {
   } = useMaintenanceStatus(client, konnector)
   const isMaintenanceLoaded =
     fetchStatus === 'loaded' || fetchStatus === 'failed'
-
+  const isLegacy = !(konnector.oauth || konnector.clientSide)
```
Crash--, line 39: this is not legacy, this is `serverSideKonnector`

packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.spec.jsx
```diff
@@ -1,28 +0,0 @@
...
-  it('should ignore invalid urls', () => {
-    const link = 'www.trainline.fr'
-    const root = render(
-      <I18n dictRequire={() => enLocale} lang="en">
-        <WebsiteLinkCard link={link} />
-      </I18n>
-    )
-    expect(root.queryByText('www.trainline.fr')).toBeFalsy()
-  })
-})
```
Crash--, line 28: Why did you remove those tests? There are still relevant. 

Also, I'm not a big fan of the renaming of `WebsiteLinkCard` to `InformationsCards`. In the future, there will a a lot of "InformationsCards" and those informations cards will contain `websitelinkcard` `supportlinkcard` and else.

packages/cozy-harvest-lib/tsconfig.json
acezard: Definitely

packages/cozy-harvest-lib/src/components/cards/WebsiteLinkCard.spec.jsx
acezard: Well, they were linked to a component that was deleted. And replaced by `InformationsCard` indeed. I didn't really see a meaningful way to test the component as it virtually contains no logic. But I should add back test the link fallback

acezard: About the other part of your comment, the InformationCard is intended to hold all future info displays like support etc. Is that ok for you? That info will just be props @Crash--

packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
acezard: Updated

Ldoppea (approved):

packages/cozy-harvest-lib/src/helpers/getErrorMessage.ts
```diff
@@ -0,0 +1,28 @@
+interface ErrorWithMessage {
+  message: string
+}
+
+const isErrorWithMessage = (error: unknown): error is ErrorWithMessage => {
```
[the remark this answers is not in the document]
acezard: Yeah, I find myself writing the same helpers many times. I'm not sure `cozy-client` is a good fit. Actually, I don't see which existing lib we could use

PR #2028, opened 2023-02-19
feat: Add cozy-tsconfig lib

acezard:
Currently copy-pasting this config a lot everywhere. Seems a lot better to just add it as a lib, that way we can control TS globally in our apps.

Please note that this config does NOT transpile files to javascript. It's only meant to be used as a declaration file generator, even for JS files. It's also used as a type check config in the IDE and/or during the CI process if opted in.

Please see the [readme](https://github.com/cozy/cozy-libs/blob/feat--Add-cozy-tsconfig-lib/packages/cozy-tsconfig/readme.md) file for more information

## Note:
Paths don't work at all when extending a tsconfig from node_modules it has been an issue for years. For now, I think the easiest workaround is to let people set the paths themselves on their projects.  It's the solution used by Microsoft themselves: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#tsconfig-bases

An alternative solution would be to provide this library as a script that would generate correct paths on installation (that still relies on the idea that everyone is using src/dist etc. which might not be the case anyway)

### DIFF
--- a/packages/cozy-tsconfig/license
+++ b/packages/cozy-tsconfig/license
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2023-present Cozy.io
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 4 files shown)


packages/cozy-tsconfig/license
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017-present, Cédric Patchane
```
Crash--, line 3: Can you change it please.

acezard: Didn't see it, changed 👍

Crash-- (approved):

zatteo (approved):

PR #2030, opened 2023-02-20
feat(harvest): Add create account wording in CannotConnectModal

JF-Cozy:


packages/cozy-harvest-lib/src/locales/en.json
```diff
@@ -23,7 +23,8 @@
     },
     "cannotConnectModal": {
       "title": "You can't connect?",
-      "content": "You can check your login or reset your password on the website "
+      "content": "You can check your login or reset your password on the website ",
+      "noAccount": "If you do not have an account you will need to create one on the website "
```
Crash--, line 27: vendorName should be passed as arg to the string because we should say: 

```
If you do not have an account you will need to create one on the {vendorName}'s website.
```

Plus, the `.` should not be part of the HTML.

Merkur39 (approved):

PR #2041, opened 2023-03-01
fix: Colors for shared withMe

Crash--:
Since the removal of the cozy-bar v7, Drive
don't load an old version of cozy-ui. These
variables are not defined anymore

see https://github.com/cozy/cozy-ui/pull/1743

The best way to use absolute color is to
import them from the palette.json
```
palette=json('settings/palette.json', { hash: true })
```
and use this way
```
background-color: palette['barney']
```

I needed to add the `:` because sometimes I had
compilations errors, sometime it was compiling in
cozy-sharing but the classname was not present
in the final bundle...

### DIFF
--- a/packages/cozy-sharing/src/components/badge.styl
+++ b/packages/cozy-sharing/src/components/badge.styl
@@ -1,4 +1,4 @@
-@require 'settings/palette.styl'
+palette=json('settings/palette.json', { hash: true })
 
 .shared-badge
     display flex
@@ -8,7 +8,8 @@
     background-clip padding-box
 
     &.--small
-        border .125rem solid var(--white)
+        border .125rem solid 
+        border-color: palette['white']
         width 1.5rem
         height 1.5rem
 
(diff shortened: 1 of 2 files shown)


packages/cozy-sharing/src/components/badge.styl
```diff
@@ -8,7 +8,8 @@
     background-clip padding-box
 
     &.--small
-        border .125rem solid var(--white)
+        border .125rem solid 
+        border-color: palette['white']
```
cballevre, line 12: In styl, you don't need `:` are they mandatory ? 🤔

cballevre (approved):

packages/cozy-sharing/src/components/badge.styl
Crash--: see the commit message 🙏

zatteo (approved):

packages/cozy-sharing/src/components/badge.styl
cballevre: Sry I just view the change file 😌 thx for the answer

Crash--: I saw that I didn't update the PR description, this is now done with the latest commit message.

PR #2046, opened 2023-03-06
Fix/duplicate owner in sharing

zatteo:
fix: Display correctly owner recipient when receiving sharing

Previously, we always displayed "You" as the owner. So when you received a sharing, you saw "You" two times : one for the owner, and one for you. Now, we correctly display the owner, whether it is you or not.

fix: Display always instance address as secondary text for owner

When we received a sharing, the owner was displayed as owner but with "invitation sent" status. Now, we consider that an owner is always "ready", so we always display the instance address instead of the status if it is an owner.

### DIFF
--- a/packages/cozy-sharing/src/components/Recipient/OwnerRecipient.jsx
+++ b/packages/cozy-sharing/src/components/Recipient/OwnerRecipient.jsx
@@ -1,0 +1,31 @@
+import React from 'react'
+
+import { useClient, useQuery, hasQueryBeenLoaded } from 'cozy-client'
+
+import Recipient from './Recipient'
+import { buildInstanceSettingsQuery } from '../../queries/queries'
+
+const OwnerRecipient = () => {
+  const client = useClient()
+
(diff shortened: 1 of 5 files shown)


packages/cozy-sharing/src/components/Recipient/OwnerRecipient.jsx
```diff
@@ -0,0 +1,48 @@
...
+const DefaultOwnerRecipient = () => {
+  const client = useClient()
+
+  const instanceSettingsQuery = buildInstanceSettingsQuery()
+  const instanceSettingsResult = useQuery(
+    instanceSettingsQuery.definition,
+    instanceSettingsQuery.options
+  )
+
+  const public_name = instanceSettingsResult?.data?.attributes?.public_name
```
cballevre, line 18: nit: use Camelcase for naming this variable even if propertie from attributes is in Snakecase to consistency

Crash-- (approved):

PR #2053, opened 2023-03-14
refactor: Use helper for get account name

Merkur39:
In order to be more resilient, it is better to use the `getAccountName` helper to retrieve the account name


Crash-- (approved):
<3

PR #2056, opened 2023-03-15
feat(Harvest): Back to idle state when startLauncher resolves

doubleface:
When a clisk konnector is launched, the flagship app can send back a message to allow the conection flow to go back to IDLE state

See https://github.com/cozy/cozy-flagship-app/pull/659 for more context

from Screenshot_1678867755

to Screenshot_1678867818

- refactor: Convert ccc -> clisk
- fix: Import order lint via --fix lint option


packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -627,11 +627,16 @@ export class ConnectionFlow {
...
+      konnectorPolicy
+        .onLaunch({
+          konnector: this.konnector,
+          trigger: this.trigger,
+          account: this.account
+        })
+        .then(() => this.setState({ status: IDLE }))
+        .catch(err =>
+          logger.error(`Error while launching policy : ${err.message}`)
+        )
```
Crash--, line 641: Is it on purpose to not block the process with an `await`? Because even if an error is thrown, we will subscribe to the realtime and so on.  If this is done for purpose, than can you explain why in the commit?

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
```diff
@@ -1,4 +1,8 @@
 import EventEmitter from 'events'
+import { waitFor } from '@testing-library/react'
+
+import { Q } from 'cozy-client'
```
Crash--, line 5: how previous version of this file can have been worked? :o Are we ignoring somehow the result of the tests?

packages/cozy-harvest-lib/src/policies/clisk.js
```diff
@@ -53,7 +94,21 @@ function onLaunch({ konnector, account, trigger }) {
...
+ * @param {Function} callback - callback function
+ */
+function addMessageListener(callback) {
+  window.document?.addEventListener?.('message', function (e) {
+    callback(e.data)
+  })
+  window?.addEventListener?.('message', function (e) {
+    callback(e.data)
+  })
 }
```
Crash--, line 112: not blocking for now, but can't we start to use cozy-intent instead of calling postMessage / listener manually? cc @Ldoppea

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: This was on purpose. I added explanations about it in the commit

packages/cozy-harvest-lib/src/models/ConnectionFlow.spec.js
doubleface: This is rebase error after eslint changed the order of imports. I fix it

packages/cozy-harvest-lib/src/policies/clisk.js
Ldoppea: We discussed about that yesterday.

If we want to use cozy-intent here, then we will have to rethink the flagship app architecture:
- current postMessage are handled by the HomeView (cf [this code](https://github.com/cozy/cozy-flagship-app/blob/master/src/screens/home/components/HomeView.js#L216-L217)) and then we set the Launcher state using a drilled prop
- if we use cozy-intent, then the message won't be intercepted by HomeView anymore but by the `localMethods` (cf [this code](https://github.com/cozy/cozy-flagship-app/blob/master/src/App.js#L177)). Then we should find a way to interact with the Launcher from `localMethods` (this can be done using the same LauncherEvent item we implemented in [this PR](https://github.com/cozy/cozy-flagship-app/pull/659))

So if we want to use cozy-intent we may want to rework this part, and we may want to extract the LauncherView from the HomeScreen component (as  the drilled `setLauncherContext` prop is the only reason for having them in the same component tree)

One last thing, we heard that MesPapiers may need to launch konnectors in the future, if this is the case, there is one more reason to do this rework because today only the HomeView can do this. But we will also have to think about how we should handle app permissions for this.

Crash--: > So if we want to use cozy-intent we may want to rework this part, and we may want to extract the LauncherView from the HomeScreen component

please yep 🙏 
https://github.com/cozy/cozy-flagship-app/pull/534#discussion_r1066026603

packages/cozy-harvest-lib/src/policies/clisk.js
```diff
@@ -53,12 +94,26 @@ function onLaunch({ konnector, account, trigger }) {
...
+  })
+}
+
+/**
+ * Listens to postMessage event. Compatible with android and ios webviews
+ *
+ * @param {Function} callback - callback function
+ */
+function addMessageListener(callback) {
+  window.document?.addEventListener?.('message', function (e) {
```
Ldoppea, line 106: nit

```suggestion
  // Android's WebView messaging interface
  window.document?.addEventListener?.('message', function (e) {
```

packages/cozy-harvest-lib/src/policies/clisk.js
```diff
@@ -53,12 +94,26 @@ function onLaunch({ konnector, account, trigger }) {
...
+    callback(e.data)
+  })
+  window?.addEventListener?.('message', function (e) {
```
Ldoppea, line 109: nit
```suggestion
  // iOS's WebView messaging interface
  window?.addEventListener?.('message', function (e) {
```

Ldoppea (approved):

PR #2058, opened 2023-03-15
feat: Return also readOnly status from getSharedDocument

cballevre:
The `getSharedDocument` function is only used in cozy-drive at the moment.  Cozy-notes has its own version that comes close to these changes. It may be possible to use the same for both applications in the future.

BREAKING CHANGE: The function now returns the id encapsulated in an object. You can extract the id like this `const { id } = getSharedDocument(client)`

### DIFF
--- a/packages/cozy-sharing/src/getSharedDocument.js
+++ b/packages/cozy-sharing/src/getSharedDocument.js
@@ -1,15 +1,23 @@
-import get from 'lodash/get'
+import { isReadOnly } from 'cozy-client/dist/models/permission'
 
+/**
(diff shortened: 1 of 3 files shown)


Crash--:
@cballevre you need to update the tests according to your BC

Crash--:
(also it seems to have console.log within a few tests of cozy-sharings, can you remove them? (https://app.travis-ci.com/github/cozy/cozy-libs/jobs/598132136#L5471) )

cballevre:
@Crash-- can I have a new review pls ? 🙏

Crash-- (approved):

PR #2063, opened 2023-03-17
fix: Document methods work with both clients

taratatach:
The `Document` methods make different internal calls depending on the
client being used (i.e. `cozy-client` or `cozy-client-js`).
However, the `cozy-client` calls' responses don't have the same
signature as the `cozy-client-js` ones and were not transformed
leading to errors in (at least) the `BankingReconciliator` when trying
to match transactions with their account.

We now make sure that the `cozy-client` calls returning an object with
a `data` attribute are transformed so that the `Document` methods
always return the content of that attribute (`cozy-client-js` already
does that).


doubleface (approved):
Thanks for the fix @taratatach

PR #2072, opened 2023-03-23
fix(harvest): Display backdrop in a portal

acezard:
- Also delete other Backdrop renders
- This should fix all issues regarding backdrop being under harvest part

Simulator Screen Shot - iPhone 13 - 2023-03-23 at 10 17 07


Crash-- (approved):
looks great to me, waiting for @doubleface review

PR #2080, opened 2023-03-30
fix(harvest): Show account name in header for CCC

acezard:
It was hidden, now we fetch the account name
with the Account model from cozy-doctypes.
It is tied to `showAccountSelection` prop, not to clientSide.
Must be wary of that in the future if changing something in this scope

image

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountModal.jsx
@@ -4,10 +4,12 @@ import PropTypes from 'prop-types'
 import React, { Component } from 'react'
 
 import { withClient } from 'cozy-client'
+import { Account } from 'cozy-doctypes'
 import Button from 'cozy-ui/transpiled/react/Button'
 import DialogContent from 'cozy-ui/transpiled/react/DialogContent'
 import Infos from 'cozy-ui/transpiled/react/Infos'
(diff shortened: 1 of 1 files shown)


packages/cozy-harvest-lib/src/components/AccountModal.jsx
```diff
@@ -127,7 +129,7 @@ export class AccountModal extends Component {
     return (
       <>
         <KonnectorModalHeader konnector={konnector}>
-          {showAccountSelection && (
+          {showAccountSelection ? (
```
acezard, line 132: at the moment, this props is only false when running a clientSide konnector

packages/cozy-harvest-lib/src/components/AccountModal.jsx
```diff
@@ -4,10 +4,12 @@ import PropTypes from 'prop-types'
 import React, { Component } from 'react'
 
 import { withClient } from 'cozy-client'
+import { Account } from 'cozy-doctypes'
```
Crash--, line 7: cozy-doctypes is deprecated, you should not use it for new code.

You should also check in the cozy-client's models https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/account.js#L94

Crash-- (approved):
Approving since this I just want a replacement of the import / lib used, but I except the change before merging 🙏

PR #2087, opened 2023-03-30
fix: Detect also Flagship app with isMobileApp

cballevre:
For me there is a mismatch between the name `isMobileApp` and what it actually does. I therefore propose to also check if it is a flagship app in this function. To check cordova we can do it with `isCordova`

BREAKING CHANGE: Now `isMobileApp` also checks if it is on the Flagship app. If you only want to check if the application you are running is based on Cordova, you can use `isCordova`


Crash--:
I agree in broad strokes at first approach. But let me think about it for 2 seconds to see if it really makes sense

cballevre:
This PR is no longer relevant as we no longer have a Cordova application in use.

PR #2089, opened 2023-03-30
fix: Set connection id of current bank's accounts

taratatach:
When (re-)connecting a bank, we try to match existing accounts with
newly fetched ones.
When we find matches, we update the existing accounts with the fresh
data and set the connection id with the new one.
We also update the connection id of other bank accounts that were
sharing the same old connection id with the matching bank account.

e.g. :

  Say we have
  - bank account `#1` with connection id A
  - bank account `#2` with connection id A
  - bank account `#3` with connection id B

  And we fetch
  - bank account `#1` connection id C

  Then we'll make these updates
  - bank account `#1` gets connection id C
  - bank account `#2` gets connection id C

Now, if an existing bank account matches with a fetched one but did
not have a connection id anymore, we'll use the new connection id as
well.

However, the way we were doing so resulted in the update of all bank
accounts without connection ids whenever one of them would match a
newly fetched one.
This means that bank accounts from different banks could get
associated with the same connection.

To avoid this, we now prevent connection id updates for bank accounts
missing one unless they have a match in the newly fetched bank
accounts.

### DIFF
--- a/packages/cozy-doctypes/src/banking/BankAccount.js
+++ b/packages/cozy-doctypes/src/banking/BankAccount.js
@@ -9,6 +9,8 @@ const matching = require('./matching-accounts')
 const { getSlugFromInstitutionLabel } = require('./slug-account')
 const Document = require('../Document')
 
+const connectionId = account => account?.relationships?.connection?.data?._id
+
 class BankAccount extends Document {
   /**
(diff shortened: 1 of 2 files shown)


doubleface (approved):
Thanks @taratatach and nice explanation of the problem :+1:

PR #2091, opened 2023-03-31
Feat: Display backdrop only when user is needed

doubleface:
When the konnector is a clisk konnector and before LOGIN_SUCCESS
And on first run (account creation or identifiers update or action needed error state) for other types of konnectors and before LOGIN_SUCCESS

Add the possibility to get LOGIN_SUCCESS event from a launcher event.

- fix: Do not fail on loginSuccess when no job watcher is defined
- fix: Only display backdrop before login success
- fix: Misleading log
- fix: Do not redirect to account if no account
- feat: Get LOGIN_SUCCESS event from flagship app
- feat: Do not wait for LOGIN_SUCCESS event
- fix: Import order
- feat: Upgrade cozy-client harvest dev dependency
- fix: Lint
- fix: Protection in clisk policy
- feat: Backdrop display rules change
- feat: Display backdrop if the error needs an action from the user
- test: Add some unit test on needUser ConnectionFlow state

### DIFF
--- a/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
+++ b/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
@@ -190,7 +190,9 @@ export class ConnectionFlow {
     this.unwatch = this.unwatch.bind(this)
 
     this.state = {
-      status: IDLE
+      status: IDLE,
+      accountError: null,
+      firstRun: false
     }
 
     // Stores data necessary for custom connection flow, for example
@@ -469,7 +471,10 @@ export class ConnectionFlow {
 
       const client = this.client
 
-      this.setState({ status: CREATING_ACCOUNT })
+      this.setState({
+        status: CREATING_ACCOUNT,
+        firstRun: true // when the user submits new authentication information, this is considered a first run
+      })
       this.trigger = trigger
       this.account = account
       this.konnector = konnector
@@ -872,16 +877,22 @@ export class ConnectionFlow {
 
   getDerivedState() {
     const trigger = this.trigger
-    const { status, accountError } = this.state
+    const { status, accountError, firstRun } = this.state
     const triggerError = triggersModel.getKonnectorJobError(trigger)
     const running =
       trigger?.current_state?.status === 'running' ||
       ![ERRORED, IDLE, SUCCESS].includes(status)
+    const konnectorPolicy = this.getKonnectorPolicy()
     const expectingTriggerLaunch = status === EXPECTING_TRIGGER_LAUNCH
     const error =
       !running &&
       !expectingTriggerLaunch &&
       (this.getMockError() || accountError || triggerError)
+    const userNeeded =
+      running &&
+      !this.state[LOGIN_SUCCESS_EVENT] &&
+      (konnectorPolicy.name === 'clisk' || firstRun)
+
     return {
       running,
       twoFARunning: status === RUNNING_TWOFA,
@@ -891,7 +902,8 @@ export class ConnectionFlow {
       accountError,
       error,
       konnectorRunning: triggersModel.isKonnectorRunning(trigger),
-      expectingTriggerLaunch
+      expectingTriggerLaunch,
+      userNeeded
     }
   }
 
--- a/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
+++ b/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
@@ -654,7 +654,8 @@ export class ConnectionFlow {
         .onLaunch({
           konnector: this.konnector,
           trigger: this.trigger,
-          account: this.account
+          account: this.account,
+          flow: this
         })
         .then(() => this.setState({ status: IDLE }))
         .catch(err =>
--- a/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
+++ b/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
@@ -700,9 +700,29 @@ export class ConnectionFlow {
   }
 
   /**
+   * Tell if the error is an error which needs an action from the user
+   *
+   * @param {Error} error - error object
+   * @returns {boolean}
+   */
+  isUserActionError(error) {
+    if (!error) return false
+
+    const userErrors = ['USER_ACTION_NEEDED', 'LOGIN_FAILED']
+    return userErrors.some(userError => error.message.includes(userError))
+  }
+
+  /**
    * Launches the job and sets everything up to follow execution.
    */
   async launch({ autoSuccessTimer = true } = {}) {
+    const { error } = this.getState()
+
+    if (this.isUserActionError(error)) {
+      // accounts with user actionnable errors are considered as the first run.
+      // this will especially allow to show the backdrop effect to tell the user to stay until the login is successful
+      this.setState({ firstRun: true })
+    }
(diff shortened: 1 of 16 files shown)


packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -395,7 +395,7 @@ export class ConnectionFlow {
   }
 
   handleLoginSuccess() {
-    this.jobWatcher.handleLoginSuccess()
+    this.jobWatcher?.handleLoginSuccess()
```
Crash--: how is it possible? How can we want to handle login success without having a jobWatcher?

packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
```diff
@@ -48,6 +48,9 @@ const NewAccountModal = ({ konnector, onSuccess, onDismiss }) => {
     }
 
     const accountId = triggersModel.getAccountId(trigger)
+    if (!accountId) {
```
Crash--, line 51: We're in a `handleSuccess` method of the `NewAccountModal`. How can we call it without having an `account`?

packages/cozy-harvest-lib/src/policies/clisk.js
```diff
@@ -11,7 +11,7 @@ import { LOGIN_SUCCESS_EVENT } from '../models/flowEvents'
  * @param {import('cozy-client/types/types').KonnectorsDoctype} konnector - konnector object
  * @returns {Boolean}
  */
-const isClisk = konnector => Boolean(konnector.clientSide)
+const isClisk = konnector => Boolean(konnector?.clientSide)
```
Crash--, line 14: Why this method can be called without having a konnector? We should fix the caller, not this method, no? Also, type & jsdoc have to been updated

packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
doubleface: It can happen with clisk konnectors when the flagship application navigates to the account before the realtime event with the account creation.
That is why the flagship application passes the account id with the loginSuccess message. And I forgot to use it. This is now fixed.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: This is a mistake, it cannot happen

packages/cozy-harvest-lib/src/policies/clisk.js
doubleface: Yes, I fixed the call (in the FlowProvider) to test if there is a konnector

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -385,17 +435,21 @@ export class ConnectionFlow {
...
       // eslint-disable-next-line no-console
+      // @ts-ignore
       logger.error(error)
       this.setState({ status: ERRORED, error })
     }
   }
 
   handleLoginSuccess() {
-    this.jobWatcher.handleLoginSuccess()
+    if (this.jobWatcher?.handleLoginSuccess) {
```
Crash--, line 450: why this test?

doubleface: I did the comment before the push of the code sorry. This test is now removed (I juste needed to fix some unit tests)

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
Crash--: commit to remove?

packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
Crash--: Commit to remove?

yarn.lock
```diff
@@ -18621,6 +18727,13 @@ regenerator-transform@^0.15.0:
...
+regenerator-transform@^0.15.1:
+  version "0.15.1"
+  resolved "https://registry.yarnpkg.com/regenerator-transform/-/regenerator-transform-0.15.1.tgz#f6c4e99fc1b4591f780db2586328e4d9a9d8dc56"
+  integrity sha512-knzmNAcuyxV+gQCufkYcvOqX/qIIfHLv0u5x79kRxuGojfYVky1f15TzZEu2Avte8QGepvUNTnLskf8E6X6Vyg==
+  dependencies:
+    "@babel/runtime" "^7.8.4"
+
 regex-cache@^0.4.2:
   version "0.4.4"
   resolved "https://registry.yarnpkg.com/regex-cache/-/regex-cache-0.4.4.tgz#75bdc58a2a1496cec48a12835bc54c8d562336dd"
```
Crash--, line 18742: Did you have conflict in the lock file? All these modification for babel stuff do not inspire me.

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
```diff
@@ -904,7 +902,7 @@ export class ConnectionFlow {
     const running =
       trigger?.current_state?.status === 'running' ||
       ![ERRORED, IDLE, SUCCESS].includes(status)
-    const konnectorPolicy = this.getKonnectorPolicy()
+    const konnectorPolicy = this.konnector ? this.getKonnectorPolicy() : null
```
Crash--, line 905: We have everywhere in this file call to `this.konnector`. How is it possible to be undefined? 

Does it mean that https://github.com/cozy/cozy-libs/blob/eed38566c7fe2eed1d4e83083d0a97538d27eb23/packages/cozy-harvest-lib/src/models/ConnectionFlow.js#L176 can be undefined here? Then the jsdoc is not right. 

But is is really intended to have a connectionFlow without a konnector? 🤔

doubleface: No, I added a new commit to make the konnector mandatory to FlowProvider/ConnectionFlow. I had to change some components having the FlowProvider as dependency/subdependency and some unit tests. I'm wondering if this should be in another PR

yarn.lock
doubleface: No, no conflict, just an upgrade of cozy-client

doubleface:
@Crash-- PR updated

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
Crash--: @doubleface should we remove this commit?

packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
Crash--: Same here...

Crash-- (approved):
Approving but please rework the first commits 🙏

packages/cozy-harvest-lib/src/models/ConnectionFlow.js
doubleface: done

packages/cozy-harvest-lib/src/components/NewAccountModal.jsx
doubleface: done

Crash-- (approved):

PR #2097, opened 2023-04-03
Fix(harvest): handle navback from edit dialog

acezard:
Previously the history was mutated so we couldn't go back. Now it's correctly updated so we can go back


Crash-- (changes requested):
router v4 / v6

Crash-- (approved):

PR #2100, opened 2023-04-05
fix: Use io.cozy.settings.flags and io.cozy.settings.context id

zatteo:
Instead of just flags and contexts to avoid store issues.

See cozy-client 3a35f4fd34a74b1133bb9fc36a1f915c69e4122f

Hidden goal : chasing and removing deprecated stuff flooding flagship app logs

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/konnectors.js
+++ b/packages/cozy-harvest-lib/src/helpers/konnectors.js
@@ -174,7 +174,7 @@ export const fetchSupportMail = async client => {
   const result = await client.fetchQueryAndGetFromState({
     definition: Q('io.cozy.settings').getById('io.cozy.settings.context'),
     options: {
-      as: 'contextSupportMail',
+      as: 'io.cozy.settings/io.cozy.settings.context',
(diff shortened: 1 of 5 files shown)


packages/cozy-harvest-lib/src/helpers/konnectors.js
```diff
@@ -172,7 +172,7 @@ export const getErrorLocale = (
 
 export const fetchSupportMail = async client => {
   const result = await client.fetchQueryAndGetFromState({
-    definition: Q('io.cozy.settings').getById('context'),
+    definition: Q('io.cozy.settings').getById('io.cozy.settings.context'),
     options: {
       as: 'contextSupportMail',
```
Crash--, line 177: This is not the right `as`. Should be `io.cozy.settings/io.cozy.settings.context` (without the right as, cache is not working correctly since the dedup is done on the as)

zatteo: Right I did not even read it. Let's fix it in the same time.

PR #2102, opened 2023-04-06
fix(harvest): Check for file infos before fetch

acezard:
We had a bug because `useDataCardFiles`
was called with `(undefined, undefined)`
This bug happened because `useParams` was used after `useDataCardFiles`

But it still returned some fallback values due to query cache.
So we always defaulted to some files the useQuery was able to get.

Now we will throw if the hook is called without arguments.
This case wasn't supposed to exist,
as such it should be treated as an error

It would be difficult to know what to do with empty arguments.
Are they coming later? never?

On the other hand I'm not sure at this moment if
there are scenarios where that error can happen.
We have no elegant error handler for it as of this commit.

### DIFF
--- a/packages/cozy-harvest-lib/src/datacards/ViewerModal.jsx
+++ b/packages/cozy-harvest-lib/src/datacards/ViewerModal.jsx
@@ -8,10 +8,9 @@ import { useDataCardFiles } from './useDataCardFiles'
 import { MountPointContext } from '../components/MountPointContext'
 
 export const ViewerModal = () => {
+  const { accountId, folderToSaveId, fileIndex } = useParams()
   const { pushHistory, replaceHistory } = useContext(MountPointContext)
   const { data, fetchStatus } = useDataCardFiles(accountId, folderToSaveId)
-  const { accountId, folderToSaveId, fileIndex } = useParams()
-
   const handleCloseViewer = () => replaceHistory(`/accounts/${accountId}`)
   const handleFileChange = (_file, newIndex) =>
     pushHistory(`/viewer/${accountId}/${folderToSaveId}/${newIndex}`)
--- a/packages/cozy-harvest-lib/src/datacards/useDataCardFiles.js
+++ b/packages/cozy-harvest-lib/src/datacards/useDataCardFiles.js
@@ -74,6 +74,9 @@ const getResponse = (folderToSaveFiles, sourceAccountFiles) => {
 }
 
 export const useDataCardFiles = (accountId, folderToSaveId) => {
+  if (!accountId || !folderToSaveId)
+    throw new Error('Missing arguments in useDataCardFiles, cannot fetch files')
+
   const folderToSaveFiles = useFolderToSaveFiles(folderToSaveId)
   const sourceAccountFiles = useSourceAccountFiles(accountId)
 


packages/cozy-harvest-lib/src/datacards/ViewerModal.jsx
```diff
@@ -8,10 +8,9 @@ import { useDataCardFiles } from './useDataCardFiles'
 import { MountPointContext } from '../components/MountPointContext'
 
 export const ViewerModal = () => {
+  const { accountId, folderToSaveId, fileIndex } = useParams()
   const { pushHistory, replaceHistory } = useContext(MountPointContext)
   const { data, fetchStatus } = useDataCardFiles(accountId, folderToSaveId)
```
Crash--, line 13: how our linter didn't tell us about this error?

Crash--:
> But it still returned some fallback values due to query cache.

This is not 100% true. Since `folderTosaveId` was undefined here : 
https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/datacards/useDataCardFiles.js#L8-L23

The `dir_id: undefined` was removed from the request. So the final request was: 
```
 Q('io.cozy.files')
      .where({
        trashed: false
      })
      .indexFields(['dir_id', 'cozyMetadata.createdAt'])
      .sortBy([{ dir_id: 'desc' }, { 'cozyMetadata.createdAt': 'desc' }])
      .limitBy(5),
    {
      as: `fileDataCard_io.cozy.files/${folderToSaveId}/io.cozy.files`,
      fetchPolicy: CozyClient.fetchPolicies.olderThan(30 * 1000)
    }
```

packages/cozy-harvest-lib/src/datacards/ViewerModal.jsx
acezard: I'm not sure the linter could have helped here (can't think of a rule). But TypeScript would have detected that the parameters could be undefined

acezard:
> > But it still returned some fallback values due to query cache.
> 
> This is not 100% true. Since `folderTosaveId` was undefined here : https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/datacards/useDataCardFiles.js#L8-L23
> 
> The `dir_id: undefined` was removed from the request. So the final request was:
> 
> ```
>  Q('io.cozy.files')
>       .where({
>         trashed: false
>       })
>       .indexFields(['dir_id', 'cozyMetadata.createdAt'])
>       .sortBy([{ dir_id: 'desc' }, { 'cozyMetadata.createdAt': 'desc' }])
>       .limitBy(5),
>     {
>       as: `fileDataCard_io.cozy.files/${folderToSaveId}/io.cozy.files`,
>       fetchPolicy: CozyClient.fetchPolicies.olderThan(30 * 1000)
>     }
> ```

Yes, that's a pretty lousy query without the folder ID, so I choose to throw it instead of waiting, for instance. Do you agree with that choice?

packages/cozy-harvest-lib/src/datacards/ViewerModal.jsx
Crash--: accountId was not defined yet. So linter should have been able to tell us that we were reading a non created variable yet.

acezard: Yes, that's correct, maybe the linter was confused by react-specific rules and thought it was hoisted or something

PR #2104, opened 2023-04-06
feat(harvest): Update config page header

acezard:
- Go Back
- Only show the account name
- Different styles than the default modal

This was painful to implement because I could not use the CozyDialogs API (used in Figma). So I had to create a sort of custom style for that page. Thankfully didn't need to create any CSS so I think it's ok, but ultimately I think Harvest should be refactored to use CozyDialogs everywhere, so this implementation is not necessary.

Before:
image

After:
image

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountModalWithoutTabs/AccountModalHeader.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountModalWithoutTabs/AccountModalHeader.jsx
@@ -1,7 +1,9 @@
 import PropTypes from 'prop-types'
 import React from 'react'
+import { useMatch } from 'react-router-dom'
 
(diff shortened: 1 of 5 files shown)


packages/cozy-harvest-lib/src/components/AccountModalWithoutTabs/AccountModalHeader.jsx
```diff
@@ -1,7 +1,9 @@
 import PropTypes from 'prop-types'
 import React from 'react'
+import { useMatch } from 'react-router-dom'
```
acezard, line 3: There is another Header component for V4 (`packages/cozy-harvest-lib/src/components/AccountModalWithoutTabs/ForV4Router/AccountModalWithoutTabs.jsx`)

Crash-- (approved):

PR #2106, opened 2023-04-11
fix: Regression in useKonnectorWithTrigger

doubleface:
{data: []} structure was not used everywhere.
This caused realtime error messages and some trigger values where not
updated


Crash--:
ping @doubleface

doubleface:
@Crash-- Now we have unit tests thanks to @acezard

Crash-- (approved):

PR #2112, opened 2023-04-12
fix: Refetch current trigger when it is running

doubleface:
When creating a new account/trigger couple, when the given trigger has
started running, the FlowProvider is reloaded, which means it
unsubscribes from realtime events until it is reloaded.

If the konnector execution is really fast (example: dummy-> 1s), its
execution could end during the FlowProvider reload (which is a little
less than 1s on my development environment).

To fix this issue, I force the ConnectionFlow to refetch the current
trigger if this trigger is in the state "running", in case we missed the
konnector execution end.

### DIFF
--- a/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
+++ b/packages/cozy-harvest-lib/src/models/ConnectionFlow.js
@@ -799,8 +799,10 @@ export class ConnectionFlow {
    */
   watchTriggerJobs() {
     if (this.trigger) {
-      // When the trigger comes from realtime, cozy-stack does not add the current_state to the object. So we need to request the stack to get it
-      if (!this.trigger?.current_state) {
+      if (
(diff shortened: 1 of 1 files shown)


Crash--:
> when the given trigger has started running, the FlowProvider is reloaded

You mean unmounted and then mounted again, right? Why do we do that?

doubleface:
@Crash-- 

> You mean unmounted and then mounted again, right? Why do we do that?

With the url change from `#/connected/dummy/new` to `#/connected/dummy/accounts/<accountid>` a new Component is mounted. Both components have their own FlowProvider

Crash--:
Quick note:
> Temps réel sur les triggers ne met pas à jour current_state donc on reste avec trigger.current_state en running quand on arrive sur la page #/connected/dummy/accounts/<accountid>
> Vu qu'on écoute les jobs plus haut, on pourrait certainement changer en mémoire le current_state sur le trigger.

Crash-- (approved):

PR #2136, opened 2023-04-28
fix: Upgrade lerna from 4.0.0 to 5.0.0

taratatach:
v4.0.0 ships with an old version of `node-gyp` which is incompatible
  with recent versions of python (v3.11+) and thus prevents installing
  `parcel` because its `deasync` dependency needs to be compiled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "css-loader": "2.1.1",
     "git-directory-deploy": "1.5.1",
     "husky": "1.3.1",
-    "lerna": "4.0.0",
+    "lerna": "5.0.0",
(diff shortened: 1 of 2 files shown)


zatteo (approved):

package.json
```diff
@@ -31,7 +31,7 @@
     "css-loader": "2.1.1",
     "git-directory-deploy": "1.5.1",
     "husky": "1.3.1",
-    "lerna": "4.0.0",
+    "lerna": "5.0.0",
```
Crash--, line 34: why 5.0.0 and not the latest 5.X.X?

PR #2142, opened 2023-05-10
fix: Do not overwrite onDeleteAccount in KonnectorAccountWrapper

doubleface:
This property is already defined in components using this componenent
I checked other overwritten properties and it looks OK.

This fixes an error when trying to delete an account with `harvest.inappconnectors.enabled` flag enabled


Crash--:
Can we add test or something somewhere? We have too much regression 🙏

PR #2143, opened 2023-05-10
fix(mespapiers): Harvest banner wasn't shown for disconnected account

JF-Cozy:
When a konnector account is disconnected, the account object in
couch is removed. We want to show the banner even in this
situation. The banner will be in "disconnected" status in this case


Crash--:
Can you add a test please? We have a lot of regressions since the introduction of the new flag. And we're fixing them slowly. To prevent this situation, can we be more robust with tests? 🙏

Merkur39 (approved):

PR #2148, opened 2023-05-12
Add scanDocument type

zatteo:


packages/cozy-intent/src/api/models/methods.ts
```diff
@@ -21,6 +21,7 @@ interface _NativeMethodsRegister {
   ) => Promise<boolean | null>
   isBiometryDenied: () => Promise<boolean>
   isNativePassInstalledOnDevice: () => Promise<boolean>
+  scanDocument: () => Promise<string | undefined>
```
Crash--, line 24: nit: maybe we can be more precise than just `string` https://stackoverflow.com/questions/66930931/is-there-a-type-for-base64-encoded-images-in-typescript

PR #2149, opened 2023-05-13
feat(cozy-scanner): Hide health theme

Merkur39:
On the condition of a flag (`hide.healthTheme.enabled`), we wish to be able to hide the Health theme.

### DIFF
--- a/packages/cozy-scanner/src/DocumentQualification.jsx
+++ b/packages/cozy-scanner/src/DocumentQualification.jsx
@@ -1,25 +1,23 @@
-import React, { Component } from 'react'
-import PropTypes from 'prop-types'
 import classNames from 'classnames'
+import PropTypes from 'prop-types'
+import React, { Component } from 'react'
 
+import flag from 'cozy-flags'
+import { translate } from 'cozy-ui/transpiled/react/I18n'
 import Icon from 'cozy-ui/transpiled/react/Icon'
-import Label from 'cozy-ui/transpiled/react/Label'
+import QualifyIcon from 'cozy-ui/transpiled/react/Icons/Qualify'
 import Input from 'cozy-ui/transpiled/react/Input'
 import InputGroup from 'cozy-ui/transpiled/react/InputGroup'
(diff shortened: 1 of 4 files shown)


packages/cozy-scanner/src/DocumentQualification.jsx
```diff
@@ -42,11 +40,19 @@ export class DocumentQualification extends Component {
...
       selected: { categoryLabel, item },
       filename: this.defaultFilename,
-      hasUserWrittenFileName: false
+      hasUserWrittenFileName: false,
+      overridedThemes: null
     }
     this.textInput = React.createRef()
   }
 
+  componentDidMount() {
```
Crash--, line 49: can you add the UNSAFE_ prefix please? Also, why not during the `willMount` ?

packages/cozy-scanner/src/DocumentQualification.spec.jsx
```diff
@@ -79,4 +94,10 @@ describe('DocumentQualification', () => {
     fireEvent.click(queryByText('Scan.items.birth_certificate'))
     expect(inputFileName.value).toBe('Manual name')
   })
+
+  it('Hide health theme', () => {
+    const { queryByText } = setup({ isFlag: true })
```
Crash--, line 99: nit: `isFlag` is not speaking for itself, but let's keep it that way since this component will be removed pretty soon.

packages/cozy-scanner/src/DocumentQualification.jsx
Merkur39: Indeed, the `UNSAFE_componentWillMount` method is more suitable.

PS: The `componentDid...` and `componentWillUnmount` methods do not have a prefix.

Crash--: > PS: The componentDid... and componentWillUnmount methods do not have a prefix.

Oh. To old to remember that stuff ^^'

Crash-- (approved):

PR #2152, opened 2023-05-15
Add `other_administrative_document` paper

JF-Cozy:


packages/cozy-mespapiers-lib/src/helpers/queries.js
```diff
@@ -50,7 +51,8 @@ export const buildFilesQueryWithQualificationLabel = () => {
         })
         .partialIndex({
           'metadata.qualification.label': {
-            $exists: true
+            $exists: true,
+            ...(flag('hide.healthTheme.enabled') && { $nin: healthLabels })
```
paultranvan, line 55: I think this is dangerous: your partial index MUST be static, i.e. the values should not change. Otherwise, you will create a new index each time your `healthLabels`change.

packages/cozy-mespapiers-lib/src/helpers/queries.js
```diff
@@ -68,7 +70,10 @@ export const buildFilesQueryByLabel = label => ({
     Q(FILES_DOCTYPE)
       .where({
         'metadata.qualification': {
-          label: label
+          label: {
+            $eq: label,
+            ...(flag('hide.healthTheme.enabled') && { $nin: healthLabels })
+          }
```
paultranvan, line 76: The index should be on `metadata.qualification.label` rather than `metadata.qualification`, don't you have a warning on this query?
Also, keep in mind the `$nin` operator will force a full-scan of the whole index (which is somewhat acceptable as long as there are not too many qualified files).

packages/cozy-mespapiers-lib/src/helpers/queries.js
JF-Cozy: nouvelle approche : on ne touche plus aux requêtes

packages/cozy-mespapiers-lib/src/helpers/queries.js
JF-Cozy: nouvelle approche : on ne touche plus aux requêtes

PR #2156, opened 2023-05-16
feat: Allow using non-native `WebSocket`

taratatach:
Node core does not offer a `WebSocket` implementation. Node
applications have to use a third-party package, most likely `ws`, to
create `WebSocket` instances.

Although `ws` defines a global `WebSocket`, its constructor's
signature is different from the native one used in browsers as it
accepts a third, `options`, argument.
Since Node applications might want to pass options to their websockets
(to provide a different `http.Agent` for example), we'll add the
ability to pass a `createWebSocket` function in both `CozyRealtime`
and `RealtimePlugin` options. This function will be used to create
`WebSocket` instances.

A default `createWebSocket` using the globally defined `WebSocket`
without the third argument is used when this option is not provided.

### DIFF
--- a/packages/cozy-realtime/src/CozyRealtime.js
+++ b/packages/cozy-realtime/src/CozyRealtime.js
@@ -12,6 +12,7 @@ import {
 } from './config'
 import logger from './logger'
 import {
+  createWebSocket,
   getUrl,
   getToken,
   doctype,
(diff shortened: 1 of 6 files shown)


packages/cozy-realtime/src/CozyRealtime.js
```diff
@@ -27,10 +27,12 @@ class CozyRealtime {
   /**
    * @constructor
    * @param {object} options
-   * @param {CozyClient}  options.client
+   * @param {CozyClient}  options.client A cozy-client instance
+   * @param {Agent} options.agent A Node http.Agent for apps using a custom one to create sockets
```
Crash--, line 31: Agent is undefined. Can you define it even if it's an "any" type?

packages/cozy-realtime/src/RealtimePlugin.js
```diff
@@ -14,9 +14,12 @@ class RealtimePlugin {
    *
    * @constructor
    * @param {CozyClient} client A cozy-client instance
+   * @param {object} options
+   * @param {Agent} options.agent A Node http.Agent for apps using a custom one to create sockets
```
Crash--, line 18: same here

Crash--:
There is something that bothers me in this PR. We are adding Node specific stuff to a "Websocket" available to anyone (browser / node / and so one). We should find a way to have this specific node stuff to a dedicated node stuff.

taratatach:
@Crash-- we don't mention http agents anymore :)

Crash-- (approved):

PR #2158, opened 2023-05-17
feat: Remove cozy-flags from cozy-scanner

Merkur39:
We no longer want to condition the display of the Health theme.


JF-Cozy (approved):
c'est un BC non ? `cozy-flags is no longer required as dependency`

PR #2163, opened 2023-05-19
Fix/harvest flagship

Crash--:
See commits for description but fixes mainly to have a better support of the immersive mode.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountSelectBox/AccountSelectorHeader.tsx
+++ b/packages/cozy-harvest-lib/src/components/AccountSelectBox/AccountSelectorHeader.tsx
@@ -1,9 +1,11 @@
+import cx from 'classnames'
 import React from 'react'
 
 import DialogBackButton from 'cozy-ui/transpiled/react/CozyDialogs/DialogBackButton'
 import DialogTitle from 'cozy-ui/transpiled/react/Dialog/DialogTitle'
 
 import AccountSelectBox from './AccountSelectBox'
+import { useDialogContext } from '../DialogContext'
 
 interface AccountSelectorHeaderProps {
   konnector: { slug: string }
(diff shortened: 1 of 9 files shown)


packages/cozy-harvest-lib/src/components/AccountSelectBox/AccountSelectorHeader.tsx
```diff
@@ -18,27 +20,40 @@ export const AccountSelectorHeader = ({
...
-        accountsAndTriggers={accountsAndTriggers}
-        onChange={(option: { account: { _id: string } }): void => {
-          pushHistory(`/accounts/${option.account._id}`)
-        }}
-        onCreate={(): void => {
-          pushHistory('/new')
-        }}
-        variant="big"
+}: AccountSelectorHeaderProps): JSX.Element => {
+  // @ts-expect-error IDK
```
Ldoppea, line 24: nit: DialogContext should be easily converted to TS, maybe we can try to do it so we don't have to add those ignore comments?

But if we do this we will only be able to partially type the context as it also contain some cozy-ui values that are not typed (useCozyDialog).

Ldoppea (approved):
Everything looks good to me, but I'm not able to 100% approve the [fix: Account creation from the config modal is working](https://github.com/cozy/cozy-libs/pull/2163/commits/f87d61161bfa43f8d2bb240a5bfd42741e1c0fdf) commit as I'm not sure why we do need to remove that prop

Crash--:
> but I'm not able to 100% approve the [fix: Account creation from the config modal is working](https://github.com/cozy/cozy-libs/pull/2163/commits/f87d61161bfa43f8d2bb240a5bfd42741e1c0fdf) commit as I'm not sure why we do need to remove that prop

Same thing as this first fix: https://github.com/cozy/cozy-libs/pull/2142 . The props is already defined with the right data where it needs to be defined and used. Here it was there, for the exemple: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/components/AccountModal.jsx#L191 

Also, we should never use hooks from react-router, always use the hooks from the MountPointProvider

PR #2167, opened 2023-05-23
fix: Wrap IntentTriggerManager inside ComponentsPropsProvider

Crash--:
We added this ComponentsPropsProvider a few weeks ago. We added it to the Routes component, so if we use Harvest from the HarvestRoute, the CannotConnectModal was well displayed.

But if we used Harvest from the IntentTriggerManager, which is used by the cozy-home when called from the Store, then this Provider was not set and we throwed an error.

This PR fix this issue, so the CannotConnectModal is now displayed even from the Intent.


Merkur39 (approved):
Merci! C'est effectivement un cas que je n'avais pas vu 👍

PR #2181, opened 2023-05-31
feat: Send debug if flag is set to true

Crash--:
If the cozy is in debug mode, then let's send DEBUG true to the Launcher in order to display the debug view


Merkur39 (approved):
LGTM

PR #2190, opened 2023-06-13
feat: Add max accounts by konnector paywall

cballevre:
I've found two places that ask you to create an account either with the '/new' route and with an intent that uses TriggerManager from cozy-home. Verification is carried out in both places to limit the opening and closing of successive modals as quickly as possible.

I'll try to optimise the request to retrieve the triggers

**Related PRs:**
- https://github.com/cozy/cozy-client/pull/1356
- https://github.com/cozy/cozy-ui/pull/2452

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountsPaywall/AccountsPaywall.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountsPaywall/AccountsPaywall.jsx
@@ -1,0 +1,34 @@
+// @ts-check
+import React from 'react'
+
+import {
+  MaxAccountsByKonnectorPaywall,
+  MaxAccountsPaywall
+} from 'cozy-ui/transpiled/react/Paywall'
+
+import { computeMaxAccountsByKonnector, computeMaxAccounts } from './helpers'
+
+/**
+ *
+ * @param {Object} props
+ * @param {string} props.reason
+ * @param {import('cozy-client/types/types').IOCozyKonnector} props.konnector
+ * @param {Function} props.onClose
+ */
+const AccountsPaywall = ({ reason, konnector, onClose }) => {
+  if (reason === 'max_accounts_by_konnector') {
+    const max = computeMaxAccountsByKonnector(konnector.slug)
+    return (
+      <MaxAccountsByKonnectorPaywall
+        max={max}
+        konnectorName={konnector.name}
+        onClose={onClose}
+      />
+    )
+  } else if (reason === 'max_accounts') {
+    const max = computeMaxAccounts()
+    return <MaxAccountsPaywall max={max} onClose={onClose} />
+  }
+}
+
+export default AccountsPaywall
--- a/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
+++ b/packages/cozy-harvest-lib/src/components/TriggerManager.jsx
@@ -16,6 +16,7 @@ import { ModalBackButton } from 'cozy-ui/transpiled/react/Modal'
 import Spinner from 'cozy-ui/transpiled/react/Spinner'
 
 import AccountForm from './AccountForm'
+import AccountsPaywall from './AccountsPaywall/AccountsPaywall'
 import FlowProvider from './FlowProvider'
 import HarvestVaultProvider from './HarvestVaultProvider'
 import OAuthForm from './OAuthForm'
@@ -23,6 +24,7 @@ import ComponentsPropsProvider from './Providers/ComponentsPropsProvider'
 import VaultCiphersList from './VaultCiphersList'
 import VaultUnlockProvider from './VaultUnlockProvider'
 import { fetchAccount } from '../connections/accounts'
+import { checkMaxAccounts } from '../helpers/accounts'
 import manifest from '../helpers/manifest'
 import { intentsApiProptype } from '../helpers/proptypes'
 import { findKonnectorPolicy } from '../konnector-policies'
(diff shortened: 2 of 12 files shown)


packages/cozy-harvest-lib/src/components/AccountsPaywall/AccountsPaywall.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
```
Crash--, line 1: Can you add `//@ts-check` at the beginning on every new file you create on harvest please?

packages/cozy-harvest-lib/src/components/AccountsPaywall/AccountsPaywall.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+import {
+  MaxAccountsByKonnectorPaywall,
+  MaxAccountsPaywall
+} from 'cozy-ui/transpiled/react/Paywall'
+
+import { computeMaxAccountsByKonnector, computeMaxAccounts } from './helpers'
+
+const AccountsPaywall = ({ reason, konnector, onClose }) => {
```
Crash--, line 10: jsdoc

cballevre:
I've changed my approach. The check to show paywall is only append in the TriggerManager. This avoids some side effects and makes the code clearer.

Verification is only carried out when an account is created. In the TriggerManager, this corresponds to having the account property set to undefined. That's why, I change the `getInitialStep` to always check paywall condition before display the account form

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
```diff
@@ -73,10 +55,12 @@ export class DumbTriggerManager extends Component {
     this.state = {
       account,
       error: null,
-      step: getInitialStep(props),
+      // On account creation, we need to check max accounts
+      step: account ? 'accountForm' : null,
```
Crash--, line 59: Tu as fait sauter le check `if (konnectorPolicy.saveInVault)` c'est voulu ?

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -242,6 +247,25 @@ export const fetchAccountProcess = async (client, trigger) => {
   return fetchAccount(client, triggersModel.getAccountId(trigger))
 }
 
+export const checkMaxAccounts = async (slug, client) => {
+  const triggersQuery = buildCountTriggersQuery()
+  const { data: triggers } = await client.query(triggersQuery.definition)
```
Crash--, line 252: Cette requête devrait avoir du cache. Tu devrais passer par `fetchQueryAndGetFromState` et mettre une fetchPolicy. Et t'assurer que Harvest écoute bien le temps réel sur les triggers.

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -242,6 +247,25 @@ export const fetchAccountProcess = async (client, trigger) => {
   return fetchAccount(client, triggersModel.getAccountId(trigger))
 }
 
+export const checkMaxAccounts = async (slug, client) => {
```
Crash--, line 250: jsdoc

packages/cozy-harvest-lib/src/helpers/queries.js
```diff
@@ -61,3 +61,12 @@ export const buildAppsByIdQuery = appSlug => ({
     singleDocData: true
   }
 })
+
+export const buildCountTriggersQuery = () => ({
+  definition: Q('io.cozy.triggers')
+    .where({ worker: ['client', 'konnector'] })
+    .indexFields(['worker']),
+  options: {
+    as: 'io.cozy.triggers'
```
Crash--, line 70:
```suggestion
    as: 'io.cozy.triggers/by_worker_client_konnector'
```

packages/cozy-harvest-lib/src/components/TriggerManager.jsx
Crash--: > Verification is only carried out when an account is created. In the TriggerManager, this corresponds to having the account property set to undefined. That's why, I change the getInitialStep to always check paywall condition before display the account form

Je viens juste de lire ça. A quel endroit on check la condition du paywall avant l'affichage du accountForm?

cballevre: On check la condition du paywall dans la fonction `componentDidMount` ligne 210 si c'est une création de compte. On ne bloque pas l'utilisateur quand il veut éditer son compte. 

J'en suis venu à sauter le check `if (konnectorPolicy.saveInVault)` car la fonction aurait ressembler à cela pour passer systématiquement dans le `componentDidMount` avant d'afficher un composant à la création
```
const getInitialStep = ({ account, konnector }) => {
  const konnectorPolicy = findKonnectorPolicy(konnector)
  if (konnectorPolicy.saveInVault) {
    return account ? 'accountForm' : null
  } else {
    return account ? 'accountForm' : null
  }
}
```

L'étape `konnectorPolicy.saveInVault` est refait dans le `componentDidMount` s'il n'y a pas de paywall

packages/cozy-harvest-lib/src/helpers/accounts.js
cballevre: yes for sure

packages/cozy-harvest-lib/src/helpers/accounts.js
cballevre: J'ai l'impression que Harvest écoute bien le temps réel sur les triggers mais je n'ai pas d'impact sur ma query

packages/cozy-harvest-lib/src/helpers/queries.js
Crash--: This is not only way better... Cozy client use a deduplication system based on the id. 

Imagine this stupid code: 

```jsx
const [triggers1, triggers2] = Promise.all([client.query(Q('io.cozy.triggers'), { as : 'io.cozy.triggers'})), client.query(Q('io.cozy.triggers').where({worker: 'client'})]
```

Then, triggers1 and triggers2 will have the same result even though the request is not the same. Your data will be wrong.

Crash-- (approved):

PR #2194, opened 2023-06-16
fix(mespapiers): The cozyMetadata property does not always exist

Merkur39:
Old cozyFiles do not have the `cozyMetadata` property, and files from old konnectors may not have the `cozyMetadata.uploadedBy` property.

As these are essential data for the proper functioning of Mespapiers, they will not appear in the application.

### DIFF
--- a/packages/cozy-mespapiers-lib/src/helpers/queries.js
+++ b/packages/cozy-mespapiers-lib/src/helpers/queries.js
@@ -25,6 +25,7 @@ export const buildFilesQueryWithQualificationLabel = () => {
     'metadata.number',
     'metadata.contractType',
     'metadata.refTaxIncome',
+    'cozyMetadata.uploadedBy',
     'created_at',
     'dir_id',
     'updated_at',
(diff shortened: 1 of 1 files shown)


packages/cozy-mespapiers-lib/src/helpers/queries.js
```diff
@@ -51,9 +52,13 @@ export const buildFilesQueryWithQualificationLabel = () => {
...
           'metadata.qualification.label': {
             $exists: true
-          }
+          },
+          'cozyMetadata.uploadedBy': { $exists: true }
         })
-        .indexFields(['metadata.qualification.label'])
+        .indexFields([
+          'metadata.qualification.label',
+          'cozyMetadata.uploadedBy'
```
paultranvan, line 60: There is no need to index fields inside a partial index, as they actually won't be in the index itself

packages/cozy-mespapiers-lib/src/helpers/queries.js
```diff
@@ -73,9 +78,14 @@ export const buildFilesQueryByLabel = label => ({
...
         type: 'file',
-        trashed: false
+        trashed: false,
+        'cozyMetadata.uploadedBy': { $exists: true }
       })
-      .indexFields(['created_at', 'metadata.qualification'])
+      .indexFields([
+        'created_at',
+        'metadata.qualification',
+        'cozyMetadata.uploadedBy'
```
paultranvan, line 87: Same remark, no need to index ;)

Merkur39: Indeed, it seemed repetitive

PR #2195, opened 2023-06-16
fix: Display reconnection webview for powens konnectors

doubleface:
Even with harvest.inappconnectors.enabled flag activated.

The difference with previous interface is that the "reconnect" button
and the synchronize button have merge.

When the konnector is a powens one and when the error is solvable via
reconnection, the reconnection action (display the powens webview) is
done instead of just launching the konnector.

The reconnection button has exactly the same style as the
synchronization button.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
+++ b/packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
@@ -16,6 +16,7 @@ import { useTrackPage } from '../../../components/hoc/tracking'
 import useMaintenanceStatus from '../../../components/hooks/useMaintenanceStatus'
 import KonnectorUpdateInfos from '../../../components/infos/KonnectorUpdateInfos'
 import * as konnectorsModel from '../../../helpers/konnectors'
+import { intentsApiProptype } from '../../../helpers/proptypes'
 import getRelatedAppsSlugs from '../../../models/getRelatedAppsSlugs'
 import Datacards from '../../Datacards'
 
@@ -32,6 +33,7 @@ export const DataTab = ({
   trigger,
   client,
   flow,
+  intentsApi,
   account
 }) => {
   const { isMobile } = useBreakpoints()
@@ -61,6 +63,8 @@ export const DataTab = ({
           <LaunchTriggerCard
             konnectorRoot={konnectorRoot}
             flow={flow}
+            intentsApi={intentsApi}
+            account={account}
             withMaintenanceDescription
           />
           {isMobile && <Divider style={styles.divider} />}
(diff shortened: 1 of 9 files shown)


packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
```diff
@@ -118,16 +122,27 @@ export const LaunchTriggerAlert = ({
...
+                ) : (
+                  <Button
+                    variant="text"
+                    color={isInError ? 'error' : undefined}
+                    size="small"
+                    disabled={running}
+                    label={t('card.launchTrigger.button.label')}
+                    onClick={SyncButtonAction}
+                  />
+                ))}
```
Crash--, line 145: It seems that `isKonnectorRunnable` is also used here https://github.com/cozy/cozy-libs/blob/b379309603f23d33f7f38c983f9a5ead77f4a516/packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlertMenu.jsx#L28

should we make the same change?

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
```diff
@@ -61,6 +63,8 @@ export const DataTab = ({
           <LaunchTriggerCard
             konnectorRoot={konnectorRoot}
             flow={flow}
+            intentsApi={intentsApi}
+            account={account}
```
Crash--, line 67: Why this addition (of the account)?

Crash--: OK I missed the fact that `LaunchTriggerCard` can be `LaunchTriggerAlert`. 

But you have to change the `LaunchTriggerCard.propTypes` to add the account

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
doubleface: I missed that we could also run the konnector in the menu. I will add the same change here

doubleface: Now the synchronize button in menu triggers the same action. Since the ActionMenuItem does not allow an item to be disabled, we check that the extraParams are available before showing the popup or webview

packages/cozy-harvest-lib/src/components/KonnectorConfiguration/DataTab/index.jsx
doubleface: Also added propTypes for intentsApi property

doubleface:
@Crash-- PR ready

Crash--:
No test? I'm pretty sure we'll want to use the policy next time to get the `syncAction` instead of all these conditions. And we'll bless to have tests that help us to get the right behavior.

Crash-- (approved):

PR #2198, opened 2023-06-20
feat(harvest): Add dismissable alert when running a trigger in CCC+Native

acezard:
## Summary

This PR introduces the `RunningAlert` component and includes unit tests for it. The `RunningAlert` component is designed to display an alert message to the user when a certain condition is met : Konnector running a job, on Flagship App, as a CCC.

### Major changes
- Added `RunningAlert` component under `src/components/cards`.
- Added unit tests for `RunningAlert` to ensure it renders correctly based on the state of `isFlagshipApp`.

#### RunningAlert Component
- The component uses the `cozy-ui` library for styling and icons.
- It renders an alert with a title and body text which is obtained through the `useI18n` hook from the `cozy-ui` library. This makes it easy to handle translations.
- The visibility of the alert is dependent on the `isFlagshipApp` function from the `cozy-device-helper`. When the app is a flagship app, the alert is visible (if the parent card allows it depending on konnector/job state/properties)
- A user can dismiss the alert by clicking a button.

#### Unit Tests
- Mocks are used to handle the external modules and hooks. This isolates the component and ensures that the tests are not affected by the dependencies.
- The tests check three main functionalities:
    1. Whether the component renders correctly when `isFlagshipApp` returns true.
    2. That it does not render when `isFlagshipApp` returns false.
    3. Whether the alert is hidden when the dismiss button is clicked.

## How to test

1. Pull the changes from this PR.
2. Run the unit tests by executing `yarn test`.
3. To visually inspect the component, insert it in a view and check if it's rendering as expected.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
+++ b/packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
@@ -13,6 +13,7 @@ import Typography from 'cozy-ui/transpiled/react/Typography'
 import { makeStyles } from 'cozy-ui/transpiled/react/styles'
 
 import LaunchTriggerAlertMenu from './LaunchTriggerAlertMenu'
+import { RunningAlert } from './RunningAlert'
 import { makeLabel } from './helpers'
 import { isDisconnected } from '../../helpers/konnectors'
 import { getAccountId, getKonnectorSlug } from '../../helpers/triggers'
@@ -64,6 +65,16 @@ export const LaunchTriggerAlert = ({
   const isKonnectorRunnable = konnectorPolicy.isRunnable()
   const isClisk = konnectorPolicy.name === 'clisk'
   const isKonnectorDisconnected = isDisconnected(konnector, trigger)
+
+  const shouldDisplayRunningAlert = () => {
+    if (isInError) return false
+    if (isInMaintenance) return false
+    if (!isKonnectorRunnable) return false
+    if (isKonnectorDisconnected) return false
+    if (running && konnector.clientSide) return true
+
+    return false
+  }
 
   useEffect(() => {
     if (status === SUCCESS) {
@@ -202,6 +213,9 @@ export const LaunchTriggerAlert = ({
           )}
         </div>
       </Alert>
+
+      {shouldDisplayRunningAlert() && <RunningAlert />}
+
       <Snackbar
         open={showSuccessSnackbar}
         onClose={() => setShowSuccessSnackbar(false)}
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/src/components/cards/RunningAlert.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const RunningAlert = (): JSX.Element | null => {
+  const { t } = useI18n()
+  const [isVisible, setIsVisible] = useState(isFlagshipApp())
+  const handleClose = (): void => setIsVisible(false)
+
+  return isVisible ? (
+    <Alert
+      className="u-mt-1"
+      block
+      icon={<Icon icon={<ArrowUp />} color="var(--primaryColor)" />}
```
Crash--, line 20: @JF-Cozy Is it the API? Can't we use "primary" directly?

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
```diff
@@ -65,6 +66,16 @@ export const LaunchTriggerAlert = ({
...
 
+  const shouldDisplayRunningAlert = () => {
+    if (isInError) return false
+    if (isInMaintenance) return false
+    if (!isKonnectorRunnable) return false
+    if (isKonnectorDisconnected) return false
+    if (running && konnector.clientSide) return true
+
+    return false
+  }
```
Crash--, line 77: Wondering if this should not be inside the policy?

acezard: I think that could be better yeah. @LucsT what do you think please?

Crash--: @doubleface will have more insights to share I think. I already shared my though here about the policy https://github.com/cozy/cozy-libs/pull/2195#issuecomment-1598259418 

Maybe it can be done next time...

packages/cozy-harvest-lib/src/components/cards/RunningAlert.tsx
acezard: I'm gonna check anyway

acezard: @Crash-- updated, the icon inherits the correct colour from the alert component default style currentColor. So I don't need to specify anything here

JF-Cozy: seems to be solved :)

packages/cozy-harvest-lib/src/components/cards/LaunchTriggerAlert.jsx
doubleface: Since I already moved the same kind of logic in policies for https://github.com/cozy/cozy-libs/pull/2195, I can do it if this PR is merged before mine

doubleface (approved):

Crash-- (approved):

PR #2205, opened 2023-06-23
fix: Show accounts paywall for oAuth konnector

cballevre:
Stupid error: the check to display the paywall was too early and was bypassed by the oAuth form


Crash--:
No stupid mistakes. Just unwritten tests that would have helped us. 

I don't know if it's easy, but we should at least give a try.

cballevre:
@Crash-- I updated the test to match every scenario to show a paywall

Crash-- (approved):

PR #2208, opened 2023-06-26
fix(mespapiers): Step modal with Radio buttons

Merkur39:
This `useEffect` had been removed, among other things, during the last evolution of the component.
But it is really necessary for the proper functioning of the functionality (creation/editing).


JF-Cozy (approved):
hors contexte on perd un peu le fil, peut-être mettre un commentaire à côté du useEffect qui explique pourquoi via un useEffect on utilise un setter d'un parent...

PR #2225, opened 2023-07-05
feat: Update cozy-ui to 88 with codemods

zatteo:
BREAKING CHANGE:
- you need to update cozy-ui to >=88.2.1


Crash--:
can you tag it as a breaking change please?

PR #2226, opened 2023-07-05
feat: Update cozy-ui to 88 and other deps

zatteo:
BREAKING CHANGE:
- you need to update cozy-ui to 88.2.1
- you need to update cozy-client to 38.9.2

### DIFF
--- a/packages/cozy-scanner/package.json
+++ b/packages/cozy-scanner/package.json
@@ -25,24 +25,30 @@
     "@babel/cli": "7.16.8",
     "@babel/core": "7.16.12",
     "@testing-library/react": "10.4.9",
+    "@cozy/minilog": ">=1.0.0",
     "babel-jest": "26.6.3",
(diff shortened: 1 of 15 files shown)


packages/cozy-scanner/package.json
```diff
@@ -60,6 +60,9 @@
   },
   "sideEffects": false,
   "dependencies": {
-    "cozy-device-helper": "^2.7.0"
+    "@cozy/minilog": "1.0.0",
+    "cozy-device-helper": "^2.7.0",
+    "cozy-flags": "^2.8.6",
+    "cozy-logger": "^1.10.1"
```
Crash--, line 66: shouldn't be devDep & peerDep ?

zatteo: So not @cozy/minilog ?

Crash-- (approved):

PR #2235, opened 2023-07-10
fix: Cleaning of a few old stuff

Crash--:
BREAKING CHANGE: LegacyTriggerManager should only be used internally by Harvest. Third party app / lib, shoud use HarvestRoutes or IntentTriggerManager.

Also a few other breaking change by removing exported component that should not be used anymore.


Merkur39 (approved):
Let's roll, good cleaning 👍

PR #2236, opened 2023-07-12
fix: Wrong export for DumbTriggerManager

cballevre:
DumbTriggerManager (cf: [code](https://github.com/cozy/cozy-libs/blob/ed9e18998c30701d07292c44cea46534525df182/packages/cozy-harvest-lib/src/components/DumbTriggerManager.jsx#L28)) imported as default in TriggerManager (cf: [code](https://github.com/cozy/cozy-libs/blob/ed9e18998c30701d07292c44cea46534525df182/packages/cozy-harvest-lib/src/components/TriggerManager.jsx#L12)) but just export which lead to white screen in cozy-home


Crash-- (approved):
thanks, my bad

PR #2239, opened 2023-07-13
fix(mespapiers): Behavior of information step on mobile device

Merkur39:
when input is focused.

This fix prevents the content of the information step modal from being off-center when you focus the input.


Crash--:
What is the status of this PR @Merkur39 ?

Merkur39:
This same fix appeared in this PR 🤷 
Done ✅

PR #2241, opened 2023-07-18
fix(sharing): Use `.fetchQueryAndGetFromState` instead `.query`

Merkur39:
When using `fetchPolicy`, the `client.query` method does not work correctly,
use `client.fetchQueryAndGetFromState` instead.

Known issue: cozy/cozy-client#931


Crash--:
C'était quoi le bug rencontré ?

PR #2268, opened 2023-08-07
feat: Removed code needed before full deploy of powens webviews

doubleface:
- Removed all references to harvest.bi.* flags
- Removed contracts edition in harvest (handled in powens webview)
- Removed budget-insight.js policy (biWebview policy is used for all powens konnectors now)


Crash-- (approved):
Champion! \o/

PR #2269, opened 2023-08-07
feat: Removed the use of powens bank ids -> uuids

doubleface:
Now powens uuids are used, which removes the need to have different
identifiers according to the current environnement and makes it a lot
easier to add new banks (just need to add the bank to the registry and
add corresponding uuids in the manifest of the konnectors.

But to fully work, this needs to have
https://github.com/cozy/cozy-stack/pull/4082 fully deployed in
production (or else the webview won't open on a correct url)

But since uuids are available in banking konnectors manifest since a long time, this can be deployed without any update in banking konnectors.

### DIFF
--- a/packages/cozy-harvest-lib/src/policies/biWebView.spec.js
+++ b/packages/cozy-harvest-lib/src/policies/biWebView.spec.js
@@ -27,12 +27,10 @@ const realtimeMock = {
   unsubscribe: jest.fn()
 }
 
-const TEST_BANK_COZY_ID = '100000'
-
 const konnector = {
   slug: 'boursorama83',
-  parameters: {
-    bankId: TEST_BANK_COZY_ID
+  biUuids: {
+    'test konnector uuid': 'test konnector'
   },
   partnership: {
     domain: 'https://budget-insight.com'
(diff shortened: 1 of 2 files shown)


packages/cozy-harvest-lib/src/policies/biWebView.spec.js
```diff
@@ -269,43 +262,43 @@ describe('fetchExtraOAuthUrlParams', () => {
...
-  it('should add id_connector param for multiple bank ids', async () => {
+  it('should add connector_uuids param for multiple bank uuids', async () => {
     const client = new CozyClient({
       uri: 'http://testcozy.mycozy.cloud'
     })
     client.query = jest.fn().mockResolvedValue({
       data: {
         timestamp: Date.now(),
-        code: 'bi-temporary-access-token-123',
-        biMapping: { [TEST_BANK_COZY_ID]: 2 }
```
Crash--, line 280: Does that mean that we stored this `biMapping` inside the account? Should we remove it now?

Crash-- (approved):
Nice ! \o/

packages/cozy-harvest-lib/src/policies/biWebView.spec.js
doubleface: This will be done by the banking konnector : https://gitlab.cozycloud.cc/scrapers/banking-konnectors/-/merge_requests/431

PR #2277, opened 2023-08-17
fix(mespapiers): Restrict mime type images for paper creation

Merkur39:
Restrict mime type images, because the process of creating a paper does not accept mime types like `image/svg+xml`, `image/webp`, etc


Crash-- (approved):
Nice catch ;)

PR #2278, opened 2023-08-18
feat(mespapiers): Remove the `stepIndex` property from the config file

Merkur39:
This property was added when the project was created.
Its main purpose was to guarantee the order of the steps.
Which is useless in JS because the order of JSON arrays is guaranteed by the language. (https://www.rfc-editor.org/rfc/rfc8259.html#section-1)

This will make it easier to manipulate the steps in addition to making them lighter.


JF-Cozy (approved):
Je pense qu'on pourrait ajouter une petite note dans la doc qui indique que les steps sont dans l'ordre du tableau... ?j

va pour des steps qui commencent à 0 et non à 1 ;)

PR #2285, opened 2023-08-30
fix: Skip the document not found during indexing after a realtime update

cballevre:
When a new share is made, realtime trigger the indexation of all documents in the new sharing as well as the shortcut. If the recipient has not synchronised the sharing, the document does not exist, leading to errors in the indexing script. This is why we only index files if the share is ready, so that the files exist

### DIFF
--- a/packages/cozy-sharing/src/state.js
+++ b/packages/cozy-sharing/src/state.js
@@ -395,6 +395,31 @@ export const getSharingDocIds = sharing => {
   return docs
 }
 
+/**
+ * Get ids of shared documents, but only if sharing is ready so files exist
+ * @param {object} sharing
+ * @param {string} instanceUri
+ * @returns {string[]} List of document ids of a sharing
+ */
+export const getExternalSharingIds = (sharing, instanceUri) => {
+  const member = sharing.attributes.members.find(
+    member => member.instance === instanceUri
+  )
+
+  let docs = []
+  if (member?.status === 'ready') {
+    docs = sharing.attributes.rules
+      .map(r => r.values)
+      .reduce((acc, val) => acc.concat(val), [])
+  }
+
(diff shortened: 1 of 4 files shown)


Crash--:
> When a new share is made, realtime trigger the indexation of all documents 

How? I thought we just send listen a realtime message and update cozy-client state (with RealTimeQueries or something similar). Maybe you don't talk about couchdb indexation but something else? 


If I understand correctly: 
- We receive a realtime event about a created sharing https://github.com/cozy/cozy-libs/blob/053e1862f4be6e4175bdf8e7d6398702bc41868f/packages/cozy-sharing/src/SharingProvider.jsx#L136-L140
- Then we update or create it https://github.com/cozy/cozy-libs/blob/053e1862f4be6e4175bdf8e7d6398702bc41868f/packages/cozy-sharing/src/SharingProvider.jsx#L151-L170 
- And we add an io.cozy.files id that doesn't exist on the instance yet. Is that correct? 

If yes, maybe we should add this `io.cozy.files id` only when the sharing status is done ?

cballevre:
@zatteo 🙏

packages/cozy-sharing/src/state.js
```diff
@@ -395,6 +395,31 @@ export const getSharingDocIds = sharing => {
...
+ * @param {string} instanceUri
+ * @returns {string[]} List of document ids of a sharing
+ */
+export const getExternalSharingIds = (sharing, instanceUri) => {
+  const member = sharing.attributes.members.find(
+    member => member.instance === instanceUri
+  )
+
+  let docs = []
+  if (member?.status && member.status === 'ready') {
```
zatteo, line 410: nit++: member?.status === 'ready' is equivalent

zatteo (approved):

Crash--:
nice!

PR #2296, opened 2023-09-29
feat: Exclude CCC from max accounts paywall

cballevre:


doubleface (approved):

packages/cozy-harvest-lib/src/helpers/accounts.js
```diff
@@ -255,9 +257,14 @@ export const fetchAccountProcess = async (client, trigger) => {
...
-    triggersQuery
-  )
+  const { data } = await client.fetchQueryAndGetFromState(triggersQuery)
+
+  const triggers = data.filter(trigger => {
+    if (flag('harvest.accounts.exclude-ccc')) {
+      return trigger.worker !== 'client'
+    }
+    return true
+  })
```
Crash--, line 267: You should do the filter only if the flag is on. No need to map on the triggers if not, no?

PR #2303, opened 2023-10-09
feat: Add information button redirecting to the store

paultranvan:
When a opening a konnector, we used to only display the vendor link into the 'Information' section.
We now add a button that redirects to the store, so the user can access the description there.

Here what it looks like: 
image

Note we mention a "Cozy Cloud Store" but this might eventually change to a dynamic name the day we need to support different names


Crash--:
Can you add a screenschot of the result? Also can you squash the commits?

Crash-- (approved):

PR #2309, opened 2023-10-17
fix: Minor adjustment before deploy ccc

cballevre:
**Related PRs:**
- https://github.com/cozy/cozy-client/pull/1404

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm/InstallFlagshipButton.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/InstallFlagshipButton.jsx
@@ -1,0 +1,25 @@
+import React from 'react'
+
+import { getFlagshipDownloadLink } from 'cozy-client/dist/models/utils'
+import Buttons from 'cozy-ui/transpiled/react/Buttons'
+import { useI18n } from 'cozy-ui/transpiled/react/providers/I18n'
+
+const InstallFlagshipButton = ({ className }) => {
+  const { t, lang } = useI18n()
+
+  const handleClick = () => {
+    let downloadLink = getFlagshipDownloadLink(lang)
+    window.open(downloadLink, '_blank')
+  }
+
+  return (
+    <Buttons
+      className={className}
+      fullWidth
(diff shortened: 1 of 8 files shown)


doubleface (approved):

packages/cozy-harvest-lib/src/locales/fr.json
```diff
@@ -545,7 +545,7 @@
   },
   "connectionBackdrop": {
     "connecting": "Connexion en cours",
-    "progress": "Nous interrogeons %{name}, cela peut prendre jusqu'à 30 secondes..."
+    "progress": "Etablissement de la connexion à %{name}, cela peut prendre jusqu'à 30 secondes.."
```
Crash--, line 548: Établissement

packages/cozy-harvest-lib/src/components/AccountForm/InstallFlagshipButton.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+  const handleClick = () => {
+    let downloadLink = `https://cozy.io/${lang}/download`
+    if (isMobile()) {
+      downloadLink = isAndroid()
+        ? `https://play.google.com/store/apps/details?id=io.cozy.flagship.mobile&hl=${lang}`
+        : `https://apps.apple.com/${lang}/app/id1600636174`
+    }
+    window.open(downloadLink, '_blank')
+  }
```
Crash--, line 18: I'm not sure about the links. What about Ma bulle? If the Cozy's context is GL, should we point to ma bulle? 

Also should we always push the links even if there is an installed app? (aka OAuth client linked to the app)?

cballevre: Yes, you're right, I'd forgotten. I used the same code as in store. So we also have to make an adaptation there. @Ldoppea Is there a way of obtaining the application identifier from the context?

Checking whether an application is installed is planned in another ticket later. The product should come back to me with the design instead of the button.

PR #2313, opened 2023-10-20
feat(checkMaxAccounts): Exclude connectors under maintenance

cballevre:
This PR is used to exclude konnectors under maintenance from the calculation for displaying the paywall about maximum accounts

**Related PRs:**
- https://github.com/cozy/cozy-client/pull/1406
- https://github.com/cozy/cozy-client/pull/1407

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/queries.js
+++ b/packages/cozy-harvest-lib/src/helpers/queries.js
@@ -84,3 +84,11 @@ export const buildCountTriggersQuery = () => ({
     fetchPolicy: defaultFetchPolicy
   }
 })
+
+export const buildAppsRegistryMaintenance = () => ({
+  definition: Q('io.cozy.apps_registry').getById('maintenance'),
+  options: {
+    as: 'io.cozy.apps_registry/maintenance',
+    fetchPolicy: defaultFetchPolicy
+  }
+})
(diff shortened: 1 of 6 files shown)


packages/cozy-harvest-lib/src/helpers/queries.js
```diff
@@ -84,3 +84,11 @@ export const buildCountTriggersQuery = () => ({
...
   }
 })
+
+export const buildAppsRegistryMaintenance = () => ({
+  definition: Q('io.cozy.apps_registry').getById('maintenance'),
+  options: {
+    as: 'io.cozy.apps_registry/maintenance',
+    fetchPolicy: defaultFetchPolicy
+  }
+})
```
Crash--, line 94: maybe you can rework this hook in order to use it here? (and make it use your new mechanism) 

https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/hooks/useAppsInMaintenance.jsx

Crash--: This is a nit for this PR since you can't use the hook directly. 

But I think you should update the hook anyway in order to optimise it by using a useQuery call ;)

cballevre: Very good idea to optimise performance at home. I changed the hook to use the same request ([PR](https://github.com/cozy/cozy-client/pull/1407))

Crash-- (approved):

PR #2315, opened 2023-10-24
feat: Allow to offer account with flag harvest.accounts.offers

cballevre:
The main idea behind this flag is to offer user accounts for a given period and a given konnector. This will also be used for migration to CCC so that users have the same number of connectors at the end

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountsPaywall/helpers.js
+++ b/packages/cozy-harvest-lib/src/components/AccountsPaywall/helpers.js
@@ -45,11 +45,61 @@ export function computeMaxAccounts() {
 }
 
 /**
+ * @typedef {Object} AccountCountByKonnector
+ * @property {string} slug - Slug of the konnector
+ * @property {number} count - Number of accounts for the konnector
+ */
+
+/**
+ * @typedef {Object} KonnectorOffer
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/AccountsPaywall/helpers.js
```diff
@@ -44,12 +44,58 @@ export function computeMaxAccounts() {
...
+ * @typedef {Object} KonnectorOffer
+ * @property {string} slug - Slug of the konnector
+ * @property {number} offer - Number of accounts offered for the konnector
+ * @property {string} [expiresAt] - Date of expiration of the offer
+ */
+
 /**
  *
- * @param {number} nbAccounts the current number of account for the konnector to check
+ * @param {AccountCountByKonnector[]} accounts the accounts to check
```
Crash--, line 62: you can import the IoCozyAccounts type from cozy-client, no,

cballevre: I'm not sure, the cozy-client type doesn't represent the result of the number of accounts per connector. This object only contains a connector slug and its associated number of accounts. Maybe it's the name that's confusing 🤔

Crash-- (approved):

PR #2322, opened 2023-10-31
fix(harvest): Refactor SuggestionModal to standards

acezard:
This modal seems to use an older version of the Cozy UI library

To fix our issue we update the dialog to the new CozyDialogs interface. It comes with minor display variations.

before: 
{F02E7172-60D2-4977-95BA-74402F880109}

after updating to CozyDialogs
image
image


Crash--:
Why can't we use the new dialog?

acezard:
> Why can't we use the new dialog?

we can, I wasn't sure if it was possible to do it right now in terms of priorities. Doesn't look like a long refactor at all but I preferred to be cautious.
Also, I was under the impression that all harvest dialogs had been refactored already, so not sure why this one is different

PR #2323, opened 2023-11-02
feat: Add a `hideCross` method

Crash--:
Some services can handle by themself the display of the cross to close the intent modale. So in order to let them do that, we can now pass an `onHideCross` method to the intent.

```
hideCross(){
    this.setState({
      closable: false
    })
  }

<Modal
        {...modalProps}
        key="modal"
        className={styles.intentModal}
        closeBtnClassName={styles.intentModal__cross}
        dismissAction={this.dismiss}
        overflowHidden
        closable={this.state.closable}
      >
        <IntentIframe
          action={action}
          create={create}
          data={options}
          onCancel={this.dismiss}
          onError={onError}
          onTerminate={onComplete}
          type={doctype}
          onHideCross={this.hideCross}
        />
      </Modal>
```

Also add showCross

### DIFF
--- a/packages/cozy-interapp/src/client.js
+++ b/packages/cozy-interapp/src/client.js
@@ -98,6 +98,16 @@ export function start(createIntent, intent, element, data, options = {}) {
         } finally {
           dom.show(iframe)
         }
+      },
+      onHideCross: () => {
+        if (options.onHideCross) {
+          options.onHideCross()
+        }
+      },
+      onShowCross: () => {
+        if (options.onShowCross) {
+          options.onShowCross()
+        }
       }
     })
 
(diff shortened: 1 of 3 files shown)


packages/cozy-interapp/src/client.js
```diff
@@ -98,6 +98,16 @@ export function start(createIntent, intent, element, data, options = {}) {
         } finally {
           dom.show(iframe)
         }
+      },
+      onHideCross: () => {
+        if (options.onHideCross) {
```
cballevre, line 103: maybe a `typeof options.onHideCross === 'function'` is better 🤔

cballevre (approved):

PR #2349, opened 2023-12-08
[SCR-22] fix: No more disabled submit button on 2nd 2FA

doubleface:
When the konnector was asking 2FA code multiple times, the user could
not submit the 2nd code since the submit button was forefever disabled.

This bug was due to the dependency on "konnectorRunning" in the flow
state. I don't know why this dependency was needed and neither the git
history or unit tests would help me.

But with this condition removed, every 2FA code are asked properly.
I also changed the related unit test to have more tests on this case
(which failed silently before).

### DIFF
--- a/packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
+++ b/packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
@@ -47,7 +47,7 @@ export class TwoFAModal extends PureComponent {
     const flowState = this.props.flow.getState()
     this.setState({
       hasErrored: flowState.twoFARetry,
-      isJobRunning: flowState.twoFARunning || flowState.konnectorRunning
+      isJobRunning: flowState.twoFARunning
     })
   }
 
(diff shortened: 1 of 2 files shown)


Crash-- (approved):

packages/cozy-harvest-lib/src/components/TwoFAModal.jsx
```diff
@@ -47,7 +47,7 @@ export class TwoFAModal extends PureComponent {
     const flowState = this.props.flow.getState()
     this.setState({
       hasErrored: flowState.twoFARetry,
-      isJobRunning: flowState.twoFARunning || flowState.konnectorRunning
```
JF-Cozy, line 50: est-ce que twoFARunning est toujours définit ? La condition précédente ce n'était pas pour gérer un cas ou la 2FA n'était pas prise en charge, et donc on utilisait un autre fonctionnement qui nécessitait de se baser sur konnectorRunning (les CCC peut-être ? 🤔 ) ? Est-ce que tu as testé cet "autre cas", et si oui marche t-il encore ?

Crash--: Le code semble dit que twoFARunning est toujours défini oui https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/models/ConnectionFlow.js#L1001

PR #2355, opened 2023-12-15
feat(cozy-intent): Add hook to get remote register

acezard:
temp

```ts
  const remoteMethods = useRemoteMethods()

  useEffect(() => {
    if (remoteMethods) {
      console.log('💥💥💥', remoteMethods) // Record<string, boolean>
    }
  }, [remoteMethods])
```


acezard:
Spec will change/has changed, closing without deleting for now

PR #2356, opened 2023-12-15
feat(mespapiers): Use CozyDevTools instead FlagSwitcher

Merkur39:
`CozyDevTools` offers better management of flags, it provides other useful tools in development (visualization of requests, version number of dependencies).

The “switcher” flag is replaced by “debug” to display the `CozyDevTools` component.


cballevre (approved):
Nice 👍

PR #2358, opened 2023-12-18
fix: Display the right konnector

Crash--:
We can have several konnectors for a paper but we always passed the first on the array resulting we displayed something wrong to the user.

Let's pass all the konnectors & slugs and find the right one.

### DIFF
--- a/packages/cozy-mespapiers-lib/src/components/Papers/Empty/EmptyWithKonnectors.jsx
+++ b/packages/cozy-mespapiers-lib/src/components/Papers/Empty/EmptyWithKonnectors.jsx
@@ -4,28 +4,36 @@ import React from 'react'
 import EmptyNoHeader from './EmptyNoHeader'
 import EmptyWithHeader from './EmptyWithHeader'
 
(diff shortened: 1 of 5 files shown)


packages/cozy-mespapiers-lib/src/components/Papers/Empty/EmptyWithKonnectors.jsx
```diff
@@ -4,33 +4,41 @@ import React from 'react'
...
+    return (
+      <EmptyWithHeader key={index} konnector={konnector} account={account} />
+    )
+  })
 }
 
-EmptyWithKonnector.propTypes = {
-  konnector: PropTypes.object,
+EmptyWithKonnectors.propTypes = {
+  konnectors: PropTypes.object,
```
Merkur39, line 36:
```suggestion
  konnectors: PropTypes.array,
```

Merkur39 (approved):
Thanks ❤️ 
(just fix Empty.spec)

Merkur39 (approved):

PR #2360, opened 2023-12-18
feat(cozy-intent): Add `componentId` parameter to `tryEmit`

Ldoppea:
In the Flagship app we want to implement a new API for FlagshipUI

This new API requires all calls to contains a `componentId` parameter and so we want the CozyWebViews to pass this parameter

To allow this, we want the `tryEmit` to require this parameter. Then it will pass this parameter to `setFlagshipUI` and `setTheme` calls that would trigger the new FlagshipUI on native side

### DIFF
--- a/packages/cozy-intent/src/api/services/NativeService.ts
+++ b/packages/cozy-intent/src/api/services/NativeService.ts
@@ -12,7 +12,13 @@ import {
   numbers,
   WebviewMethods
 } from '../../api'
-import { getErrorMessage, interpolate, isNativeDevMode } from '../../utils'
+import {
+  getErrorMessage,
+  interpolate,
+  isFlagshipUiArgsArray,
+  isNativeDevMode,
+  isThemeArg
+} from '../../utils'
 
 const log = debug('NativeService')
 
@@ -94,13 +100,18 @@ export class NativeService {
 
     const parsedEvent = this.parseNativeEvent(event)
 
-    if (parsedEvent.methodName === 'setFlagshipUI' && parsedEvent.args) {
-      // @ts-expect-error we know that `setFlagshipUI` args are in an array
+    if (
+      parsedEvent.methodName === 'setFlagshipUI' &&
+      isFlagshipUiArgsArray(parsedEvent.args)
+    ) {
       parsedEvent.args[0].componentId = componentId
     }
 
-    if (parsedEvent.methodName === 'setTheme' && parsedEvent.args) {
-      // @ts-expect-error we know that `setTheme` args is a string and we want to convert it to an object
+    if (
+      parsedEvent.methodName === 'setTheme' &&
+      Array.isArray(parsedEvent.args) &&
+      isThemeArg(parsedEvent.args[0])
+    ) {
       parsedEvent.args[0] = {
         homeTheme: parsedEvent.args[0],
         componentId: componentId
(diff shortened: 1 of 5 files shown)


packages/cozy-intent/src/api/services/NativeService.ts
```diff
@@ -86,11 +86,27 @@ export class NativeService {
...
+  public tryEmit = async (
+    event: NativeEvent,
+    componentId: string
+  ): Promise<void> => {
     if (!this.isNativeEvent(event)) return
 
     const parsedEvent = this.parseNativeEvent(event)
 
+    if (parsedEvent.methodName === 'setFlagshipUI' && parsedEvent.args) {
+      // @ts-expect-error we know that `setFlagshipUI` args are in an array
```
acezard, line 98: Not sure to understand, if args is still a string at this point, accessing [0] would be a letter value? If it can be an array, we should update the `PostMeMessage` interface that is returned by `parseNativeEvent()`

packages/cozy-intent/src/api/services/NativeService.ts
```diff
@@ -86,11 +86,27 @@ export class NativeService {
...
+      parsedEvent.args[0].componentId = componentId
+    }
+
+    if (parsedEvent.methodName === 'setTheme' && parsedEvent.args) {
+      // @ts-expect-error we know that `setTheme` args is a string and we want to convert it to an object
```
acezard, line 103: Same as above 👍

packages/cozy-intent/src/api/services/NativeService.ts
```diff
@@ -86,11 +86,27 @@ export class NativeService {
   ): Promise<Connection> =>
     await ParentHandshake(messenger, this.localMethods, numbers.maxAttempts)
 
-  public tryEmit = async (event: NativeEvent): Promise<void> => {
+  public tryEmit = async (
+    event: NativeEvent,
+    componentId: string
```
acezard, line 91: `componentId` will always be defined? It's not marked as optional here. Does it mean every single `webviewIntent.call` necessitate an ID, even if it doesn't touch the UI scope?

acezard (commented):
Globally I still think this looks to introduce some more complexity. But as for the native part of this refactor, I trust your judgment on this

packages/cozy-intent/src/api/services/NativeService.ts
Ldoppea: No, this componentId would be injected by the FlagshipApp here: https://github.com/cozy/cozy-flagship-app/pull/1082/files#diff-94ee99771b8d7d089bb046a665823be8eedf5c5a317115debff9a97a69638ce4R218

This has been done this way so we don't need to update the cozy-apps and retro-compatibility will be ensured

If I understand it correctly, the code I linked is the only place in our entire code base where `tryEmit()` is called

acezard: but according to the interface https://github.com/cozy/cozy-flagship-app/pull/1082/files#diff-189dde1ff47afcac8c4a195539f67518014f77da3802e819c641838fb7739ebd `componentId` can be `undefined` at some point (not sure about the context, but to me it seems like intended). 
Basically if this variable can be `undefined` I think the cozy-intent should be prepared for it, as in 
```ts
type tryEmit = (event: NativeEvent, componentId?: string) => Promise<void>
```

acezard (approved):

packages/cozy-intent/src/api/services/NativeService.ts
Ldoppea: This was an error on Flagship App side. As this is always injected from `tryEmit`, then it cannot be undefined on Flagship App side.

Fixed in https://github.com/cozy/cozy-flagship-app/compare/4bb373b0bb797f98fd9ad1c915466c499c38a22f..7b2a6c9394ef67df5b5b4b74b361742eb5bd9872

PR #2361, opened 2023-12-19
fix: An error when file doesn't have a createdByApp cozyMetadata

Crash--:
In https://github.com/cozy/cozy-libs/commit/72e5f5f073d6ddc575cae4cef26b4ca9624efa54 we replaced the mango query to only check the existence of `createdByApp`

But we were using `uploadedBy` later in the code. For old files, we can have a `createdByApp` without a `uploadedBy` attribute.

Since we check the existence of `createdByApp` let's use this field later on the code, like that we are sure that this field exists.


Merkur39 (approved):
Thanks ❤️

PR #2364, opened 2023-12-19
feat(mespapiers): Increase the maximum size of images before integrating them into PDF

Merkur39:
Today we are resizing uploaded images to 900px, for optimization reasons.
It turns out that this resolution is too low, resulting in poor quality PDFs.
We have chosen to push the limit quite high, which should guarantee better quality PDFs.


JF-Cozy (approved):
effectivement en print si on utilise des images 72 DPI il faut une plus grande taille 👍

PR #2370, opened 2023-12-29
test: Add story ConnecionBackdrop

Crash--:
We have perf issue on this component. We suspect the LinerProgress component to take too much CPU. 

I added a story in order to be able to debug that later.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/AccountForm/ConnectionBackdrop.stories.jsx
+++ b/packages/cozy-harvest-lib/src/components/AccountForm/ConnectionBackdrop.stories.jsx
@@ -1,0 +1,14 @@
+import { ConnectionBackdrop } from './ConnectionBackdrop'
+
+const meta = {
+  component: ConnectionBackdrop,
+  arg: {
+    name: 'EDF'
+  }
+}
+
+export default meta
+
+export const Default = {
+  name: 'Default'
+}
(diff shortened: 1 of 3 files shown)


packages/cozy-harvest-lib/src/components/AccountForm/ConnectionBackdrop.stories.jsx
```diff
@@ -0,0 +1,18 @@
+import { ConnectionBackdrop } from './ConnectionBackdrop'
+// import { UnrunnableAlert } from '../cards/UnrunnableAlert'
```
cballevre, line 2: This is not used anymore I think we can remove it

packages/cozy-harvest-lib/src/components/AccountForm/ConnectionBackdrop.stories.jsx
```diff
@@ -0,0 +1,18 @@
...
+    name: 'EDF'
+  }
+}
+
+export default meta
+
+export const Default = {
+  name: 'Default',
+  args: {
+    konnectorName: 'EDF'
```
cballevre, line 16: I don't think we need this argument because the parameter is `name` and it's already defined in the meta object

cballevre (approved):

PR #2371, opened 2024-01-02
fix(mespapiers): `makeContactFlexsearchProps` must handle string emails

Merkur39:
On the connector side, the `email` field of the `Contact` object could be a string instead of an object array.
Issue fixed here: konnectors/libs#987
Pending the propagation of the fix, it is relevant to predict the case here.


Crash-- (approved):
thanks <3

PR #2378, opened 2024-01-05
chore(cozy-flags): Fix types of CozyClient

Merkur39:


Crash--:
Same, types fix should be tagged as fix.

PR #2379, opened 2024-01-05
chore(cozy-logger): Fix types

Merkur39:
The nested `declare` is not necessary and may cause problems on `cozy-client`.


Crash--:
This commit should have been tagged as fix because here a new version will not be triggered.

PR #2406, opened 2024-01-19
feat(eslint-config): Update to eslint-config version 8 (COZY-31)

acezard:
## Changed

- Eslint config will not install `devDependencies` anymore. The mono-repo will be responsible for installing them. This fixes a variety of issues with dependencies management inside the mono-repo. This won't affect external projects.
- Update `@babel/eslint-parser` from `7.19.1` to `7.23.3`.
- Update `@typescript-eslint/eslint-plugin` from `5.54.0` to `6.19.0`. (see [full release notes](https://typescript-eslint.io/blog/announcing-typescript-eslint-v6/))
- Update `@typescript-eslint/parser` from `5.54.0` to `6.19.0`.
- Update `eslint` from `8.35.0` to `8.56.0`.
- Update `eslint-config-prettier` from `8.6.0` to `9.1.0`.
- Update `eslint-plugin-import` from `2.27.5` to `2.29.1`.
- Update `eslint-plugin-jest` from `27.2.1` to `27.6.3`.
- Update `eslint-plugin-prettier` from `4.2.1` to `5.1.3`.
- Update `eslint-plugin-react` from `7.32.2` to `7.33.2`.
- Update `prettier` from `2.8.4` to `3.2.4`. (see [full release notes](https://prettier.io/blog/2023/07/05/3.0.0.html))
- Update `typescript` from `4.9.5` to `5.3.3`. (see [full release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/))

## Breaking changes

Due to the upgrade of most of the dependencies, there are some breaking changes with new rules to implement. It is expected that any project using this config will have errors and warnings to fix. Unfortunately, is it difficult to know the exhaustive list of breaking changes. However, it is expected that most errors will be linked to TypeScript code. Here is a list of the most important ones:

- No invalid void type: `@typescript-eslint/no-invalid-void-type` is now enabled by default. This rule forbids the use of `void` as a type argument. This is a breaking change because it will now report errors on code that was previously valid.
- No useless constructor: `@typescript-eslint/no-useless-constructor` is now enabled by default. This rule forbids the use of a constructor when it is not necessary. This is a breaking change because it will now report errors on code that was previously valid.
- No dynamic delete: `@typescript-eslint/no-dynamic-delete` is now enabled by default. This rule forbids the use of the `delete` operator with computed key expressions. This is a breaking change because it will now report errors on code that was previously valid.

For Prettier, there will be new formatting errors, but they should all be fixable with the `--fix` option. They are most often related to ternary operators.

### In `package.json`

Start by updating the `eslint-config-cozy-app` package to the latest version.
You now can remove all the previous eslint dependencies from your project. They are now managed by the `eslint-config-cozy-app` package, unless you had a very specific configuration that is not covered by this package. Please refer to the list of dependencies above to know which versions are included in this package.

You also need to update your lint command. You should use the following one: `"lint": "eslint -c eslint.config.js ."`, where `eslint.config.js` is the file that contains the configuration of eslint. We will see how to create this file in the next section.

### Updating your eslint config to a flat-config

First, you need to delete your old eslint configuration file (`.eslintrc.js`, `.eslintrc.json`, or `.eslintrc.yml`), AND the ignore file (`.eslintignore`) if any.

Then, you need to create a new `eslint.config.js` file at the root of your project with the following content:

```ts
const eslintConfigCozyApp = require('eslint-config-cozy-app') // Import the config

module.exports = eslintConfigCozyApp // In this case we just export the config as is, but you can extend it if needed, please refer to the eslint documentation for more information
```

### Frictions you may encounter

- If one of your dependencies is importing an old version of eslint, you may have to force your project to use the latest version of eslint. You can do this by adding the following line to your `package.json`:

```json
"resolutions": {
  "**/eslint": "^8.0.0"
}
```

- If you have trouble adding customizations to the eslint configuration, you should refer to the [official documentation](https://eslint.org/docs/latest/use/configure/configuration-files) to know how to extend the configuration. The eslint configuration in flat-config mode is basically an array of objects, so you can push new rules to it, be it a plugin or a rule, or override existing ones, or include/exclude files.

- Your IDE should be able to handle this new config type, but on VSCode for instance, you have explicitly ask for it to be enabled in your settings.json :
`"eslint.experimental.useFlatConfig": true`

### Updating your code

You can now run `yarn lint --fix` to fix most of the issues. You may have to fix some issues manually, especially if you have a lot of TypeScript code.
In most cases, you shouldn't have too many errors. 

Example implementation: https://github.com/cozy/cozy-home/pull/2104/files


packages/cozy-device-helper/src/deeplink.ts
```diff
@@ -17,8 +17,8 @@ const _createHiddenIframe = (
 ): HTMLIFrameElement => {
   const iframe = document.createElement('iframe')
   iframe.src = uri
-  iframe.id = `hiddenIframe_${randomId}`
-  iframe.style.display = 'none'
+  iframe.id = `  hiddenIframe_${randomId}`
+  iframe.style.display = ' none'
```
Ldoppea, line 21: Seems like those added spaces are a mistake?

packages/cozy-device-helper/src/plugins.ts
```diff
@@ -18,10 +18,10 @@ export const hasSafariPlugin = (): Promise<boolean> => {
...
 /**
- * Check if the Cordova's cordova-plugin-network-information plugin is installed
- * @see https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-network-information/
+ * Cordova apps are deprecated, we should not use this plugin anymore
+ * Still, we need to expose this function to avoid breaking changes for now
  * @returns {boolean}
  */
 export const hasNetworkInformationPlugin = (): boolean => {
-  return isCordova() && window.navigator.connection !== undefined
+  return false
```
Ldoppea, line 26: This change is not part of the ESLint upgrade, would be great to inform about it in the commit message, or even better, to split this into another commit

packages/cozy-intent/src/view/contexts/NativeContext.ts
```diff
@@ -2,6 +2,8 @@ import React from 'react'
 
 import { NativeService } from '../../api'
 
+// React typings are out of date here
```
Ldoppea, line 5: Do we need to create a task to fix this later?

packages/eslint-config-cozy-app/index.js
cballevre, line 1: I think you should keep this file or change the import in your guide : `const eslintConfigCozyApp = require('eslint-config-cozy-app') // Import the config` because otherwise it should not work

cballevre: This comment was not consistent because my rlink does not work locally and doesn't have the new package.json which was pointing the main to `eslint.config.js`

cballevre (commented):
I think I found a blocking problem. Our webpack scripts use eslint to check the syntax as the configuration has changed and they are lost. We'll also need to migrate eslint to cozy-scripts before we can update the version smoothly. The related code : https://github.com/cozy/create-cozy-app/blob/a35dcf0bb0cc12b8892e73c5f62b9e3524bddea1/packages/cozy-scripts/config/webpack.config.eslint.js#L11

Capture d’écran 2024-04-08 à 16 09 14

acezard:
Closed (but not deleted) because stale

PR #2408, opened 2024-01-22
File Sharing provider

zatteo:


zatteo:
Implement does not support Android back button. New PR here https://github.com/cozy/cozy-libs/pull/2415.

PR #2414, opened 2024-01-25
[SCR-116] feat: Doctype debug cards integration

doubleface:
image

- feat: Pass the sourceAccountIdentifier to harvest cards
- feat: Add doctype debug cards for bills, files and identities
- feat: Add realtime to doctype debug cards

### DIFF
--- a/packages/cozy-harvest-lib/src/components/Datacards/index.jsx
+++ b/packages/cozy-harvest-lib/src/components/Datacards/index.jsx
@@ -1,5 +1,7 @@
 import React, { useMemo } from 'react'
 
+import { models } from 'cozy-client'
+
 import { useDatacardOptions } from './DatacardOptionsContext'
 
 const findSuitableDataCards = (datacardOptions, datacardContext) => {
@@ -24,6 +26,7 @@ const Datacards = ({ konnector, account, trigger }) => {
           konnector={konnector}
           trigger={trigger}
           accountId={trigger.message.account}
+          sourceAccountIdentifier={models.account.getAccountLogin(account)}
         />
       ))}
     </>
--- a/packages/cozy-harvest-lib/src/datacards/DoctypeDebugCard.jsx
+++ b/packages/cozy-harvest-lib/src/datacards/DoctypeDebugCard.jsx
@@ -1,8 +1,12 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 import ReactJsonPrint from 'react-json-print'
-
-import CozyClient, { Q, useQuery, hasQueryBeenLoaded } from 'cozy-client'
+import CozyClient, {
+  Q,
+  useQuery,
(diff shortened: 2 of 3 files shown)


packages/cozy-harvest-lib/src/components/Datacards/index.jsx
```diff
@@ -24,6 +26,7 @@ const Datacards = ({ konnector, account, trigger }) => {
           konnector={konnector}
           trigger={trigger}
           accountId={trigger.message.account}
+          sourceAccountIdentifier={models.account.getAccountLogin(account)}
```
Crash--, line 29: why don't we use `getIdentifier` from `models.manifest`? In order to have the full implementation? Instead of relying of something that is not future proof?

Crash--: Also I'm wondering if you should not remove this `sourceAccountIdentifier` and just pass the `account`. Also if we do that, we can remove the `accountId` props. 

I think it's way better to pass the full object instead of a subset. The subset is never future proof.

doubleface: getIdentifier only returns which field in the manifest is the identifier, it does not return the value of the identifier field, which is the sourceAccountIdentifier only for node konnectors.

But actually, I see that getAccountLogin (which is used by getAccountName as a second possibility of value) does not use getIdentifier and I will fix that.

For the second message on passing the account, that was actually my first guess but I thought each card would have to guess how to find the sourceAccountIdentifier and I found it safer to find the sourceAccountIdentifier in only one place. But since every card is defined in only one place, I suppose we can also do it there.

doubleface: I think also getIdentifier should be renamed to getIdentifierField to avoid the confusion

doubleface: https://github.com/cozy/cozy-client/pull/1439

doubleface:
@Crash-- Should be Ok now

doubleface:
@Merkur39 @JF-Cozy @zatteo can I have a review of this one please :pray:

packages/cozy-harvest-lib/src/components/Datacards/index.jsx
Crash--: can you update the peerDep and the devDep of cozy-client in order to get  https://github.com/cozy/cozy-client/pull/1439 ?

Crash-- (approved):

packages/cozy-harvest-lib/src/components/Datacards/index.jsx
Crash--: @doubleface Too bad to have merged it without the upgrade of cozy-client 🤔

PR #2440, opened 2024-02-16
[SCR-114] fix: Get sourceAccountIdentifier with getAccountName

doubleface:
This way, we get the sourceAccountIdentifier for both following cases :
 - ccc accounts which all have `auth.accountName`
 - node konnectors accounts which follow the getAccountLogin logic since
 getAccountName defaults to getAccountLogin value.

### DIFF
--- a/packages/cozy-harvest-lib/src/components/Datacards/index.jsx
+++ b/packages/cozy-harvest-lib/src/components/Datacards/index.jsx
@@ -26,7 +26,7 @@ const Datacards = ({ konnector, account, trigger }) => {
           konnector={konnector}
           trigger={trigger}
           accountId={trigger.message.account}
-          sourceAccountIdentifier={models.account.getAccountLogin(account)}
+          sourceAccountIdentifier={models.account.getAccountName(account)}
         />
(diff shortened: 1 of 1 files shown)


packages/cozy-harvest-lib/src/components/Datacards/index.jsx
```diff
@@ -26,7 +26,7 @@ const Datacards = ({ konnector, account, trigger }) => {
           konnector={konnector}
           trigger={trigger}
           accountId={trigger.message.account}
-          sourceAccountIdentifier={models.account.getAccountLogin(account)}
+          sourceAccountIdentifier={models.account.getAccountName(account)}
```
Crash--, line 29: nit: Do we need a specific version of cozy-client to have this behavior? Is it the right one in dev & peerDeps ?

doubleface: Yes, the 35.1.0 version of cozy-client is needed to have this behavior, we are in 45.8.1 in deps and peer deps.

PR #2448, opened 2024-02-23
feat: Log every error from native side

zatteo:
We log every error from native side to avoid being in the dark when there is no catch in both native and webview sides.

We use a warn log to trigger a Sentry upload on native side.

On native side : ` WARN  cozy-intent Error during intent to http://photos.zatteo.10-0-2-2.nip.io:8080/ :  {"name":"Error","message":"ABCD"}`

What do you think of this implementation ?

### DIFF
--- a/packages/cozy-intent/src/api/services/NativeMessenger.ts
+++ b/packages/cozy-intent/src/api/services/NativeMessenger.ts
@@ -1,5 +1,7 @@
 import { Messenger, debug } from 'post-me'
 
+import Minilog from 'cozy-minilog'
+
 import {
   ListenerRemover,
   MessageListener,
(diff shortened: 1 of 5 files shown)


packages/cozy-intent/src/api/services/NativeMessenger.ts
```diff
@@ -8,17 +10,29 @@ import {
...
+  private webviewUri?: string
 
   constructor(webviewRef: WebviewRef) {
     this.injectJavaScript = webviewRef.injectJavaScript
+    this.webviewUri =
+      webviewRef.props.source.uri ?? webviewRef.props.source.baseUrl
   }
 
   public postMessage = (message: Record<string, unknown>): void => {
+    if (message.error) {
```
acezard, line 29: Personally I think the error wrap should be closer to the actual method call. But if this implementation suits your purpose that's fine for me

acezard (approved):

PR #2453, opened 2024-02-28
fix: Remove scenario where sharingModal lose state (VO-301)

acezard:
This would trigger a cascading rerender scenario,
creating a side effect where the SharingModal would end up thinking
it has no sharedChild, rendering the wrong view
and allowing mistakingly the user to share the parent of a shared dir
We use the document name as a fallback to avoid this scenario


packages/cozy-sharing/src/helpers/files.js
```diff
@@ -8,7 +8,11 @@ export const fetchFilesPaths = async (client, doctype, files) => {
   const filePaths = files
     .map(f => parentDirs.data.find(d => d.id === f.dir_id))
     .filter(parentDir => parentDir !== undefined)
-    .map(parentDir => {
+    .map((parentDir, index) => {
+      // If the file is a directory, we return its path directly
+      if (files[index].type === 'directory') {
```
cballevre, line 13: I'm not sure how well it works with an index, because since we're doing a filter before the size of the files array and the one we're mapping here is no longer the same 🤔

acezard: > I'm not sure how well it works with an index, because since we're doing a filter before the size of the files array and the one we're mapping here is no longer the same 🤔

that's true, should take the index before any filtering. Though I'm unsure in which scenarios the files array is longer than 1, I couldn't find such a case

cballevre: I think it's possible to have several here: https://github.com/cozy/cozy-libs/blob/e3e6b6ca27bfcc2c61893eeb763c43f926308630/packages/cozy-sharing/src/SharingProvider.jsx#L203 

You could do a unit test to check that this case works and adapt the function to handle it

acezard: Updated with more streamlined logic. If any directory in input (should always be one according to code, but never sure), just ignore the file specific logic and return the directory path for each directory. added tests

packages/cozy-sharing/src/components/useFetchDocumentPath.jsx
cballevre, line 1: nit: we could simplify the `useState` to `null` only because make `setDocumentPath` directly after into the folder

cballevre (approved):
The PR looks good to me 👍 

Sorry you need to rebase because I've moved all the hooks together in my PR for group handling

PR #2471, opened 2024-03-08
chore: Minimize services when cozy-notifications is used

zatteo:
cozy-notifications uses mjml to generate mail templates. mjml can not be minimized as is. So previously, we disabled minimize for the whole services configuration. It could lead to huge services bundle but it was working.

We discovered that configuring Terser plugin with keep_fnames to true is enough to allow minimization without breaking mjml. So let's use it to reduce services bundle size.

From the Terser plugin documentation :
keep_fnames (default false) -- Pass true to not mangle function names. Pass a regular expression to only keep function names matching that regex. Useful for code relying on Function.prototype.name. See also: the keep_fnames compress option.

### DIFF
--- a/packages/cozy-notifications/src/webpack/config.js
+++ b/packages/cozy-notifications/src/webpack/config.js
@@ -1,5 +1,7 @@
-import webpack from 'webpack'
 import path from 'path'
+
+import TerserPlugin from 'terser-webpack-plugin'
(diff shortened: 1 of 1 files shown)


Crash--:
(this is not a chore, with this change, you're changing the released code. You also need a new version to be published, so let's call it a feat or a fix ;))

PR #2487, opened 2024-03-20
[VO-434] feat: Show detail of recipient groups

cballevre:
Capture d’écran 2024-03-20 à 17 06 39

### DIFF
--- a/packages/cozy-sharing/src/components/ConfirmTrustedRecipientsDialog.jsx
+++ b/packages/cozy-sharing/src/components/ConfirmTrustedRecipientsDialog.jsx
@@ -1,6 +1,9 @@
+import cx from 'classnames'
 import React from 'react'
 
+import List from 'cozy-ui/transpiled/react/List'
 import Typography from 'cozy-ui/transpiled/react/Typography'
+import useBreakpoints from 'cozy-ui/transpiled/react/providers/Breakpoints'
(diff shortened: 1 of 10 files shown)


packages/cozy-sharing/src/components/Recipient/GroupRecipientDetailWithAccess.jsx
```diff
@@ -0,0 +1,57 @@
...
+    <div>
+      <Typography variant="subtitle2">
+        {t('GroupRecipientDetail.withAccess.title')}
+      </Typography>
+      <List>
+        {withAccess.map((recipient, index) => {
+          const name = getDisplayName(recipient)
+
+          return (
+            <ListItem disableGutters key={index}>
```
Ldoppea, line 28: Same comment as in previous PR, here we should use the recipient's id (the one you created yesterday) instead of the `.map` index.

As a more global rule, we should try to never use `.map` indexes as React keys.

Ldoppea: Same comment for the `WithoutAccess` component

cballevre: I've corrected the problem, thank you for catching my blunder

Ldoppea (approved):

PR #2490, opened 2024-03-21
fix(mespapiers): Use currentStep from props instead hook

Merkur39:
Depending on whether we are on the creation or editing component, the current step is not retrieved in the same way.


Merkur39:
Je merge ce fix, attendu pour une MEP

PR #2492, opened 2024-03-22
feat: Add useIsAvailable hook

zatteo:
This pattern is used very often in cozy apps so we add it here as a hook.

Usage : 
```javascript
import React from 'react'
import { useIsAvailable } from 'cozy-intent'

const MyComponent = () => {
  const { isAvailable } = useIsAvailable('print')

  if (isAvailable) {
    return (<div>You can print</div>)
  } else if (isAvailable === null) {
    return (<div>You do not know if you can print</div>)
  } else {
    return (<div>You can not print</div>)
  }
}

```

Todo :
- [x] check on a cozy app


packages/cozy-intent/src/view/hooks/useIsAvailable.ts
```diff
@@ -0,0 +1,22 @@
+import { useState, useEffect } from 'react'
+
+import { useWebviewIntent } from './useWebviewIntent'
+
+export const useIsAvailable = (methodName: string): boolean => {
+  const webviewIntent = useWebviewIntent()
+  const [isAvailable, setIsAvailable] = useState(false)
```
zatteo, line 7: Do you thinl we should return undefined or null when the webviewIntent.call has not been made ?

JF-Cozy: si on veut tester ce cas, un simple `!isAvailable` ne suffira pas il faudra être explicite avec un `isAvailable === something`. Et c'est plus court je pense de tester avec `null` que `undefined` :p

zatteo: Ça retourne null maintenant par défaut : 
- permet de tester si on veut ce cas
- toujours possible de faire un simple if(isAvailable)

packages/cozy-intent/src/view/hooks/useIsAvailable.ts
```diff
@@ -0,0 +1,22 @@
...
+import { useWebviewIntent } from './useWebviewIntent'
+
+export const useIsAvailable = (methodName: string): boolean | null => {
+  const webviewIntent = useWebviewIntent()
+  const [isAvailable, setIsAvailable] = useState<boolean | null>(null)
+
+  useEffect(() => {
+    const checkIsAvailable = async (): Promise<void> => {
+      if (webviewIntent) {
+        const result = await webviewIntent.call('isAvailable', methodName)
```
acezard, line 12: what if an error is thrown during the async call here? I guess it could be useful to distinguish on the web app side if the feature is not available or if we don't know because something went wrong

zatteo: Interesting use case. It can be the case on old flagship app.

I pushed a new version. What do you think about it ?

acezard: > Interesting use case. It can be the case on old flagship app.
> 
> I pushed a new version that returns `{ isAvailable, error }`. What do you think about it?

I think that's fine. Just a nitpick: 

```ts
interface UseIsAvailableResult {
  isAvailable: boolean | null
  error: string | undefined
}
```
we can just use typescript optional interface property syntax `error?`
Also I would set the useState generics with a bit more verbosity but with no need of having an init value in error and also more linked to the interface which is always good for eventual future changes. eg:
```ts
interface UseIsAvailableResult {
  isAvailable: boolean | null
  error?: string
}

export const useIsAvailable = (methodName: string): UseIsAvailableResult => {
  const webviewIntent = useWebviewIntent()
  const [isAvailable, setIsAvailable] = useState<UseIsAvailableResult["isAvailable"]>(null)
  const [error, setError] = useState<UseIsAvailableResult["error"]>() // undefined is implicit here

  useEffect(() => {
    if (!webviewIntent) return // return early so we have lest nesting depth

    const checkIsAvailable = async (): Promise<void> => {
      try {
        const result = await webviewIntent.call('isAvailable', methodName)

        if (typeof result !== 'boolean') // We don't care about non bool result so just throw
          throw new Error(`Invalid result from isAvailable method: ${result}`)

        setError(undefined)
        setIsAvailable(result)
      } catch (error) {
        setIsAvailable(false)
        setError(getErrorMessage(error)) // using getErrorMessage() helper for inference
      }
    }

    void checkIsAvailable()
  }, [webviewIntent, methodName])

  return { isAvailable, error }
}
```

acezard (approved):

PR #2495, opened 2024-03-26
fix: We should request the files for the right slug and account

Crash--:
Since https://github.com/cozy/cozy-libs/pull/2447/files we changed the way to fetch the files related to a konnector by using the sourceAccountIdentifier.

But this sourceAccountIdentifier can be the same for other konnectors, so if we had several konnectors with the same sourceAccountIdentifier (since it's generaly your main mail address...) we'll display the wrong files.

We should also filter by konnector slug.


cballevre (approved):
LGTM 👍

PR #2501, opened 2024-03-27
fix: Use sourceAccountIdentifier and konnector slug to show viewer modal

cballevre:
Parameter of useDataCardFiles change but we didn't update ViewerModal properly. I chose to search this param with a query to io.cozy.accounts as we had already this information without changing route path.  This is a temporary fix and I think we need to refactor the viewer part to have an abstraction to account in this modal

### DIFF
--- a/packages/cozy-harvest-lib/src/helpers/queries.js
+++ b/packages/cozy-harvest-lib/src/helpers/queries.js
@@ -38,6 +38,24 @@ export const buildAccountQuery = ({ slug, sourceAccountIdentifier }) => ({
     fetchPolicy: defaultFetchPolicy
   }
 })
+
+/**
+ *  Build an account query for the given konnector.
+ * ("getById" throws an error even if the query is not enabled)
+ * @param {string} accountId - io.cozy.accounts document's id
+ * @returns {object} - a query spec
+ */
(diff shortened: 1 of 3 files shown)


Crash-- (approved):

doubleface (approved):

packages/cozy-harvest-lib/src/helpers/queries.js
```diff
@@ -38,6 +38,24 @@ export const buildAccountQuery = ({ slug, sourceAccountIdentifier }) => ({
     fetchPolicy: defaultFetchPolicy
   }
 })
+
+/**
+ *  Build an account query for the given konnector.
+ * ("getById" throws an error even if the query is not enabled)
+ * @param {string} accountId - io.cozy.accounts document's id
+ * @returns {object} - a query spec
```
Crash--, line 46: Si je devais vraiment pinailler, tu dois pouvoir importer le type de cozy-client le IoCozyAccount ;)

cballevre: Mince j'ai cliqué trop vite dès que j'ai vu la CI était passé

PR #2502, opened 2024-03-29
[VO-592] fix: Correct some of the feedback on dynamic group sharing

cballevre:


packages/cozy-sharing/src/queries/queries.js
```diff
@@ -59,10 +59,52 @@ export const buildReachableContactsQuery = () => ({
...
+  definition: Q(Group.doctype).getByIds(ids),
   options: {
-    as: 'io.cozy.contacts.groups'
+    as: `io.cozy.contacts.groups/by-ids/${ids.join('/')}`,
+    fetchPolicy: defaultFetchPolicy,
+    enabled: ids.length > 0
+  }
+})
+
+export const buildUnreachableContactsFromGroupsQuery = groupIds => ({
```
cballevre, line 71: @paultranvan I'd like your opinion on this query. I've used `$elemMatch` which works well locally but I'm wondering about performance. For a bit more context, we want to retrieve all the contact groups with contactable members as well as their non-contactable members.

I've opted for a two-stage approach: 
1. I retrieve all the contacts that can be contacted. This request allows me to retrieve their group ids and is used in another part of the interface.
2. I retrieve the groups via their ids as well as their non-contactable contacts via this query. 

I had considered another option, which was to retrieve all the contacts and all the groups in order to do the same processing on the front end, but the request can be quite large, so I tried to keep the initial optimisations.

cballevre:
@acezard or @Ldoppea if you have a little time to spend on the review would be appreciated 🙏

packages/cozy-sharing/src/helpers/successMessage.js
```diff
@@ -1,16 +1,39 @@
...
-  return newRecipients.filter(contact => {
-    const email = models.contact.getPrimaryEmail(contact)
-    const cozyUrl = models.contact.getPrimaryCozy(contact)
-    return !currentRecipients.find(
-      r =>
-        (email && r.email && r.email === email) ||
-        (cozyUrl && r.instance && r.instance === cozyUrl)
-    )
-  }).length
+  return newRecipients.reduce(
```
Ldoppea, line 6: nit: If we assume that a sharing always have a few recipients, maybe we can sacrifice a bit of performance and improve readability by using two `filter` loops instead of a single `reduce`?

packages/cozy-sharing/src/components/ShareModal/EditableSharingModal.jsx
```diff
@@ -47,8 +31,6 @@ export const EditableSharingModal = ({ document, ...rest }) => {
 
   return (
     <DumbShareModal
```
Ldoppea, line 33: nit: As you moved the queries inside of child component, I suppose we cannot call this a `dumb` component anymore 😅

packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
```diff
@@ -1,73 +1,147 @@
...
-      contacts.fetchStatus === 'loaded' &&
-      !contactGroups.hasMore &&
-      contactGroups.fetchStatus === 'loaded'
+      !reachableContactsResult.hasMore &&
+      reachableContactsResult.fetchStatus === 'loaded' &&
+      !contactGroupsByIdsResult.hasMore &&
+      contactGroupsByIdsResult.fetchStatus === 'loaded' &&
+      (flag('sharing.show-recipient-groups') !== true ||
+        (!unreachableContactsFromGroupsResult.hasMore &&
+          unreachableContactsFromGroupsResult.fetchStatus === 'loaded'))
```
Ldoppea, line 76: nit: this `if` starts to be hard to read, maybe we can split it into multiple named variables? (except if the performances are important but seems like it would have only few impact)

Ldoppea: also can't we use `isQueryLoading` here like in the following block?

packages/cozy-sharing/src/helpers/contacts.js
```diff
@@ -39,3 +39,17 @@ export const getOrCreateFromArray = async (client, contacts, createContact) => {
...
+    if (
+      (contact.relationships?.groups?.data || [])
+        .map(group => group._id)
+        .includes(groupId)
+    ) {
+      return [...contactsfromGroup, contact]
+    }
+
+    return contactsfromGroup
+  }, [])
```
Ldoppea, line 54: Here again I assume there are only a few contacts on sharings so the impact should be low, but did you check perfs on this?

As this method is called from ShareRecipientsInput, my understanding is that we have a `.map` inside of a `.reduce` inside of a `.map`, so execution time can grow quickly

Ldoppea (approved):

packages/cozy-sharing/src/queries/queries.js
cballevre: We discuss together and we decide to search all unreachable that have a group to avoir having dynamic variable in a partial index

packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
cballevre: I found a better way to make it readable. I've separated the responsibilities. This component will only be responsible for knowing if there is data loading. And the component below will be responsible for displaying the loading status only if the user has focused the input. This allows me to remove this useEffect from the component 🙂

packages/cozy-sharing/src/helpers/contacts.js
cballevre: Unfortunately your assumption is not correct. It corresponds to all the contacts of a user 🙁. I'm going to check the performance of this algo and we have a complexity of  O(n^2 * m^2) so the performance is going to be problematic. I'm going to try to get a bigger set of data locally. Its purpose is just to provide additional information, so I'm going to try to migrate it so that it doesn't block the user's view

cballevre: > Don't you mean all the reachable contacts, i.e. with email or url ?

Also unreachable contacts with a group as I do the filter twice into the ShareRecipientsInput

> For the getContactsFromGroupId, isn't it O(n * m)

You a right, the complexity of `getContactsFromGroupId` is not O(n^2 * m^2). My response was missing the context I wanted to call about the map on group inside `ShareRecipientsInput` which was using this function. 

> And arguably, q is supposed to be rather small compared to n

Totally, I carried out tests with 800 contacts divided into 20 groups corresponding to classes with half of the contact unreachable. I have a Mac M1 and with the simulation of a single core slowed down to * 4 we rarely exceed 50ms, so it's probably acceptable

packages/cozy-sharing/src/helpers/successMessage.js
cballevre: Yes good point I split the filter step into its own before the reduce 👍

PR #2511, opened 2024-04-10
feat: Allow Document.bulkSave to force documents update (SCR-444)

doubleface:
Even if Document.checkAttributes are the same.
We can do this with:

```javascript
Document.bulkSave({...}, {
  forceUpdate: true
})
```

This is needed in special cases in banking konnectors, to also update
categorized transactions even if they already have been saved once.


packages/cozy-doctypes/src/Document.spec.js
```diff
@@ -517,6 +517,37 @@ describe('Document used with CozyClient', () => {
         Simpson.addCozyMetadata({ name: 'Marge', son: 'Bart' })
       )
     })
+
+    it('should not update the document if important attributes are the same', async () => {
```
Crash--, line 521:
```suggestion
    it('should not update the document if checked attributes are the same', async () => {
```

packages/cozy-doctypes/src/Document.spec.js
```diff
@@ -517,6 +517,37 @@ describe('Document used with CozyClient', () => {
...
+        .spyOn(SpecialSimpson, 'queryAll')
+        .mockReturnValueOnce([{ name: 'Marge' }])
+      jest.spyOn(cozyClient, 'save').mockReturnValueOnce({ data: {} })
+
+      await SpecialSimpson.createOrUpdate({ name: 'Marge', son: 'Bart' })
+
+      expect(cozyClient.save).not.toHaveBeenCalled()
+    })
+
+    it('should update the document if important attributes are the same but forceUpdate option is true', async () => {
```
Crash--, line 534:
```suggestion
    it('should update the document if checked attributes are the same but forceUpdate option is true', async () => {
```

packages/cozy-doctypes/src/Document.js
```diff
@@ -201,20 +201,28 @@ class Document {
...
+   * or is not forced with forceUpdate option, the original
    * document is returned. Otherwise, the update document is
    * returned with metadata updated.
    *
+   * @param {object} doc - The document already existing in db
+   * @param {object} update - The update to apply to the document
+   * @param {object} options - Options object
+   * @param {boolean} options.forceUpdate - Should the method force the update even if important attributes are identical
+   * @returns {object} - The updated document with cozy new metadata when an update has been done
+   *
```
paultranvan, line 214: Thanks for the jsdoc!
I would just have changed 2 things:
- Describe a bit more what an "important" attributes is, and what it means. From an external reader, "important" is bit vague. Is "deduplication attributes" relevant, for instance?
- I feel like this jsdoc should especially be on the `bulkSave` method, as this is the actual API called by the app/konnector

packages/cozy-doctypes/src/Document.js
```diff
@@ -360,7 +368,14 @@ class Document {
   }
 
   /**
-   * Save many documents concurrently
+   * Save many documents concurrently using the createOrUpdate method
+   *
+   * @param {Array<object>} documents - The document already existing in db
```
paultranvan, line 373: Why "already existing"? Can't you bulk save non-existing documents?

paultranvan (approved):

PR #2548, opened 2024-06-04
feat: Add options object in NativeService args

zatteo:
When receiving an intent on the flagship app, we want to know which webview sent the intent. Here we inject at the beginning of the `args` array an `options` object with the `webviewUri`.

*Example*

On a front end app, we will do

`webviewIntent.call('multiply', 2, 3)`

On flagship app, we will set a local method

`multiply: (options, arg1, arg2) => return arg1 * arg2`

=> the options object has been added at the beginning of the `args` array

We also add a new NativeMethodsRegisterWithOptions to modelize this new behavior.

BREAKING: A options object has been added in NativeService message args. You need to update your local methods (see example). FLAGSHIP APP ONLY. Nothing to do on a front end app.

### DIFF
--- a/packages/cozy-intent/src/api/models/methods.ts
+++ b/packages/cozy-intent/src/api/models/methods.ts
@@ -1,4 +1,4 @@
-import { AppManifest, FlagshipUI } from '../../api'
+import { AppManifest, FlagshipUI, PostMeMessageOptions } from '../../api'
 
 type PostMeDefault = Record<string, (...args: unknown[]) => Promise<null>>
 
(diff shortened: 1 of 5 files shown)


packages/cozy-intent/src/api/models/methods.ts
```diff
@@ -13,10 +13,8 @@ interface _NativeMethodsRegister {
     app: AppManifest,
     iconParams?: DOMRect
   ) => Promise<null>
-  openSettingBiometry: () => Promise<boolean>
   setDefaultRedirection: (defaultRedirection: string) => Promise<null>
   setFlagshipUI: (flagshipUI: FlagshipUI, caller?: string) => Promise<null>
-  showSplashScreen: () => Promise<null>
```
zatteo, line 19: Unused methods.

Ldoppea (approved):

PR #2598, opened 2024-10-30
feat: Download file on mobile viewer on press

zatteo:
We want to download file on mobile viewer on press (so when we put the finger on the screen) and not on tap (so when we put the finger on the screen and we release it).

I added a throttle to avoid download multiple time the same file because we can no use anymore 'tapCount' with press which was used to avoid download 3 times the file if we tapped 3 times.

I also moved to useMemo instead of useCallback to support throttle (see https://github.com/facebook/react/issues/19240).

### DIFF
--- a/packages/cozy-viewer/src/ViewersByFile/PdfMobileViewer.jsx
+++ b/packages/cozy-viewer/src/ViewersByFile/PdfMobileViewer.jsx
@@ -1,5 +1,6 @@
+import throttle from 'lodash/throttle'
 import PropTypes from 'prop-types'
-import React, { useState, useEffect, useRef, useCallback } from 'react'
+import React, { useState, useEffect, useRef, useMemo } from 'react'
 
 import { useClient } from 'cozy-client'
 import { downloadFile } from 'cozy-client/dist/models/file'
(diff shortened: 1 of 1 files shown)


packages/cozy-viewer/src/ViewersByFile/PdfMobileViewer.jsx
```diff
@@ -32,36 +33,42 @@ export const PdfMobileViewer = ({ file, url, t, gestures }) => {
...
+              message: t('Viewer.error.generic'),
+              severity: 'error',
+              variant: 'filled',
+              icon: false
+            })
+          }
+        },
+        1000,
+        { trailing: false }
+      ),
```
Merkur39, line 53: info: This approach is also functional
https://github.com/cozy/cozy-contacts/blob/26b30f5160ff6e7543e7f2e134bd1893f701c99b/src/components/Modals/ContactInfoTitle.jsx#L62-L75

Merkur39 (approved):

packages/cozy-viewer/src/ViewersByFile/PdfMobileViewer.jsx
zatteo: Interesting thanks 👍

PR #2621, opened 2024-11-19
fix: Handle shared notes

paultranvan:
When a note is coming from a Cozy sharing, a special URL must be built, including the sharing instance URL, a sharecode and a the note ID on this instance.
The client does not have these information, so the stack is required to build the opening URL. This is directly handled by Drive, so when the user wants to open a shared note, we first need to redirect on Drive, that will redirect on Notes.

See https://github.com/cozy/cozy-drive/pull/3244/files


cballevre (approved):
LGTM 👍 We'll just have to bear in mind that version 1.74.2 has to be in production to be able to use it in production

PR #2623, opened 2024-11-20
[DAT-70] feat: Do not use flexsearch store

paultranvan:
Flexsearch has a built-in store capability, in order to store the full content of indexed documents.
This is useful, however, this takes extra memory for nothing, as we already have the cozy-client store. All the indexed documents for the search are supposed to exist in cozy-client store, as we query them from Pouch at init, and retrieve updates from realtime, already plugged in cozy-client.

### DIFF
--- a/packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
+++ b/packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
@@ -22,7 +22,7 @@ import {
   shouldKeepFile
 } from './helpers/normalizeFile'
 import { normalizeSearchResult } from './helpers/normalizeSearchResult'
-import { queryAllDocs, queryFilesForSearch } from './queries'
+import { queryAllDocs, queryFilesForSearch, queryDocsByIds } from './queries'
 import {
   CozyDoc,
   RawSearchResult,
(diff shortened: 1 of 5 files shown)


packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
```diff
@@ -359,8 +360,25 @@ export class SearchEngine {
...
+              ]
+          },
+          {
+              "field": "fullname",
+              "result": [
+                  "604627c6bafee013ec5f27f7f72029f6", "604627c6bafee013ec5f27f3f714568"
+              ]
+          }
+        ]
+      */
```
Ldoppea, line 381: Was that comment expected to be commited?

packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
```diff
@@ -179,14 +180,13 @@ export class SearchEngine {
     const fieldsToIndex = SEARCH_SCHEMA[doctype]
 
     const flexsearchIndex = new FlexSearch.Document<CozyDoc, true>({
-      tokenize: 'forward',
+      tokenize: 'full',
```
Ldoppea, line 183: nit: this change is not documented in the commit message and is not part of the "do not use flexsearch store" scope. Maybe you can add some precision in the commit message or extract it in another commit?

packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
paultranvan: Yes it was, it does not look like a good practice, but I honestly felt like it was helpful to better grasp the expected result quickly

packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
paultranvan: Oh my, it's a mistake, thank you!

PR #2627, opened 2024-11-25
[VO-1184] refactor(cozy-harvest-libs): Only use react-route-dom v6

cballevre:
This PR aims to simplify Harvest now that all our apps have been migrated to `react-router-dom >= 6.0.0`. We're using relative routes as much as possible to be agnostic to the execution context.

**App tested :**
- `cozy-drive` with the banner
- `cozy-home` 
- `cozy-store` with the intent of `cozy-home`

The PR can be read in two parts. The first part goes up to commit e65b64fcae9c7f1f586dc5648d833369cb7e2c42, which includes the removal of `react-router 3.0.0` and all the adaptations. In the second part, I replace the adaptations with the `useNavigate` hook in all the components


Merkur39 (approved):
LGTM 👏

PR #2642, opened 2024-12-05
feat(cozy-sharing): Add isPublic prop to SharingProvider

Merkur39:
In public mode, execution of the promise with calls to the following methods: `this.sharingCol.findByDoctype(doctype)`,
`this.permissionCol.findLinksByDoctype(doctype)` and `client.fetchQueryAndGetFromState(fetchApps())` failed (cf 403).

So we have useless calls and the `fetchAllSharings` method, which in the end does nothing.

The `isPublic` prop quickly exits `fetchAllSharings` to avoid 3 useless requests.

Real time is retained as functional and useful. (link sharing and Cozy to Cozy added but not sync)


zatteo:
Nice 👍

zatteo (approved):

PR #2649, opened 2024-12-10
feat: Remove cozy-scanner

zatteo:
This library was used for qualifications that are now in cozy-client. It is not used anymore.


Merkur39 (approved):
👍

PR #2653, opened 2024-12-11
feat: Add NativeFileSharingProvider and shareNative action

zatteo:
Add a provider and a action to expose an ActionMenu action that can share files with mobile device share capabilities.

shareFilesNative from the NativeFileSharingProvider must be passed to the action.

Inspired from code in mespapiers. It will be used in cozy-drive, cozy-notes and mespapiers.


Merkur39 (approved):
Looks good 👍

PR #2669, opened 2024-12-19
Export multiple path in dataproxy lib and minor fixes

zatteo:


Ldoppea (approved):
Suggestion: you can also refactor the main `index.js` with something like
```
export * from './ui'
export * from './api'
```
This would prevent future mistakes so developers won't be tempted to add new imports to it and forget to add them in the `/api` or `/ui` indexes.

packages/cozy-dataproxy-lib/src/api.ts
```diff
@@ -0,0 +1,2 @@
+export { SearchEngine } from './search/SearchEngine'
+export { DataProxyProvider, useDataProxy } from './dataproxy/DataProxyProvider'
```
paultranvan, line 2: Are `DataProxyProvider` and `useDataProxy` necessary here? I think they are currently only used inside the ui search

zatteo: I can move them if they are more linked to the ui part.

paultranvan: I'm not sure, as I guess the provider can be seen as an api, more than a ui concept. Let's keep it like this, but I think you can remove the `useDataProxy` export

zatteo: Agree, even it will be used by ui components, it's more an api concept 👍

paultranvan (approved):
Thanks a lot!

PR #2685, opened 2024-12-24
feat: Enable search options

paultranvan:
It is now possible to pass a `doctypes` options to the search, that serves two purposes:
- Specify which doctypes should be searched
- Specify the doctype  order of the results


paultranvan:
Since it's holidays for pretty much everybody, let's merge it without waiting next year :wink:

PR #2698, opened 2025-01-13
fix: Disable `fastRefresh` in shared Rsbuild configuration

Ldoppea:
In our projects, we found that using
`rsbuild build --watch --mode development` instead of just `rsbuild build --watch` would speed up the re-build process

But when doing so, the bundler enables the
[fastRefresh](https://rsbuild.dev/plugins/list/plugin-react#fastrefresh) mode from plugin-react

For now, this mode does not work and trying to run the cozy-apps would produce a blank screen with some `$RefreshSig$ is not defined`. This may be due to our configuration where the cozy-apps are served by the cozy-stack instead of the Rsbuild dev server

Until we find how to make it work, we want to disable this feature by default

It is still possible to enable it by setting the `enableFastRefresh` parameter to `true` when calling `getRsbuildConfig()`


zatteo (approved):

Ldoppea:
Closing this in favor of #2700

PR #2700, opened 2025-01-14
feat: Enable HMR on shared Rsbuild configuration

Ldoppea:
By default Hot Module Rreplacement is configured to work with a localhost dev server

This server is responsible to serve the app and enable HMR

This wouldn't work with our apps because they are meant to be served by a local cozy-stack as a dev environment and so the cozy-app would try to connect to the HMR websocket using the cozy-stack's URL

Also the Rsbuild dev server is configured to serve built files from memory, so the `/build` folder would be empty and the cozy-stack wouldn't see any app

To fix this we can use both `dev.writeToDisk` and `dev.client` configurations

Then in order to benefit from this, the cozy-app should be run using `rsbuild dev` instead of `rsbuild build --watch`

More info:
- https://github.com/web-infra-dev/rsbuild/discussions/4366
- https://rsbuild.dev/config/dev/write-to-disk#writing-to-disk
- https://rsbuild.dev/config/dev/client

### DIFF
--- a/config/rsbuild-config-cozy-app/getRsbuildConfig.js
+++ b/config/rsbuild-config-cozy-app/getRsbuildConfig.js
@@ -31,7 +31,9 @@ function getRsbuildConfig({
     plugins: [
       pluginEjs(),
       pluginNodePolyfill(),
-      pluginReact(),
+      pluginReact({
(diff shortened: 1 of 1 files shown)


zatteo (approved):
🎉

PR #2711, opened 2025-01-23
fix: Cozy-pouch-link as peerDep to avoid version conflict

paultranvan:
BREAKING CHANGE: it is now required to have cozy-pouch-link as a dependency for a client to benefit from the local search functionalities.

This was typically always false: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-dataproxy-lib/src/search/SearchEngine.ts#L56 in the flagship app and was forcing online search even though local data was available. As a side effect, the offline couldn't work.

### DIFF
--- a/packages/cozy-dataproxy-lib/package.json
+++ b/packages/cozy-dataproxy-lib/package.json
@@ -12,7 +12,6 @@
   "dependencies": {
     "classnames": "2.5.1",
     "comlink": "4.4.1",
-    "cozy-pouch-link": "^50.3.1",
     "flexsearch": "0.7.43",
     "lodash": "4.17.21",
     "mime-types": "2.1.35",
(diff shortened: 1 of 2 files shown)


packages/cozy-dataproxy-lib/package.json
```diff
@@ -68,6 +67,7 @@
     "cozy-intent": ">=2.26.0",
     "cozy-logger": ">=1.10.4",
     "cozy-minilog": ">=3.3.1",
+    "cozy-pouch-link": "50.3.1",
```
Ldoppea, line 70: Declaring it in peerDeps is not enough, you should also declare it in devDeps

paultranvan: It is already declared in devDep: https://github.com/cozy/cozy-libs/blob/623a5c6aee86cba03d652950d22870fc9237156d/packages/cozy-dataproxy-lib/package.json#L35C6-L35C21

Ldoppea: My bad 😅

Ldoppea (approved):

PR #2718, opened 2025-01-29
Fixes on indexing

paultranvan:
- [fix: Index documents only after login](https://github.com/cozy/cozy-libs/commit/965c61d186ea8d2a98f483efa3b17265b8488b1d)
- [fix: Improve indexing time](https://github.com/cozy/cozy-libs/commit/08e0808219bf74e6d78a94e0fa4f1aa84bc59635)

### DIFF
--- a/packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
+++ b/packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
@@ -54,7 +54,7 @@ export class SearchEngine {
     this.searchIndexes = {} as SearchIndexes
 
     this.isLocalSearch = !!getPouchLink(this.client)
-    log.info('Use local data on trusted device : ', this.isLocalSearch)
+    log.info('Use local data on trusted device: ', this.isLocalSearch)
 
     this.debouncedReplication = (): void => {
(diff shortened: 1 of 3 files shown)


packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
```diff
@@ -54,48 +54,42 @@ export class SearchEngine {
...
     }
-    void this.indexDocuments()
+
+    if (this.client.isLogged) {
+      this.afterLogin()
+    } else {
+      this.client.on('login', () => {
+        this.afterLogin()
+      })
+    }
```
Ldoppea, line 72: In the future we may want to call `indexDocuments()`/`afterLogin()` from outside of the constructor. Calling async code from a constructor is a bad practice.

Ldoppea (approved):

packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
paultranvan: How would you do to avoid this async call ? Change the API, do have somehting like this?
```
const searchEngine = new SearchEngine()
await searchEngine.init()
```

PR #2729, opened 2025-02-06
feat: Do not rebuild search index from scratch 

paultranvan:
Each time a searchEngine is instanciated, we used to rebuild the search
indexes from scratch, i.e. query the local database and compute the
indexes.  This can be quite time consuming, especially in slow mobile
devices. Therefore, we now export the search indexes on disk, and import
them at init time.  We save the lastSeq for each index, to be the
starting point of the next changes batch updates, fetched from the local
database after a remote replication.

Some measures, performed on a 8K files database:

**Redmi Note 10**

- Query all io.cozy.files : 3700ms
- Build files index : 6500ms
- **Total: 10200 ms**

- **Import persisted index: 800 ms**
- Export index: 1057ms

**OnePlus Nord 4**

- Query all io.cozy.files : 2000ms
- Build files index : 2900ms
- **Total: 4900 ms**

- **Import persisted index: 332 ms**
- Export index: 363ms

If we compare the query+build index time w.r.t the import time, we have
a ~13x time improvement before the index is ready to use.

### DIFF
--- a/packages/cozy-dataproxy-lib/src/components/Search/useFetchResult.jsx
+++ b/packages/cozy-dataproxy-lib/src/components/Search/useFetchResult.jsx
@@ -7,6 +7,31 @@ import { getIconForSearchResult } from './getIconForSearchResult'
 
 const log = Minilog('🔍 [useFetchResult]')
 
+const searchWithRetry = async (
+  dataProxy,
+  searchValue,
+  { maxRetries = 5, delay = 500 } = {}
+) => {
+  let currentDelay = delay
+  // Make several search attemps in case it is not ready yet
+  for (let attempt = 0; attempt < maxRetries; attempt++) {
+    const searchResults = await dataProxy.search(searchValue)
+
+    if (searchResults) {
+      // A successful search will return an array, and null otherwise
+      return searchResults
+    }
+    log.info(
+      `Search attempt ${attempt + 1} failed, retrying in ${currentDelay} ms...`
+    )
+    await new Promise(resolve => setTimeout(resolve, currentDelay))
+    currentDelay *= 2 // Exponential backoff
+  }
+
+  log.error(`Search failed after ${maxRetries} attempts`)
+  return []
+}
+
 export const useFetchResult = searchValue => {
   const [state, setState] = useState({
     isLoading: true,
@@ -24,7 +49,7 @@ export const useFetchResult = searchValue => {
 
       setState({ isLoading: true, results: null, searchValue })
 
-      const searchResults = await dataProxy.search(searchValue)
+      const searchResults = await searchWithRetry(dataProxy, searchValue)
 
       const results = searchResults.map(r => {
         // Begin Retrocompatibility code, to be removed when following PR is merged: https://github.com/cozy/cozy-web-data-proxy/pull/10
--- a/packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
+++ b/packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
@@ -15,13 +15,13 @@ import {
   SEARCHABLE_DOCTYPES
 } from './consts'
 import { getPouchLink } from './helpers/client'
-import { getSearchEncoder } from './helpers/getSearchEncoder'
-import {
-  addFilePaths,
-  computeFileFullpath,
-  shouldKeepFile
-} from './helpers/normalizeFile'
 import { normalizeSearchResult } from './helpers/normalizeSearchResult'
+import {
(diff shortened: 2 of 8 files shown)


packages/cozy-dataproxy-lib/src/components/Search/useFetchResult.jsx
```diff
@@ -7,6 +7,31 @@ import { getIconForSearchResult } from './getIconForSearchResult'
...
+  dataProxy,
+  searchValue,
+  { maxRetries = 5, delay = 500 } = {}
+) => {
+  let currentDelay = delay
+  // Make several search attemps in case it is not ready yet
+  for (let attempt = 0; attempt < maxRetries; attempt++) {
+    const searchResults = await dataProxy.search(searchValue)
+
+    if (searchResults && Array.isArray(searchResults)) {
```
zatteo, line 20: If the search is unsuccesful, what can it return ? null ? Or something else ?

packages/cozy-dataproxy-lib/src/components/Search/useFetchResult.jsx
```diff
@@ -7,6 +7,31 @@ import { getIconForSearchResult } from './getIconForSearchResult'
 
 const log = Minilog('🔍 [useFetchResult]')
 
+const searchWithRetry = async (
+  dataProxy,
+  searchValue,
+  { maxRetries = 5, delay = 500 } = {}
```
zatteo, line 13: 500 is maybe an high delay for the first attempt ? But I don't know what time can this kind of thing take.

zatteo (approved):

packages/cozy-dataproxy-lib/src/components/Search/useFetchResult.jsx
paultranvan: The very first attempt will be right away. It is actually the second attempt that will make 500ms. The performances measures in the PR description give some insights of the time to have initialized index

packages/cozy-dataproxy-lib/src/components/Search/useFetchResult.jsx
paultranvan: If the index are not initialized, null will be returned
If the index *are* initialized, and the search is working it will be an array

zatteo: So theoretically no need to check `Array.isArray(searchResults)` ? But it's a nit, no problem.

packages/cozy-dataproxy-lib/src/components/Search/useFetchResult.jsx
zatteo: Let's try with this and we'll see.

packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
```diff
@@ -76,13 +84,32 @@ export class SearchEngine {
...
     }
-    // Create search indexes by querying docs, either locally or remotely
-    for (const doctype of SEARCHABLE_DOCTYPES) {
-      const searchIndex = await this.indexDocsForSearch(
-        doctype as keyof typeof SEARCH_SCHEMA
-      )
-      if (searchIndex) {
-        this.searchIndexes[doctype] = searchIndex
+
+    if (this.isLocalSearch) {
```
Ldoppea, line 88: there seems to be no code branch for the remote search anymore. Also those are 2 consecutive `if(this.isLocalSearch)`, should we merge theme?

packages/cozy-dataproxy-lib/src/components/Search/useFetchResult.jsx
paultranvan: Yes, I guess we could just do `if (searchResults)` and avoid an additional check :+1:

packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
paultranvan: This is definitely not expected, good catch! 
Fixed within 0443ed641964ef3e75e4b264ca5b7dadcb06d156

packages/cozy-dataproxy-lib/src/components/Search/useFetchResult.jsx
paultranvan: Fixed within 344c30cb2ba34dd7bd5b9cbcd37e312dbeff1848

Ldoppea (approved):

PR #2732, opened 2025-02-10
fix(cozy-dataproxy-lib): Embed assistant's icon as base64 instead of PNG

Ldoppea:
The assistant's icon cannot be a SVG because it uses some conical gradients

Also we don't want to embed it as a PNG resource for now because we encounter some issues in the Flagship app's WebViews when loaded in offline mode (the cache control seems not to work after an app's restart)

To fix those issues, we chose to embed the icon PNG as a base64 string until we find a better solution.


zatteo (approved):
Tested on Home/Drive desktop/mobile ?

paultranvan (approved):
Thanks! :tada:

PR #2734, opened 2025-02-20
Better handling of SearchEngine indexing

paultranvan:
See commit descriptions


packages/cozy-dataproxy-lib/src/search/consts.ts
```diff
@@ -40,3 +40,4 @@ export const DOCTYPE_DEFAULT_ORDER = {
 }
 
 export const DATAPROXY_STORAGE_PREFIX = '@dataproxy'
+export const MIN_DELAY_BEFORE_EXPORT = 60000 // 60s
```
Ldoppea, line 43: Not sure about the impact of this delay.

My understanding is that export is done at the end of indexing, which is done at the end of a replication.

The replication debounce is 30s (with a max of 10min).  So we can have up to 1 replication every 30s, and so 1 indexation every 30s.

In that scenario, does this means that only 1 of 2 indexations will be persisted?

What happens if we open the app and import a stale index? Will we have to wait next replication to re-index up-to-date data?

Ldoppea: After reading some code, my understanding is that we do an incremental indexing on init? So this will allow to have up-to-date data before the next replication?

Ldoppea (approved):

packages/cozy-dataproxy-lib/src/search/consts.ts
paultranvan: > My understanding is that export is done at the end of indexing, which is done at the end of a replication.

Yes

> The replication debounce is 30s (with a max of 10min). So we can have up to 1 replication every 30s, and so 1 indexation every 30s.
In that scenario, does this means that only 1 of 2 indexations will be persisted?

If there are 2 successive indexing, only the last index result will be persisted, which is what we want. The replication and export delays are actually hard to set and might evolve. So it's good have some protection to avoid exporting indexes too often. We typically could decide to trigger replication each time the app goes foreground: in this case, we do not want to export the index each time

> What happens if we open the app and import a stale index? Will we have to wait next replication to re-index up-to-date data?

Yes, the index will potentially have missing or stale data. But we will update it right away: we persist the associated `lastSeq` with the index, and once we import, we request the changes that occured since that `lastSeq`: see https://github.com/cozy/cozy-libs/blob/master/packages/cozy-dataproxy-lib/src/search/SearchEngine.ts#L126 and https://github.com/cozy/cozy-libs/blob/master/packages/cozy-dataproxy-lib/src/search/indexDocs.ts#L71

> my understanding is that we do an incremental indexing on init? So this will allow to have up-to-date data before the next replication?

that is correct!

PR #2748, opened 2025-03-14
Move search components from cozy-dataproxy-lib to cozy-search

zatteo:
We mixed logic and UI components in cozy-dataproxy-lib at the beginning, but let's separate them today.

## cozy-dataproxy-lib cutting plan

Following discussion with @paultranvan  and @Ldoppea and work in progress on search and bar

**Currently**

cozy-dataproxy-lib carries the dataproxy logic, in particular the DataProxyProvider, as well as the search UI that uses the dataproxy.

**With this PR**

cozy-dataproxy-lib : lib in cozy-libs that carries the dataproxy logic, including the DataProxyProvider
=> some package may use only the API exposed by the dataproxy and not the UI components so let's separate

cozy-search : lib in cozy-libs that carries the UI of the search that uses dataproxy
=> this meta-component will grow enormously, justifying a separate package from cozy-ui
=> the search bar will be used in cozy-home and cozy-bar, justifying a separate package for cozy-bar.

**In the future**

We'll migrate what's left of cozy-dataproxy-lib into cozy-client, since cozy-client queries will eventually rely on dataproxy, so we might as well put everything back in the same place, a priori via a subpackage in the cozy-client repository (whatever).


Ldoppea:
⚠️  The breaking change description talk about imports from `cozy-dataproxy-lib` but not from `cozy-dataproxy-lib/ui`. As both paths are available, we should mention both in the BC.

packages/cozy-search/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+1. Update `cozy-scripts` version of your app to 8.4.0.
+
+2. Add the provider in your tree :
+
+```jsx
+<DataProxyProvider>
+  { children }
+</DataProxyProvider>
+```
```
Ldoppea, line 13: Here I would mention that the SearchEngine is in the cozy-dataproxy-lib project, so it is easier to understand why the DataProxyProvider is needed.

Ldoppea (approved):

PR #2754, opened 2025-03-27
Update search style

zatteo:


packages/cozy-search/src/components/Search/SearchBarDesktop.jsx
JF-Cozy, line 1: maybe we can avoid the exception here and make the change in Cozy too?

JF-Cozy (approved):

packages/cozy-search/src/components/Search/SearchBarDesktop.jsx
zatteo: I thought about it too 🤷As we try to keep Cozy as is for current users, I decided to merge like this.

PR #2766, opened 2025-04-16
Add AssistantLink and fix assistant bugs

zatteo:


packages/cozy-search/src/components/Search/AssistantLink.jsx
```diff
@@ -0,0 +1,36 @@
...
+      {children({
+        openAssistant
+      })}
+    </AssistantProvider>
+  )
+}
+
+const AssistantLinkWrapper = ({ children }) => {
+  return (
+    <AssistantProvider>
```
JF-Cozy, line 30: it will be 2 AssistantProvider in the app, isn't it?

packages/cozy-search/src/index.ts
```diff
@@ -1,3 +1,4 @@
+export { default as AssistantLink } from './components/Search/AssistantLink'
```
JF-Cozy, line 1: we should add it in the documentation as well

JF-Cozy (approved):

packages/cozy-search/src/components/Search/AssistantLink.jsx
zatteo: It is possible to have two AssistantProvider now yes. Current code organization of cozy-search is less and less relevant for what we want to do with Twake theme. I think we will need to take times to refactor it like cozy-bar...

JF-Cozy: I think we can just add in the documentation that the provider is needed, and don't add it in the exposed component

PR #2772, opened 2025-05-07
feat: Correctly init DataProxy

paultranvan:
We now render the `DataProxyProvider` children only when the DataProxy
is  correctly initialized on the web.
To know this, we rely on the cozy-web-dataproxy app to send a "ready"
message once it is indeed ready.
This is meant to avoid some timing issues, where the app is ready to
request the DataProxy, but the DataProxy is not initialized yet.

### DIFF
--- a/packages/cozy-dataproxy-lib/src/dataproxy/DataProxyProvider.jsx
+++ b/packages/cozy-dataproxy-lib/src/dataproxy/DataProxyProvider.jsx
@@ -1,5 +1,5 @@
 import * as Comlink from 'comlink'
-import React, { useContext, useState, useEffect } from 'react'
+import React, { useContext, useState, useEffect, useCallback } from 'react'
 
 import { useClient } from 'cozy-client'
 import { isFlagshipApp } from 'cozy-device-helper'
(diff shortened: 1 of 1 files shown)


packages/cozy-dataproxy-lib/src/dataproxy/DataProxyProvider.jsx
```diff
@@ -117,21 +118,59 @@ export const DataProxyProvider = React.memo(({ children, options = {} }) => {
...
-  const value = {
-    dataProxyServicesAvailable,
-    search
-  }
+  useEffect(function () {
+    window.addEventListener('message', onReceiveMessage)
+    return function () {
+      window.removeEventListener('message', onReceiveMessage)
+    }
+  })
```
Ldoppea, line 141: This useEffect does not have any dependency array, this means that it will (un)register the eventListener on every render.

Seems like you'll have to create useCallbacks for onReceiveMessage and onIframeLoaded but those should be pretty stable as the only dependency is on `setDataProxy`

Ldoppea (commented):
> We now create the DataProxy iframe only when it is correctly initialized on the web

Do you mean "DataProxy comlink" instead of "DataProxy iframe"?

packages/cozy-dataproxy-lib/src/dataproxy/DataProxyProvider.jsx
paultranvan: Good point, I added the useCallback, does it look ok?

Ldoppea (approved):

PR #2773, opened 2025-05-07
feat: Add support for requestLink

paultranvan:
This adds the capacity to request the cozy-client through the DataProxy,
and then relying on local Pouch or remote Stack

:warning: This relies on #2772 that should be merged first

### DIFF
--- a/packages/cozy-dataproxy-lib/src/dataproxy/DataProxyProvider.jsx
+++ b/packages/cozy-dataproxy-lib/src/dataproxy/DataProxyProvider.jsx
@@ -152,9 +152,19 @@ export const DataProxyProvider = React.memo(({ children }) => {
         return result
       }
 
+      // Request through cozy-client
+      const requestLink = async (operation, options) => {
+        log.log('Send request to DataProxy : ', operation)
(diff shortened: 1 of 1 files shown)


packages/cozy-dataproxy-lib/src/dataproxy/DataProxyProvider.jsx
```diff
@@ -117,21 +118,69 @@ export const DataProxyProvider = React.memo(({ children, options = {} }) => {
...
+        const result = await dataProxy.search(search, options)
+        return result
+      }
+
+      // Request through cozy-client
+      const requestLink = async (operation, options) => {
+        log.log('Send request to DataProxy : ', operation)
+        if (options.fetchPolicy) {
+          // Functions cannot be serialized and thus passed to the iframe
+          delete options.fetchPolicy
```
Ldoppea, line 157: nit: in the future we may want to recreate a new options object without the fetchPolicy instead of deleting the attribute. This would prevent unexpected side effects.

Ldoppea (approved):

PR #2775, opened 2025-05-12
Hide hidden apps in search

zatteo:


packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
```diff
@@ -188,6 +189,17 @@ export class SearchEngine {
...
+      this.client.registerPlugin(flag.plugin as unknown as () => void, {})
+      // If we agree this is a good solution, we should maybe add a flag.store.setAll(object)
+      for (const [key, value] of Object.entries(
+        // @ts-expect-error instanceOptions not correctly typed
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+        this.client.instanceOptions.flags
+      )) {
+        flag.store.set(key, value)
+      }
+    }
```
zatteo, line 202: I am not familiar with all the cozy client instantiation in cozy-dataproxy-web and in the shared worker but :
- I had no flags in SearchEngine when using `flag('myflag')`
- flags are presents in `client.instanceOptions.flags` (which comes from shared worker which comes from cozy-dataproxy-web)

I tried to initialize flags with `await flag.initialize(this.client)` without success (it never resolves). I did not go further and tried like here to populate the flag store manually with what we have in `client.instanceOptions.flags`. What are your thought about it?

paultranvan: My hypothesis here is that flags are normally taken from the DOM, but since your are in the shared-worker, it cannot work. As you said, the `instanceOptions` are given by cozy-dataproxy-web, which have the DOM and thus the flags correctly initialized, so that is why it exists in the `instanceOptions` of the SearchEngine client, provided by the shared-worker.

I would suggest 2 things here: 
- Do this manual mapping directly at the shared-worker initialization [here](https://github.com/cozy/cozy-web-data-proxy/blob/main/src/dataproxy/worker/shared-worker.ts#L118), so we could have the flags in the SearchEngine but also in any other feature exposed by the dataproxy.  
- There is a `flags.enable(array)` method that seems to do the work itself: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-flags/src/flag.js#L40 

Note that the issue should not exist in flagship, as it uses the AA client directly, with all the flags properly set 

Any thought @Ldoppea ?

zatteo: Interesting, I did not know about flags.enable which is exactly what I was looking for

Ldoppea: > My hypothesis here is that flags are normally taken from the DOM, but since your are in the shared-worker, it cannot work

This is part of the explanation, but when the DOM is not available, flags fallback to [initializeFromRemote()](https://github.com/cozy/cozy-libs/blob/f4f750e6f06c085baa98a8e7143f4eecfb908c19/packages/cozy-flags/src/flag.js#L142)

The problem here is that calling the stack is slow and so the worker's code will execute before the end of the initialization.

You should be able to wait for the flags initialization [like we did in the Flagship app](https://github.com/cozy/cozy-flagship-app/blob/master/src/libs/client.js#L82-L83)

However, as we already have the flag values in the options, then @paultranvan's solution seems to be great

zatteo: 👍

See https://github.com/cozy/cozy-web-data-proxy/compare/feat/enable-flag-shared-worker?expand=1

paultranvan (approved):

PR #2780, opened 2025-05-22
Wait bridge before loading iframe

zatteo:


packages/cozy-external-bridge/src/container/index.ts
JF-Cozy, line 1: nit: for this entire commit, maybe we should return object 🤔 as we have habits to destructure returned value from hooks... (it's a very nit).

Also in commit message `This will allow to avoid to avoid`

JF-Cozy (approved):

packages/cozy-external-bridge/src/container/index.ts
zatteo: Agree, it will be more robust long term

PR #2783, opened 2025-05-27
Search fixes (remove last assistant code in search, scroll when navigating keyboard, ...)

zatteo:
## Refactored
- Definitively separate search from AI assistant. No more remains of confusing UI or wording. No more assistant showing a loader as if it was processing something whereas it was just waiting a question.

## Fixed
- Pressing Enter to open a search result when navigating with keyboard is fixed
- Scrolling when navigating with keyboard below the visibility is fixed
- Closing AssistantDialog and SearchDialog goes back to previous page instead of /

## Improved
- Use a href links when possible so the user can choose if he wants to open in same tab or new tab. If not possible, same app = same tab, other app = new tab.


packages/cozy-search/src/components/Search/SearchBarDesktop.jsx
```diff
@@ -114,23 +95,17 @@ const SearchBarDesktop = ({
           componentsProps={{
             inputBase: {
               onKeyDown: handleKeyDown,
-              endAdornment: isAssistantEnabled() ? (
-                <AssistantButton onClick={handleClick} size={size} />
-              ) : undefined
+              endAdornment: isAssistantEnabled() && (
```
JF-Cozy, line 99: no warning because replacing undefined by false (if I'm right)?

packages/cozy-search/src/components/ResultMenu/ResultMenuItem.jsx
```diff
@@ -46,8 +48,24 @@ const ResultMenuItem = ({
     secondaryText
   )
 
+  useEffect(() => {
+    if (selected) {
+      itemRef?.current?.scrollIntoView({
+        behavior: 'smooth',
```
JF-Cozy, line 56: nit: if the values are those by default, maybe we can remove these key to simplify

packages/cozy-search/src/components/Search/useFetchResult.jsx
```diff
@@ -65,7 +69,17 @@ export const useFetchResult = searchValue => {
           primary: r.title,
           secondary: r.subTitle,
           onClick: () => {
-            window.open(r.url)
+            if (r.slug === client.appMetadata.slug) {
+              try {
+                const url = new URL(r.url)
+                const hash = url.hash.replace('#', '')
```
JF-Cozy, line 76: maybe we can use something from react-router-dom to deal with the hash and so avoid to remove it 🤔

packages/cozy-search/src/components/ResultMenu/styles.styl
```diff
@@ -6,3 +6,6 @@
   &-inner
     max-height 16.5rem
     overflow auto
+
+.resultMenuItem
+  color: var(--primaryTextColor) // to avoid visited color on a component
```
JF-Cozy, line 11: we don't use visited color anywhere, maybe we can do this in cozy-ui 🤔

packages/cozy-search/src/components/AssistantMobile.jsx
```diff
@@ -28,7 +29,7 @@ export const AssistantMobile = ({ componentsProps }) => {
         />
       }
       type="button"
-      onClick={() => navigate('connected/search')} // FIXME this route is related to home app
+      onClick={() => navigate(`connected/search?returnPath=${pathname}`)} // FIXME this route is related to home app
```
JF-Cozy, line 32: comment is still relevant?

If think (not sure) we can use param in navigate to use queryparam and avoid using `?`

packages/cozy-search/src/components/Views/AssistantDialog.jsx
```diff
@@ -13,10 +13,21 @@ const AssistantDialog = () => {
...
   const { isMobile } = useBreakpoints()
   const navigate = useNavigate()
+  const location = useLocation()
+
   const { conversationId } = useParams()
 
   const onClose = () => {
-    navigate('..')
+    try {
+      const returnPath = new URLSearchParams(location.search).get('returnPath')
```
JF-Cozy, line 22: you can use `useSearchParams` from react-router I think https://reactrouter.com/6.30.1/hooks/use-search-params `const [searchParams] = useSearchParams(); const returnPath = searchParams.get('returnPath')`

packages/cozy-search/src/components/AssistantProvider.jsx
JF-Cozy, line 1: no code has to be changed in the assistant?

JF-Cozy (approved):

packages/cozy-search/src/components/Search/useFetchResult.jsx
zatteo: I tried different way, and react-router only understand react-router. In v6 or v7 nothing exists for example to do : `navigate("https://example.com/#/mypage")`. react-router does not understand  that we want to do `navigate("/mypage")`

packages/cozy-search/src/components/AssistantMobile.jsx
zatteo: Comment is still relevant I hesitated to try to fix it but want to do other things today

packages/cozy-search/src/components/Search/SearchBarDesktop.jsx
zatteo: After verification no warning, but good idea to check 👍

packages/cozy-search/src/components/AssistantProvider.jsx
zatteo: I think not, we can't test it anymore for the moment but it will be working again today if everything is okay.

packages/cozy-search/src/components/AssistantMobile.jsx
JF-Cozy: and what about using query params?

PR #2810, opened 2025-08-28
chore: Display warning for long commit body lines

paultranvan:
We used to have an error for commit body lines > 72. This is quite annoying because it forces the dev to redo the commit. Furthermore, this rule is not always legitimate, typically when one want to put a long URL, forcing to add a line break in the URL.

So let's display a warning rather than an error.

### DIFF
--- a/packages/commitlint-config-cozy/index.js
+++ b/packages/commitlint-config-cozy/index.js
@@ -1,7 +1,7 @@
 module.exports = {
   rules: {
     'body-leading-blank': [1, 'always'],
-    'body-max-line-length': [2, 'always', 72],
+    'body-max-line-length': [1, 'always', 72],
     'footer-leading-blank': [1, 'always'],
     'header-max-length': [2, 'always', 72],
     'scope-case': [0, 'always', 'lower-case'],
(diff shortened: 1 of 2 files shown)


rezk2ll (approved):

paultranvan:
> ﻿I am less and less in favor of these rules. 

Well, I never was :smile: 
BUT, IMO some weak rules at still good to enforce good practices. Semantic commits are useful for CI and it's somewhat useful to know the purpose of a commit (fix, feat, style, ...). 
Having  a good git title is also important, but it is the developer / reviewers responsibility. Same for body description.
Restricting body lines to 72 is okish, but counter productive if it's strictly forbidden.
So basically, I like keeping few rules, but I dislike aggressive tools blindly forcing you to follow them.

zatteo:
> > I am less and less in favor of these rules.
> 
> IMO some weak rules at still good to enforce good practices. Semantic commits are useful for CI and it's somewhat useful to know the purpose of a commit (fix, feat, style, ...). Having a good git title is also important, but it is the developer / reviewers responsibility. Same for body description. Restricting body lines to 72 is okish, but counter productive if it's strictly forbidden. So basically, I like keeping few rules, but I dislike aggressive tools blindly forcing you to follow them.

(I totally forgot semantic commits so that is a very good point.)

PR #2814, opened 2025-09-05
Update rsbuild packages and watch cozy-* packages when using rlink

zatteo:


config/rsbuild-config-cozy-app/getRsbuildConfig.js
```diff
@@ -85,6 +85,10 @@ function getRsbuildConfig({
     },
     tools: {
       rspack: {
+        // We want to watch cozy-* packages when using rlink
+        watchOptions: {
+          ignored: /\.git|node_modules\/(?!cozy-)/
```
JF-Cozy, line 90: so we need to name our lib with `cozy-*` 🤔 don't have any better idea right now but it's a limitation for sure... 🤔  🤔 maybe we can at least document it somewhere?

JF-Cozy (approved):
good job 👍 Thank you for digging deeper.

config/rsbuild-config-cozy-app/getRsbuildConfig.js
zatteo: README updated

PR #2815, opened 2025-09-09
Dedup classnames and lodash packages

zatteo:
I am starting to play with packages in cozy-libs to reduce install time and size.


JF-Cozy (commented):
maybe we should use fixed version to better control them instead of using `^`. I think we decide a long time ago to use `^` only for cozy-* packages

PR #2819, opened 2025-09-17
feat(cozy-realtime): Add support for shared drives files

doubleface:
If you want to activate realtime on a given shared drive : 

```javascript
import CozyRealtime from 'cozy-realtime'

const client = ...
const sharedDriveId = 'myshareddriveid'

const realtime = new CozyRealtime({client, sharedDriveId})
realtime.start()
```

```javascript
realtime.stop()
```

To stop the subscription

cozy-stack will automatically subscribe to shared drive files events
As it is implemented now, it is only possible to subscribe to a given shared drive, one at a time.

Needs https://github.com/cozy/cozy-stack/pull/4559 to be merged


zatteo (approved):
Looks nice. I don't see what we could do differently.

PR #2831, opened 2025-10-03
fix: Resolve revoke self action does not work in view mode :bug:

lethemanh:
### Change:

Fix bug revoke self action of access rights modal does not work in view mode.

### Related to:

[cozy/cozy-drive#3532](https://github.com/cozy/cozy-drive/pull/3532)

### DIFF
--- a/packages/cozy-dataproxy-lib/CHANGELOG.md
+++ b/packages/cozy-dataproxy-lib/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [4.11.0](https://github.com/cozy/cozy-libs/compare/cozy-dataproxy-lib@4.10.0...cozy-dataproxy-lib@4.11.0) (2025-10-03)
+
+### Features
+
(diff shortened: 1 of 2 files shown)


zatteo:
Is this related to shared drive?

lethemanh:
> Is this related to shared drive?

Yes, it's. The access rights modal is using for both normal files/folders and shared drive. The button `Remove me from sharing` does not work.

packages/cozy-sharing/src/components/ShareModal/SharingDetailsModal.jsx
```diff
@@ -44,7 +44,7 @@ export const SharingDetailsModal = ({
               name={t(`${documentType}.share.sharedBy`, {
                 name: getDisplayName(owner)
               })}
-              url={owner.instance}
+              url={owner?.instance}
```
rezk2ll, line 47: if the owner didn't exist we should not display the OwnerIdentity component in the first place

packages/cozy-sharing/src/DoctypeContact.js
```diff
@@ -19,6 +19,9 @@ export const getInitials = (contactOrRecipient, defaultValue = '') => {
 }
 
 export const getDisplayName = (contact, defaultValue = '') => {
+  if (!contact) {
+    return defaultValue
+  }
```
rezk2ll, line 24: you don't need this

PR #2989, opened 2026-04-17
feat(cozy-sharing): add props to tune link sharing flow

rezk2ll:
### changes

Two independent, opt-in props let the app tune the link sharing UX without affecting the default behavior:

- `showGenerateLinkButton`: the initial button only creates the link. Copying becomes a deliberate second action once the link exists.
- `autoOpenShareRestriction`: surfaces the password / expiry / edit-rights popup right after a link is generated, so the user configures permissions immediately.

**_Both default to false — the current behavior is preserved._**

#### only `showGenerateLinkButton` is active

https://github.com/user-attachments/assets/300d3b73-83d0-4f99-bf5a-25e56d5da87e

#### both props active
https://github.com/user-attachments/assets/90a7da97-0e6e-4e6b-b88f-d8fe584e0033

## Release Notes

* **New Features**
  * Added optional controls to display link generation buttons and automatically open share restriction settings across sharing dialogs.

### DIFF
--- a/packages/cozy-sharing/src/components/ShareModal/ShareModal.jsx
+++ b/packages/cozy-sharing/src/components/ShareModal/ShareModal.jsx
@@ -36,7 +36,7 @@ export const ShareModal = withLocales(props => {
 
   const isFederatedMode = flag('drive.federated-shared-folder.enabled')
   if (isFederatedMode && allLoaded) {
-    return <FederatedFolderModal document={document} onClose={rest.onClose} />
+    return <FederatedFolderModal document={document} {...rest} />
   }
 
   const isEditable =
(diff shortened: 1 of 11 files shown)


JF-Cozy:
@rezk2ll be careful we don't use flags inside libs anymore. It's very painful to remove, because we have to update the libs everywhere.
Please use bool prop instead, true or false by default (the way that seems more logic for you) with a showSomehting or hideSomething keyword, or isActive or isInactive... this kind of things...
That way we just have to set the prop differently in the app, and take time to remove the code lib side if needed

zatteo (approved):

packages/cozy-sharing/src/components/ShareModal/ShareModal.jsx
```diff
@@ -36,7 +36,7 @@ export const ShareModal = withLocales(props => {
 
   const isFederatedMode = flag('drive.federated-shared-folder.enabled')
   if (isFederatedMode && allLoaded) {
-    return <FederatedFolderModal document={document} onClose={rest.onClose} />
+    return <FederatedFolderModal document={document} {...rest} />
```
JF-Cozy, line 39: nit: not a big fan of that because we don't see clearly what props are passed, and how many they are

JF-Cozy (approved):
maybe it's time (in an other PR) to re-organise props... by using providers maybe? To avoid passing prop unnecessary and to have a better API app-side 🤔

PR #2990, opened 2026-04-20
fix(cozy-sharing): Preserve user gesture for Safari clipboard writes

rezk2ll:
## Context

Copying a share link waits for an async permission round-trip before `navigator.clipboard` is called. Safari drops the user-gesture context across any `await`, which is stricter than Chrome or Firefox.

## Problem

On Safari, the first click on **Copy link** — or **Confirm** in the restriction modal — surfaces a red "copy failed" toast and logs `NotAllowedError`. A second click appears to work, but only because the link is now in state: the async step is skipped and the clipboard write runs in the same tick as the click.

Reported in linagora/twake-drive#3766.

## Expected behavior

The link is copied on the first click, no error toast.

---

Once this ships, linagora/twake-drive#3767 can close in favor of a dependency bump there.

### DIFF
--- a/packages/cozy-sharing/src/components/ShareByLink.jsx
+++ b/packages/cozy-sharing/src/components/ShareByLink.jsx
@@ -75,18 +75,29 @@ const ShareByLink = ({
   }
 
   const generateLinkAndCopyLinkToClipboard = async () => {
-    let linkToCopy = link
-    const wasNewlyGenerated = !linkToCopy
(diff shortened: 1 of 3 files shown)


zatteo (approved):
Interesting. For those interested: 
- https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem

doubleface (approved):

PR #2992, opened 2026-04-23
fix(cozy-sharing): Handle structured name on share recipients

rezk2ll:
## Context

The share modal is the entry point for sharing anything in Cozy. When it crashes, the whole sharing flow is unreachable.

## Problem

On certain instances, opening the share modal and selecting a recipient brings the whole page down. The `io.cozy.contacts` doctype defines a recipient's `name` as a structured object, but the modal's fallback path was reading it as a string. Any recipient arriving in its documented shape therefore took the modal down instead of rendering.

## Expected behavior

Selecting a recipient renders their initials and display name without error, whatever combination of `name`, `public_name`, and `email` the recipient carries.

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined contact display name resolution with enhanced fallback logic to ensure consistent display across various contact information structures.
  * Enhanced contact initial generation to work reliably with different types of contact data and formats.
  * Improved how contact information is displayed when certain standard fields are unavailable or incomplete.

### DIFF
--- a/packages/cozy-sharing/src/DoctypeContact.js
+++ b/packages/cozy-sharing/src/DoctypeContact.js
@@ -2,29 +2,18 @@ import { models } from 'cozy-client'
 import { Contact as DoctypeContact } from 'cozy-doctypes'
 const ContactModel = models.contact
 
-const isContact = candidate => {
-  return candidate._type === 'io.cozy.contacts'
-}
 export const getInitials = (contactOrRecipient, defaultValue = '') => {
-  if (isContact(contactOrRecipient)) {
-    return ContactModel.getInitials(contactOrRecipient)
-  } else {
-    // @todo Extract to RecipientModel ?
-    const s =
-      contactOrRecipient.public_name ||
-      contactOrRecipient.name ||
-      contactOrRecipient.email
-    return (s && s[0].toUpperCase()) || defaultValue
+  if (contactOrRecipient.public_name) {
+    return contactOrRecipient.public_name[0].toUpperCase()
   }
+  return ContactModel.getInitials(contactOrRecipient) || defaultValue
 }
 
-export const getDisplayName = (contact, defaultValue = '') => {
-  if (isContact(contact)) {
-    return ContactModel.getDisplayName(contact)
-  } else {
-    // @todo Extract to RecipientModel ?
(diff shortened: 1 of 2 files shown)


zatteo (changes requested):
👎👎👎 green team 👎👎👎

packages/cozy-sharing/src/DoctypeContact.js
```diff
@@ -2,29 +2,33 @@ import { models } from 'cozy-client'
...
-}
+// Treat as a contact any doc with `_type: io.cozy.contacts` or with the same
+// field shape (object `name`, array `email`/`cozy`) — contacts synced by
+// external services can arrive without `_type` in the body.
+const isContact = candidate =>
+  ContactModel.isContact(candidate) ||
+  (candidate.name && typeof candidate.name === 'object') ||
+  Array.isArray(candidate.email) ||
+  Array.isArray(candidate.cozy)
+
```
paultranvan: External contacts are still contacts, I think they should be `io.cozy.contacts` anyways , and not rely on some loose document shape. Otherwise we'll have false positives, like a document with `email` or `name: {...}` at root

zatteo: True we should never merge it

rezk2ll: hmm that means the migrator service has to pass the _type always and also keep this PR as a defensive step

paultranvan: How are fetched the external contacts ? If they are fetched the same way as regular contacts, the `_type` should be set by cozy-client anyway https://github.com/linagora/cozy-client/blob/master/packages/cozy-stack-client/src/normalize.js#L11

Crash--: > If they are fetched the same way as regular contacts, the _type should be set by cozy-client 

This is exactly what I wanted to write, thanks Paul. So we need to understand why we don't have this `_type`!

rezk2ll: the real fix https://github.com/linagora/cozy-libs/pull/2995

rezk2ll: i redid the fix in this PR

PR #2994, opened 2026-04-23
fix(cozy-sharing): Make matching diacritic-insensitive

doubleface:
Allow users to search contacts and groups using accented or unaccented versions interchangeably (e.g., 'Zoe' matches 'Zoé' and vice versa). This improves UX for international users who may not remember exact accentuation.

Uses Unicode normalization with diacritic removal, with fallback for older browsers that don't support the Diacritic property in regex.

[Capture vidéo du 2026-04-23 17-12-14.webm](https://github.com/user-attachments/assets/dd5ea160-5b98-4d0e-aba6-ecfe495340d0)

https://www.notion.so/linagora/2d162718bad180a0a68ad0738dbf9ebe?v=2d162718bad180ea8a91000c4ea30ddd&p=34b62718bad180d0a4b4f9af4ca70a78&pm=s

## Summary by CodeRabbit

* **Bug Fixes**
  * Matching is now diacritic-insensitive across names, group labels, URLs and emails so accented and unaccented queries match interchangeably (e.g., "José" ↔ "Jose").
  * Queries containing regex metacharacters are treated as literal text, preventing unintended regex behavior and crashes.

* **Tests**
  * Added and expanded tests covering diacritic scenarios and regression cases to ensure stable matching behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/cozy-libs/pull/2994?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/cozy-sharing/src/suggestionMatchers.js
+++ b/packages/cozy-sharing/src/suggestionMatchers.js
@@ -2,36 +2,48 @@ import escapeRegExp from 'lodash/escapeRegExp'
 
 import { Group } from 'cozy-doctypes'
 
+let removeDiacritics
+try {
+  new RegExp('\\p{Diacritic}', 'gu')
+  removeDiacritics = str => str.replace(/\p{Diacritic}/gu, '')
+} catch {
+  removeDiacritics = str => str.replace(/[̀-ͯ]/g, '')
+}
+
+const normalize = str => removeDiacritics(str.normalize('NFD'))
+
 // TODO: sadly we have different versions of contacts' doctype to handle...
 // A migration tool on the stack side is needed here
 const emailMatch = (input, contact) => {
   if (!contact.email) return false
-  const emailInput = new RegExp(escapeRegExp(input), 'i')
+  const emailInput = new RegExp(normalize(input), 'i')
   if (Array.isArray(contact.email)) {
-    return contact.email.some(email => emailInput.test(email.address))
+    return contact.email.some(email =>
+      emailInput.test(normalize(email.address))
+    )
   }
-  return emailInput.test(contact.email)
+  return emailInput.test(normalize(contact.email))
 }
 
 const cozyUrlMatch = (input, contact) => {
   if (!contact.cozy && !contact.url) return false
-  const urlInput = new RegExp(escapeRegExp(input), 'i')
+  const urlInput = new RegExp(normalize(input), 'i')
   if (contact.cozy && Array.isArray(contact.cozy)) {
-    return contact.cozy.some(cozy => urlInput.test(cozy.url))
+    return contact.cozy.some(cozy => urlInput.test(normalize(cozy.url)))
   }
-  return urlInput.test(contact.url)
+  return urlInput.test(normalize(contact.url))
 }
 
 const groupNameMatch = (input, contactOrGroup) => {
   if (contactOrGroup._type !== Group.doctype) return false
-  const nameInput = new RegExp(escapeRegExp(input), 'i')
-  return nameInput.test(contactOrGroup.name)
+  const nameInput = new RegExp(normalize(input), 'i')
+  return nameInput.test(normalize(contactOrGroup.name))
 }
 
 const fullnameMatch = (input, contact) => {
   if (!contact.fullname) return false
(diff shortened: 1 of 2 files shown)


packages/cozy-sharing/src/suggestionMatchers.js
```diff
@@ -1,35 +1,47 @@
 import { Group } from 'cozy-doctypes'
 
+let removeDiacritics
+try {
+  new RegExp('\\p{Diacritic}', 'gu')
+  removeDiacritics = str => str.replace(/\p{Diacritic}/gu, '')
```
rezk2ll, line 6: .normalize("NFD").replace(/\p{Diacritic}/gu, "")

doubleface: Why NFD and not default NFC ? I did some research and did not find which one is better

rezk2ll: - `NFD` (Canonical Decomposition) splits precomposed characters: `'é'` (U+00E9) → `'e'` (U+0065) + `'́'` (U+0301, combining acute accent).
- `NFC` (Canonical Composition) keeps/rebuilds the precomposed form: `'é'` stays as a single U+00E9 code point.

There is a secondary nuance: some scripts (e.g. Vietnamese, some African languages) encode vowels with up to two diacritics and rely on canonical decomposition to be fully normalized. NFD is also the form consistently recommended by MDN and Unicode for this exact "strip diacritics" recipe.

packages/cozy-sharing/src/suggestionMatchers.js
```diff
@@ -1,35 +1,47 @@
...
+}
+
+const normalize = str => removeDiacritics(str.normalize('NFD'))
+
 // TODO: sadly we have different versions of contacts' doctype to handle...
 // A migration tool on the stack side is needed here
 const emailMatch = (input, contact) => {
   if (!contact.email) return false
-  const emailInput = new RegExp(input, 'i')
+  const emailInput = new RegExp(normalize(input), 'i')
```
rezk2ll: If the user types any regex metacharacter  `.`, `(`, `[`, `+`, `*`, `?`, `\`, `|`, `^`, `$`  the behavior ranges from "silently wrong matches" (`.` matches anything) to a hard `SyntaxError: Invalid regular expression` that will bubble up and break the autosuggest UI. Easy to reproduce: type `(` into the share dialog.

This is not introduced by the PR, but the PR is the second time this file gets touched for matching correctness, and the fix is two lines. I'd add an `escapeRegExp` step (either `lodash/escapeRegExp` , already a transitive dep or an inline `input.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')`) before constructing the regex.

Honestly, given that none of these matchers actually need regex semantics (they're all "does the input appear in the field, case-insensitively"), the cleaner fix is:

```js
const normalizedInput = normalize(input).toLowerCase()
// then, per field:
normalize(contact.fullname).toLowerCase().includes(normalizedInput)
```

doubleface: Totally agree. I removed regexp in another commit

Crash--:
opened 1 month ago and still not merged. Can we move forward please?

packages/cozy-sharing/src/suggestionMatchers.spec.js
```diff
@@ -116,5 +116,44 @@ describe('Suggestion matchers', () => {
...
+      expect(() =>
+        matchers.fullnameMatch('(', { fullname: 'Jon Snow' })
+      ).not.toThrow()
+      expect(matchers.fullnameMatch('(', { fullname: 'Jon Snow' })).toBeFalsy()
+      expect(matchers.fullnameMatch('.', { fullname: 'Jon Snow' })).toBeFalsy()
+      expect(
+        matchers.fullnameMatch('.', { fullname: 'Dr. House' })
+      ).toBeTruthy()
+    })
+  })
```
Crash--, line 178: can you add few tests with Vietnamese langage for instance?

doubleface: done

zatteo (approved):

rezk2ll (approved):

PR #3002, opened 2026-05-13
Allow to force light mode in cozy-bar

zatteo:
## Summary by CodeRabbit

* **New Features**
  * BarComponent now accepts `cozyThemeType` prop for theme configuration.

* **Bug Fixes**
  * Fixed prop handling to ensure explicitly provided theme type overrides default values.

* **Documentation**
  * Updated README with improved Usage section, configuration props table, and enhanced customization guidance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/cozy-libs/pull/3002)

### DIFF
--- a/packages/cozy-bar/README.md
+++ b/packages/cozy-bar/README.md
@@ -1,22 +1,21 @@
-[![Travis build status shield](https://img.shields.io/travis/com/cozy/cozy-bar)](https://travis-ci.org/cozy/cozy-bar)
 [![NPM release version shield](https://img.shields.io/npm/v/cozy-bar.svg)](https://www.npmjs.com/package/cozy-bar)
-[![NPM Licence shield](https://img.shields.io/npm/l/cozy-bar.svg)](https://github.com/cozy/cozy-bar/blob/master/LICENSE)
+[![NPM Licence shield](https://img.shields.io/npm/l/cozy-bar.svg)](https://github.com/cozy/cozy-libs/blob/main/packages/cozy-bar/LICENSE)
 
-# \[Cozy]\[] Bar Library
+# Cozy Bar
 
 ## What's Cozy?
 
 ![Cozy Logo](https://cdn.rawgit.com/cozy/cozy-guidelines/master/templates/cozy_logo_small.svg)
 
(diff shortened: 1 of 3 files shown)


packages/cozy-bar/src/components/BarComponent.jsx
```diff
@@ -80,6 +80,7 @@ const BarComponent = ({
   appSlug = getAppSlug(),
   iconPath = getDefaultIcon(),
   isPublic = false,
+  cozyThemeType,
```
JF-Cozy: we introduced componentsProps, let's use it :)

packages/cozy-bar/README.md
```diff
@@ -32,50 +31,74 @@ yarn add cozy-bar
 import 'cozy-bar/dist/stylesheet.css'
 ```
 
-### How to use
+### Usage
```
JF-Cozy, line 34: in my opinion we should either remove everything and keep it simple, either create a script to update the doc automatically when manipulating props.

packages/cozy-ui-plus/src/providers/CozyTheme/CozyThemeWithQuery.jsx
```diff
@@ -19,7 +19,7 @@ const CozyThemeWithQuery = props => {
     return null
   }
 
-  return <DumbCozyTheme {...props} type={instance?.colorScheme} />
+  return <DumbCozyTheme type={instance?.colorScheme} {...props} />
```
JF-Cozy: I think it's better to be explicit and get type from props, and create local type `_type`

packages/cozy-bar/src/components/BarComponent.jsx
zatteo: Done you can check

packages/cozy-bar/README.md
zatteo: Agree that props doc update should not be done manually. But I do not want to manage this now. There were a lot of issues in the README before (with setup also) so it's better now.

JF-Cozy: so at this stage we should simplify and remove everything 🤷‍♂️ to avoid asking us the same question next time and to fall into the maintainability trap

JF-Cozy (approved):

PR #3005, opened 2026-05-20
feat(cozy-sharing): Add autoFocus to the share recipient input

doubleface:
image

https://www.notion.so/linagora/sharing-modal-polishing-36662718bad18012a393cd5103f03a3e

## Summary by CodeRabbit

* **New Features**
  * The sharing suggestion input now receives focus automatically when shown, so you can start typing right away to find and select contacts more quickly.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/cozy-libs/pull/3005?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/cozy-sharing/src/components/ShareAutosuggest.jsx
+++ b/packages/cozy-sharing/src/components/ShareAutosuggest.jsx
@@ -205,6 +205,7 @@ const ShareAutocomplete = ({
         renderInputComponent={renderInput}
         highlightFirstSuggestion
         inputProps={{
+          autoFocus: true,
           onFocus: onAutosuggestFocus,
           onChange: onAutosuggestChange,
           onPaste: onAutosuggestPaste,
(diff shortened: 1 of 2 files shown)


zatteo (approved):
Is there case where we do not want the autoFocus?

doubleface:
You are right, I don't see any. But it just looks like a better practice to not steal the focus by default

packages/cozy-sharing/src/components/ShareAutosuggest.jsx
```diff
@@ -205,6 +206,7 @@ const ShareAutocomplete = ({
         renderInputComponent={renderInput}
         highlightFirstSuggestion
         inputProps={{
+          autoFocus,
```
JF-Cozy: nit: maybe we can just add this line, and removed the others (and so enable autoFocus every time everywhere for ever and ever)

edit: didn't see the messages... maybe we can ask POs for that... I suppose we always want autofocus in a single input in a modal when it opens 🤷‍♂️

doubleface: Yes, let's always autofocus

PR #3006, opened 2026-05-20
fix(cozy-sharing): Filter out the owner from the list of suggested users

doubleface:
Do not display the user's own instance in the list of users to share with

https://www.notion.so/linagora/sharing-modal-polishing-36662718bad18012a393cd5103f03a3e

## Summary by CodeRabbit

* **Bug Fixes**
  * The sharing recipients input now excludes the current user from recipient suggestions.

* **Tests**
  * Added a test to verify the current user is not shown among recipient suggestions while other contacts and groups remain available.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/cozy-libs/pull/3006?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
+++ b/packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
@@ -17,6 +17,8 @@ import {
  * We retrieve all the contacts that are reachable and the groups they belong to.
  * In order to display the total number of members in each group, we also retrieve the contacts that are not reachable.
  */
+const isCurrentUser = contact => contact.me === true
+
 const ShareRecipientsInput = ({
   currentRecipients,
   recipients,
@@ -57,6 +59,12 @@ const ShareRecipientsInput = ({
       (isQueryLoading(unreachableContactsWithGroupsResult) ||
         unreachableContactsWithGroupsResult.hasMore))
 
+  const reachableContacts = (reachableContactsResult.data || []).filter(
+    contact => !isCurrentUser(contact)
+  )
+  const unreachableContactsWithGroups = (
+    unreachableContactsWithGroupsResult.data || []
+  ).filter(contact => !isCurrentUser(contact))
   const currentRecipientGroups = currentRecipients
     .map(({ id }) => id)
     .filter(Boolean)
@@ -68,7 +76,7 @@ const ShareRecipientsInput = ({
     )
     .map(contactGroup => {
       const reachableMembers = getContactsFromGroupId(
-        reachableContactsResult.data,
+        reachableContacts,
         contactGroup.id
       ).map(contact => ({
         ...contact,
@@ -83,7 +91,7 @@ const ShareRecipientsInput = ({
       }
 
       const unreachableMembers = getContactsFromGroupId(
-        unreachableContactsWithGroupsResult.data,
+        unreachableContactsWithGroups,
         contactGroup.id
       ).map(contact => ({
         ...contact,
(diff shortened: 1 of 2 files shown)


packages/cozy-sharing/src/components/ShareRecipientsInput.spec.jsx
```diff
@@ -116,6 +126,79 @@ describe('ShareRecipientsInput component', () => {
...
+  it("should exclude current user's instance from suggestions", async () => {
+    const contacts = [
+      {
+        id: 'df563cc4-6440',
+        _id: 'df563cc4-6440',
+        _type: 'io.cozy.contacts',
+        cozy: [
+          null,
+          {
+            url: 'HTTPS://cozy.tools/some/path/'
```
JF-Cozy: nit: caps on https

packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
JF-Cozy, line 1: could we maybe rely on isOwner method, somehow? 🤔

doubleface: I don't think we can use isOwner here since we need to suggest recipients even when the sharing does not exist yet.

What is possible is to use the "me" attribute of the contact corresponding to the user but I find it more reliable to use client uri

zatteo: Why `me` is less reliable?

packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
```diff
@@ -17,6 +18,31 @@ import {
...
  */
+const normalizeUrl = value => {
+  if (!value) return null
+
+  const stringValue = value.toString().trim()
+
+  try {
+    return new URL(stringValue).origin.toLowerCase()
+  } catch {
+    return stringValue.toLowerCase().replace(/\/+$/, '')
```
zatteo: What does this catch really catch? If new URL fails it means there is an issue and we should return null?

packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
doubleface: Actually, it was a wrong supposition on my side. Now I use the me attribute and the code is a lot simpler

packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
doubleface: Not needed anymore since I now use the "me" attribute

packages/cozy-sharing/src/components/ShareRecipientsInput.spec.jsx
doubleface: Not needed anymore since I now use the "me" attribute from contacts

packages/cozy-sharing/src/components/ShareRecipientsInput.jsx
```diff
@@ -17,6 +18,12 @@ import {
  * We retrieve all the contacts that are reachable and the groups they belong to.
  * In order to display the total number of members in each group, we also retrieve the contacts that are not reachable.
  */
+const isCurrentUser = contact => {
+  if (contact._type !== Contact.doctype) return false
+
+  return contact.me === true || contact.me === 'true'
```
zatteo: Can `me` really be a string 'true' ?

zatteo: Why do we need to check if it is a contact? Because it can be a group?

doubleface: This check is not needed. just `contact.me === true`

zatteo (approved):

PR #3011, opened 2026-05-29
feat: Add prop to disable sharing actions

shepilov:
## Summary

Adds a first-class `componentsProps.sharingActions.disabled` prop to `cozy-viewer`.

This lets consumer apps disable viewer sharing action buttons without relying on CSS selectors against `cozy-sharing` internals.

## Changes

- Add `componentsProps.sharingActions.disabled` support.
- Hide the toolbar share button when sharing actions are disabled.
- Make `Footer/SharingButton` render `null` when sharing actions are disabled.
- Add prop types for `Viewer` and `ViewerContainer`.
- Document the new prop in the viewer README.
- Add unit coverage for `SharingButton`.

### DIFF
--- a/packages/cozy-viewer/src/components/Toolbar.jsx
+++ b/packages/cozy-viewer/src/components/Toolbar.jsx
@@ -45,7 +45,7 @@ const Toolbar = ({
   const client = useClient()
   const { t } = useI18n()
   const webviewIntent = useWebviewIntent()
-  const { file } = useViewer()
+  const { file, componentsProps } = useViewer()
   const { isSharingShortcutCreated, addSharingLink, loading } =
     useSharingInfos()
   const { isOwner } = useSharingContext()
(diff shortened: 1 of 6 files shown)


packages/cozy-viewer/src/components/Toolbar.jsx
```diff
@@ -54,6 +54,7 @@ const Toolbar = ({
 
   const isCozySharing = window.location.pathname === '/preview'
   const isShareNotAdded = !loading && !isSharingShortcutCreated
+  const areSharingActionsDisabled = componentsProps?.sharingActions?.disabled
```
zatteo, line 57: Interesting to see of it is done for other routes `const isCozySharing = window.location.pathname === '/preview'` but it is way better with an explicit prop.

zatteo (approved):

doubleface (approved):

PR #3015, opened 2026-06-02
fix: Disable new contact creation for shared drives forms

zatteo:
For now, shared drives are meant to be used only inside an organization, so with existing contacts of the organization. So we disable the ability to create new contact for these forms. In the future, we may enable it again when we will introduce guest user for organization

**What is not possible anymore in shared drives forms**

=> when typing a valid email, suggest it and it will create a new contact

image
 

## Summary by CodeRabbit

* **New Features**
  * Added ability to control whether users can create new contacts during the sharing process.

* **Tests**
  * Added and updated test coverage for the new contact creation control feature.

### DIFF
--- a/packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
+++ b/packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
@@ -206,6 +206,7 @@ const FederatedFolderModalContent = ({
               onPendingRecipientsChange={setPendingRecipients}
               selectedOption={selectedOption}
               onSelectedOptionChange={setSelectedOption}
+              enableCreateContact
             />
           </div>
           <WhoHasAccess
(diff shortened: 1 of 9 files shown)


JF-Cozy (commented):
seems strange to me to have `isNewContact` based on missing `_id` value (why/how a contact could be a contact without _id ?), and `enableNewContactSuggestion` that seems to be propagated everywhere 🤔 I suppose there is at least one case with this set to `false` but it seems maybe we can simplify and not deal with those two cases... ?

zatteo:
> seems strange to me to have `isNewContact` based on missing `_id` value (why/how a contact could be a contact without _id ?), and `enableNewContactSuggestion` that seems to be propagated everywhere 🤔 I suppose there is at least one case with this set to `false` but it seems maybe we can simplify and not deal with those two cases... ?

Maybe `isNewContact` is not the best name. When we type a valid email and there is no existing contact that match it, the input propose to create it, and temporary store a { type: 'io.cozy.contact', email: '...' } which means that this contact will be created when the form is submitted. So the _id allows to differentiate `io.cozy.contacts` and fake temporary contacts.

JF-Cozy (approved):
I think we should add doc on isNewContact method to explain the purpose and the mechanic.

Maybe we can use `isContactToBeCreated` and `enableCreateContact` terms to be more explicit ? 🤔

PR #3020, opened 2026-06-08
feat(viewer): Add an Edit button to the PDF viewer

Crash--:
## Summary

- Add an Edit button to the `PdfJsViewer` toolbar, shown when the host app enables PDF editing through `componentsProps.PdfViewer` (`isEnabled` together with an `opener`) and the user has write access.
- The button calls the opener with the file so the app can open the PDF in its own editor.
- Editing is toggled by the app via the `isEnabled` prop; the viewer keeps no feature-flag logic of its own, matching the existing `OnlyOfficeViewer` prop shape.

## Summary by CodeRabbit

* **New Features**
  * Added "Edit" button for PDFs when enabled and user has write access; allows opening files in external editor.

* **Localization**
  * Added translations for the PDF edit action in English, French, Russian, and Vietnamese.

### DIFF
--- a/packages/cozy-viewer/src/Viewer.jsx
+++ b/packages/cozy-viewer/src/Viewer.jsx
@@ -126,6 +126,14 @@ Viewer.propTypes = {
       /** To open the Only Office file */
       opener: PropTypes.func
     }),
+    /** Used to open a PDF in an external editor */
+    PdfViewer: PropTypes.shape({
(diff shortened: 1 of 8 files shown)


zatteo (approved):

Crash--:
It's an experimental feature. At some point maybe we'll change the PDF viewer directly.

PR #3022, opened 2026-06-08
fix(cozy-sharing): Refresh link state in share modal

doubleface:
https://app.notion.com/p/linagora/5d2f7429646042378356ecb072890d35?v=1c062718bad180d3847f000cd4c97139&p=37462718bad18075bab2e83c56b23d26&pm=s

## Summary by CodeRabbit

* **Refactor**
  * Updated sharing link handling in sharing dialogs and federated folder modals to use a centralized context source

* **Tests**
  * Added tests to verify "Anyone with the link" access message displays correctly when sharing links are available

### DIFF
--- a/packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
+++ b/packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types'
-import React, { useCallback, useEffect, useMemo, useState } from 'react'
+import React, { useCallback, useState } from 'react'
 
 import { useClient } from 'cozy-client'
 import Button from 'cozy-ui/transpiled/react/Buttons'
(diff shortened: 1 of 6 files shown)


packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
```diff
@@ -185,6 +185,9 @@ const FederatedFolderModalContent = ({
   const existingRecipients = existingDocument
     ? getRecipients(existingDocument._id)
     : []
+  const displayedLink = existingDocument?.driveId
```
rezk2ll, line 156: why it's static if the document has a driveId?

doubleface: Good catch. getFederatedShareLink is actually synchronous
So the useState + useEffect wrapper is unnecessary here .
 I'll compute the federated link directly in displayedLink, just like getSharingLink
is used for non-federated documents, and remove the effect/state.

rezk2ll (approved):

PR #3024, opened 2026-06-10
fix: show shared drive recipients in viewer panel

doubleface:
- Add a cozy-sharing helper to derive recipients from an existing sharing.
- Use shared drive recipients in the viewer sharing panel.
- Add tests for shared drive recipients in cozy-sharing and cozy-viewer.

https://app.notion.com/p/linagora/5d2f7429646042378356ecb072890d35?v=1c062718bad180d3847f000cd4c97139&p=36862718bad1807b9145c6714ab3f0ba&pm=s

### DIFF
--- a/packages/cozy-sharing/src/state.js
+++ b/packages/cozy-sharing/src/state.js
@@ -338,6 +338,18 @@ export const getOwner = (state, docId) =>
 
 export const getRecipients = (state, docId) => {
   const sharings = getDocumentSharings(state, docId)
+  return getRecipientsFromSharings(sharings, docId)
+}
+
+export const getRecipientsFromSharing = (sharing, docId) => {
+  if (!sharing) {
+    return []
+  }
+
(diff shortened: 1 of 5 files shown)


packages/cozy-sharing/src/state.js
```diff
@@ -352,7 +364,7 @@ const getRecipientType = (sharing, member, documentType) => {
   }
 
   // Else if it is a shared drive, for sure it's two-way
-  if (sharing.drive) {
+  if (sharing.attributes?.drive || sharing.drive) {
```
zatteo: Why do we need to check both?

zatteo (approved):

packages/cozy-sharing/src/state.js
doubleface: It's not needed. The unit test is wrong. fixed it

PR #3028, opened 2026-06-10
fix(cozy-sharing): scroll members list in federated share dialogs

rezk2ll:
## Problem
The federated share dialogs keep overflow visible so the contact suggestions dropdown can pop over the dialog, which disables the dialog's native content scrolling. With many members the list grows past the viewport and pushes the Copy link and Share buttons off screen, with no way to reach them.

## Solution
Keep overflow visible for the dropdown, and pass the height constraint down the flex chain instead: the dialog content, its inner wrapper and the body become flex columns with min-height 0, and only the members list scrolls.

## demo
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Members list in sharing dialogs now scrolls independently while suggestion dropdowns and overlays remain visible, improving usability.

* **Style**
  * Dialog layout and styling updated for consistent sizing and overflow behavior, resulting in a cleaner presentation and more reliable access-management interactions.

### DIFF
--- a/packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
+++ b/packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
@@ -186,10 +186,12 @@ const FederatedFolderModalContent = ({
       title={modalTitle}
       classes={{ paper: 'u-ov-visible' }}
       componentsProps={{
-        dialogContent: { className: 'u-ov-visible' }
+        dialogContent: {
+          className: styles['share-dialog-scrollable-content']
+        }
       }}
       content={
-        <div>
+        <div className={styles['share-dialog-scrollable-body']}>
           <div className="u-ph-2">
             <AntivirusAlert
               document={isSending ? frozenDoc : existingDocument}
@@ -240,7 +242,7 @@ const FederatedFolderModalContent = ({
             recipients={isSending ? frozenRecipients : existingRecipients}
             document={isSending ? frozenDoc : existingDocument}
             documentType="Files"
-            className="u-w-100"
+            className={styles['share-dialog-members']}
             onRevoke={revoke}
             link={displayedLink}
           />
--- a/packages/cozy-sharing/src/styles/share.styl
+++ b/packages/cozy-sharing/src/styles/share.styl
@@ -2,6 +2,39 @@
 @require 'components/popover.styl'
 @require 'settings/breakpoints.styl'
 
+// The federated share dialogs keep overflow visible so the contact
+// suggestions dropdown can pop over the dialog, which disables the dialog's
+// native scrolling. These classes pass the height constraint down through
+// the flex chain so only the members list scrolls.
+.share-dialog-scrollable-content
+    display flex
+    flex-direction column
+    min-height 0
+    // keep the contact suggestions dropdown visible over the dialog;
+    // !important to win over MUI's DialogContent overflow-y auto
+    overflow visible !important // @stylint ignore
+
+    // cozy-ui dialogs expose no componentsProps hook for their
+    // .dialogContentInner wrapper, but the classname is a stable contract:
+    // cozy-ui's own MUI theme overrides style it by name
+    > :global(.dialogContentInner) // @stylint ignore Naming imposed by cozy-ui
+        display flex
+        flex-direction column
+        flex 1 1 auto
+        min-height 0
+
+.share-dialog-scrollable-body
+    display flex
+    flex-direction column
+    flex 1 1 auto
+    min-height 0
+
+.share-dialog-members
+    flex 1 1 auto
+    min-height 0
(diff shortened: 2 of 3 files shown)


packages/cozy-sharing/src/styles/share.styl
```diff
@@ -2,6 +2,34 @@
...
+    flex-direction column
+    min-height 0
+
+    // targets cozy-ui FixedDialog's .dialogContentInner wrapper, the only
+    // div between the DialogContent and the dialog body we render
+    > div
+        display flex
+        flex-direction column
+        flex 1 1 auto
+        min-height 0
```
zatteo, line 24: The `> div` is risky and can break without any notice with a cozy-ui/MUI update. There are no other componentsProps/classname to leverage?

rezk2ll: yes we now laverage dialogContentInner

JF-Cozy (commented):
CozyDialogs naturally add scrollers for long content https://docs.cozy.io/cozy-ui/react/#/CozyDialogs we should probably remove (a lot?) css instead of adding it 🤔

rezk2ll:
> CozyDialogs naturally add scrollers for long content https://docs.cozy.io/cozy-ui/react/#/CozyDialogs we should probably remove (a lot?) css instead of adding it 🤔

hmm :thinking: native dialog scrolling is disabled here on purpose: c7338a7ca, you added u-ov-visible on the paper and DialogContent so the contact suggestions dropdown is not clipped by the dialog. 

Removing the CSS brings the native scroller back, but reintroduces that clipping (tested: in a short dialog, the dropdown gets cut at the dialog edge and adds a phantom scrollbar).

The clean way to end up with less CSS is to portal the suggestions dropdown out of the dialog, then u-ov-visible and this PR's CSS can all go. That touches ShareRecipientsInput, which the classic ShareModal also uses, so I'd rather do it as a follow-up. Happy to open it.

JF-Cozy:
> so the contact suggestions dropdown is not clipped by the dialog

yes maybe we can address this issue by an other way

rezk2ll:
just to compare @JF-Cozy 

### without the css

image


### with the css in the PR
image

JF-Cozy:
> just to compare @JF-Cozy

yes I know I remember the issue

packages/cozy-sharing/src/styles/share.styl
```diff
@@ -2,6 +2,35 @@
...
+// The federated share dialogs keep overflow visible so the contact
+// suggestions dropdown can pop over the dialog, which disables the dialog's
+// native scrolling. These classes pass the height constraint down through
+// the flex chain so only the members list scrolls.
+.share-dialog-scrollable-content
+    display flex
+    flex-direction column
+    min-height 0
+
+    // cozy-ui dialogs expose no componentsProps hook for their
```
JF-Cozy, line 17: cozy-ui provides no hook at all for wrappers, useCozyDialogs is not meant to be used in app. But provides some props for class overrides, maybe we can use componentsProps.dialogContent... and as we can override classnames of dialogTitle and dialogContent, we can also update cozy-ui to override dialogContentInner if necessary. Or maybe even deal with this case in cozy-ui dialogs itself 🤔 

We tend to remove custom css as much as possible. Adding all this will, maybe, lead us to a more complicated thing to maintain and modify.

packages/cozy-sharing/src/styles/share.styl
```diff
@@ -2,6 +2,35 @@
 @require 'components/popover.styl'
 @require 'settings/breakpoints.styl'
```
JF-Cozy, line 3: nit: not relative to this PR but I think this one is useless

packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
```diff
@@ -186,10 +187,15 @@ const FederatedFolderModalContent = ({
       title={modalTitle}
       classes={{ paper: 'u-ov-visible' }}
       componentsProps={{
-        dialogContent: { className: 'u-ov-visible' }
+        dialogContent: {
+          className: cx(
+            'u-ov-visible',
+            styles['share-dialog-scrollable-content']
```
JF-Cozy: if we use a specific class to do thing, we can remove u-classes and use it css directly in the class itself

packages/cozy-sharing/src/styles/share.styl
rezk2ll: That's an upstream cozy-ui change; it can't land in this PR regardless. This PR keeps the CSS scoped to cozy-sharing and works with today's cozy-ui

JF-Cozy: I don't understand... 🤔

JF-Cozy (approved):

PR #3029, opened 2026-06-10
fix(cozy-realtime): keep drive sharings new on background connections

rezk2ll:
## Problem
The stack marks a drive sharing as seen whenever a recipient connects to its realtime websocket, which was meant for the user opening the drive view. The dataproxy search indexer connects to every accepted drive within seconds, so auto-accepted sharings lose their new status before the user ever sees the sharings counter in Drive.

## Solution
cozy-realtime gains a `background` option that appends `background=true` to the drive realtime URL, and the search indexer in cozy-dataproxy-lib passes it so the stack keeps the sharing unseen (stack side: cozy/cozy-stack#4793). Connections without the option behave as before, so the drive folder view still clears the counter when the user opens the drive.

## Worth knowing
cozy-dataproxy-lib needs the cozy-realtime release containing this option; its peer range should be bumped to that version when it is published.

### DIFF
--- a/packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
+++ b/packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
@@ -245,9 +245,12 @@ export class SearchEngine {
   }
 
   private addSharedDriveRealtime(sharedDriveId: string): void {
+    // background: the indexer watches the drive without the user looking at
+    // it, so the stack must not mark the sharing as seen
     const realtime = new CozyRealtime({
       client: this.client,
-      sharedDriveId
+      sharedDriveId,
+      background: true
     })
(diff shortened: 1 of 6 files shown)


packages/cozy-dataproxy-lib/src/search/SearchEngine.ts
```diff
@@ -245,9 +245,12 @@ export class SearchEngine {
   }
 
   private addSharedDriveRealtime(sharedDriveId: string): void {
+    // background: the indexer watches the drive without the user looking at
+    // it, so the stack must not mark the sharing as seen
```
zatteo, line 249: nit: Useless comment (already properly documented in CozyRealtime.js)

zatteo (approved):

doubleface (approved):

PR #3032, opened 2026-06-12
feat(cozy-sharing): Show shared drive links

doubleface:
https://app.notion.com/p/linagora/LINKS-on-reciver-side-doen-t-work-37c62718bad1801c8041e23d64eae181

Needs next cozy-stack-client with https://github.com/linagora/cozy-client/pull/1689

## Release Notes

* **New Features**
  * Added automatic link prefetching for shared drives to improve responsiveness when opening sharing dialogs.

* **Bug Fixes**
  * Improved shared drive link creation to automatically recover from conflicts by fetching and reusing existing sharing links instead of failing.

### DIFF
--- a/packages/cozy-sharing/src/SharingProvider.jsx
+++ b/packages/cozy-sharing/src/SharingProvider.jsx
@@ -440,8 +440,16 @@ export class SharingProvider extends Component {
     const resp = await drivePermissionCollection.findLinksByIds([documentId])
     const permissions = resp.data || []
 
-    if (permissions.length > 0) {
-      this.dispatch(addSharingLink(permissions))
+    const existingPermissions = getDocumentPermissions(this.state, documentId)
+    const existingPermissionIds = existingPermissions.map(
+      permission => permission.id
+    )
+    const newPermissions = permissions.filter(
+      permission => !existingPermissionIds.includes(permission.id)
+    )
+
+    if (newPermissions.length > 0) {
+      this.dispatch(addSharingLink(newPermissions))
     }
 
     return permissions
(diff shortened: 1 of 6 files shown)


packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
```diff
@@ -169,13 +174,37 @@ const FederatedFolderModalContent = ({
...
     : []
   const displayedLink = existingDocument?.driveId
     ? getFederatedShareLink(existingDocument)
     : getSharingLink(existingDocument?._id)
 
+  useEffect(() => {
+    if (!existingDocument?.driveId) return
+    if (!documentId) return
+    if (!fetchSharedDriveSharingLinks) return
+    if (documentPermissions.length > 0) return
```
rezk2ll, line 193: If this check fails somehow, we will end up in an infinite loop.

If findLinksByIds([documentId]) returns a permission whose values do not contain documentId, the indexPermission never indexes it under documentId, so getDocumentPermissions(documentId) stays length 0. Because dispatch → setState → re-render (SharingProvider.jsx), the effect runs again, refetches, dispatches, and re-renders forever.

Also we should gate the effect on a fetching/loading flag

packages/cozy-sharing/src/SharingProvider.jsx
```diff
@@ -413,6 +427,26 @@ export class SharingProvider extends Component {
...
+      driveId: document.driveId
+    })
+    const documentId = document._id || document.id
+    if (!documentId) return []
+
+    const resp = await drivePermissionCollection.findLinksByIds([documentId])
+    const permissions = resp.data || []
+
+    if (permissions.length > 0) {
+      this.dispatch(addSharingLink(permissions))
```
rezk2ll: This will produce duplicates if we don't check the existing permissions.

Failure scenario: user opens the modal (effect loads the link), then clicks generate → backend replies 409 → recovery refetches and re-dispatches → duplicate link/recipient rows render.

doubleface: done

rezk2ll (approved):

PR #3036, opened 2026-06-16
fix(sharing): Update member permissions optimistically

doubleface:
Do not wait for realtime to update the change of rights on recipient. Do it optimistically. If any error occur, the state will be reverted by realtime update

https://app.notion.com/p/linagora/Recipient-Change-role-long-response-38162718bad18044b8dcd528c85b215f

## Release Notes

* **New Features**
  * Member permissions now update instantly in the UI while synchronizing with the backend, with automatic rollback if synchronization fails.

* **Bug Fixes**
  * Fixed the permission type selection menu so it closes reliably right when the permission type actually changes.

* **Tests**
  * Updated sharing permission change tests to verify optimistic UI updates and rollback behavior under failure and realtime update scenarios.

### DIFF
--- a/packages/cozy-sharing/src/SharingProvider.jsx
+++ b/packages/cozy-sharing/src/SharingProvider.jsx
@@ -383,7 +383,23 @@ export class SharingProvider extends Component {
         await this.sharingCol.setReadWrite(sharing, memberIndex)
       }
     } catch (error) {
-      this.dispatch(updateSharing(sharing))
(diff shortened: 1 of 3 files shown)


zatteo (approved):

zatteo:
Where was it so slow? Because realtime is usually quite fast.

doubleface:
> Where was it so slow? Because realtime is usually quite fast.

@zatteo Looks like this specific change (from recipient side with write permissions, is quite long

PR #3038, opened 2026-06-17
fix(cozy-sharing): Prioritize email over instance URL

doubleface:
Remove all icons from MemberRecipientStatus and display email instead of instance URL for sharing recipients

https://app.notion.com/p/linagora/Sharing-modal-Don-t-display-cloud-icon-and-display-email-38162718bad180d6a979cb34bc56c313

## Release Notes

* **New Features**
  * Sharing recipient lists now show recipient email and name details in their status text.

* **Bug Fixes**
  * Recipient email addresses now display correctly across recipient list views.

* **Refactor**
  * Simplified recipient status rendering to a single text element.
  * Updated recipient status logic to localize invitation states more consistently and handle missing identity fields.

### DIFF
--- a/packages/cozy-sharing/src/components/Recipient/MemberRecipientStatus.jsx
+++ b/packages/cozy-sharing/src/components/Recipient/MemberRecipientStatus.jsx
@@ -9,6 +9,7 @@ const MemberRecipientStatus = ({
   isMe,
   instance,
   email,
+  name,
   typographyProps
 }) => {
   const { t } = useI18n()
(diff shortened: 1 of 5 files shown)


packages/cozy-sharing/src/components/Recipient/MemberRecipientStatus.jsx
```diff
@@ -1,42 +1,45 @@
...
   const { t } = useI18n()
 
   const isSendingEmail = !isMe && status === 'mail-not-sent'
   const isReady = isMe || status === 'ready' || status === 'owner'
-  let text, icon
+  let text
   if (isReady) {
-    text = instance
-    icon = ToTheCloudIcon
+    text = email || instance
```
rezk2ll: for members without a Displayname, this will render the email twice for that member ( primary and secondary texts )

rezk2ll (commented):
we should have unit tests for the new behaviour

zatteo (approved):

packages/cozy-sharing/src/components/Recipient/MemberRecipientStatus.jsx
doubleface: Now when primary text is email, we hide the secondary text

rezk2ll (approved):

PR #3040, opened 2026-06-18
fix(cozy-viewer): Show inherited sharing recipients

doubleface:
Files inside shared folder but fetched with regular file api don't have a driveId attribute

When a shared parent path is known, fetch the corresponding folder by path and use its id as the sharing reference for recipients, permissions, links, and ownership.

https://app.notion.com/p/linagora/Who-has-access-seciton-in-viewer-is-wrong-36862718bad1807b9145c6714ab3f0ba?source=copy_link

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed sharing panel to correctly display permissions and recipients for files located within shared folders, improving reliability for nested file sharing scenarios.

### DIFF
--- a/packages/cozy-viewer/src/Panel/Sharing.jsx
+++ b/packages/cozy-viewer/src/Panel/Sharing.jsx
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 
-import { useClient } from 'cozy-client'
+import { useClient, useQuery } from 'cozy-client'
 import flag from 'cozy-flags'
 import {
   getRecipientsFromSharing,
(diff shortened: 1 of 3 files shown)


packages/cozy-viewer/src/Panel/Sharing.jsx
```diff
@@ -31,19 +32,38 @@ const Sharing = ({ file, t }) => {
...
+    useQuery(sharedParentQuery.definition, {
+      ...sharedParentQuery.options,
+      enabled: !!sharedParentPath
+    })
+  const sharingReferenceId =
+    !sharedDriveSharing && sharedParentPath && sharedParentFolders?.[0]?._id
+      ? sharedParentFolders[0]._id
+      : file._id
+
+  if (sharedParentPath && sharedParentFetchStatus === 'loading') return null
```
rezk2ll: should we add a check for `pending` state no?

doubleface: Right, let's even display it only with the `loaded` state

zatteo (approved):

PR #3041, opened 2026-06-22
fix(cozy-dataproxy-lib): exclude shared-drive shortcuts from search

rezk2ll:
## Context
When a recipient receives a shared drive, the stack drops a `<DriveName>.url` shortcut directly under the Shared Drives directory. These placeholders were indexed into global search, so searching a drive name returned `<DriveName>.url` junk instead of (or alongside) the drive content.

## Solution
`shouldKeepFile` only excluded the Shared Drives folder itself (by `_id`), not the shortcuts inside it. It now also drops any file whose `dir_id` is the Shared Drives directory. Real drive files are unaffected since they live under their own folders, never directly in that directory.

### DIFF
--- a/packages/cozy-dataproxy-lib/src/search/helpers/normalizeFile.ts
+++ b/packages/cozy-dataproxy-lib/src/search/helpers/normalizeFile.ts
@@ -36,6 +36,14 @@ export const shouldKeepFile = (file: IOCozyFile): boolean => {
   // Shared drives folder to be hidden in search.
   // The files inside it though must appear. Thus only the file with the folder ID is filtered out.
   const notSharedDrivesDir = file._id !== SHARED_DRIVES_DIR_ID
+  // The shortcut placeholders sitting directly under the Shared Drives
+  // directory represent the drive itself, not its content; they must not appear
+  // in search.
+  const notSharedDriveShortcut = !(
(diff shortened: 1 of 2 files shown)


zatteo (approved):

packages/cozy-dataproxy-lib/src/search/helpers/normalizeFile.ts
```diff
@@ -36,6 +36,12 @@ export const shouldKeepFile = (file: IOCozyFile): boolean => {
   // Shared drives folder to be hidden in search.
   // The files inside it though must appear. Thus only the file with the folder ID is filtered out.
   const notSharedDrivesDir = file._id !== SHARED_DRIVES_DIR_ID
+  // The shortcut placeholders sitting directly under the Shared Drives
+  // directory represent the drive itself, not its content; they must not appear
+  // in search.
+  const notSharedDriveShortcut = file.dir_id !== SHARED_DRIVES_DIR_ID
```
doubleface: we do not check if it is a shortcut or not here

zatteo: Agree that it could be safer to check

rezk2ll: agreed, i changed the condition

Crash-- (approved):

PR #3054, opened 2026-06-24
fix(federated): Show correct recipients for files inside shared folder

doubleface:
For the recipient case (document has driveId), use getRecipientsFromSharing on the shared drive sharing directly.

For the owner case (file inside shared folder but no driveId), fetch the parent folder via statByPath from its path, then use its _id for getRecipients.

https://app.notion.com/p/linagora/Recipient-Share-subfolder-or-file-in-the-shared-foder-Wrong-role-38862718bad180e091cfd1593883e184?source=copy_link

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shared folder recipient handling so sharing details are now shown correctly even when the drive context is unavailable.
  * Existing recipients are now derived more reliably based on the folder path or sharing context, reducing incorrect recipient lists in the sharing modal.

* **Tests**
  * Updated coverage to match the latest sharing-context behavior.

### DIFF
--- a/packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
+++ b/packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
@@ -16,6 +16,7 @@ import { getOrCreateFromArray } from '../../helpers/contacts'
 import withLocales from '../../hoc/withLocales'
 import { usePendingRecipients } from '../../hooks/usePendingRecipients'
 import { useSharingContext } from '../../hooks/useSharingContext'
+import { getRecipientsFromSharing } from '../../state'
 import styles from '../../styles/share.styl'
 import AntivirusAlert from '../AntivirusAlert'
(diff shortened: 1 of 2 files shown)


packages/cozy-sharing/src/components/FederatedFolder/FederatedFolderModal.jsx
```diff
@@ -178,7 +180,40 @@ const FederatedFolderModalContent = ({
...
+    !sharedDriveSharing && sharedParentFolder?._id
+      ? sharedParentFolder._id
+      : documentId
+
+  const existingRecipients =
+    sharedDriveSharing?.attributes?.members && documentId
+      ? getRecipientsFromSharing(sharedDriveSharing, documentId)
+      : sharingReferenceId
+        ? getRecipients(sharingReferenceId)
+        : []
```
zatteo, line 216: Did you write the same code recently somewhere else?

zatteo (approved):

PR #3066, opened 2026-06-30
Fix changelogs and deps

JF-Cozy:
We previously upgraded cozy-ui to 140+ in cozy-devtools. It was a mistake, no need to deal with that yet. In fact, at this stage, no libs does. To ease the migration, we downgrade here to avoid forcing the apply of the 140.0.0 BC

For the changelogs, it was a "bug" because we did a BC on devtools and add a sharing file in the same commit.


zatteo (approved):
❤️

PR #3072, opened 2026-07-07
feat: Add menu to modify group permission

zatteo:
image

## Summary by CodeRabbit

* **New Features**
  * Group sharing permissions can now be changed directly, with immediate updates in the interface.
* **Bug Fixes**
  * Permission changes now show an error message if the update fails.
  * Failed group permission updates are rolled back safely to keep sharing state consistent.
* **Refactor**
  * Improved permission-change flow for both group and member recipients, with cleaner menu handling and fewer unnecessary updates.


packages/cozy-sharing/src/components/Recipient/GroupRecipientPermissions.jsx
```diff
@@ -52,10 +59,32 @@ const GroupRecipientPermissions = ({
 
   const type = read_only ? 'one-way' : 'two-way'
 
-  const actions = makeActions([permission], {
-    t,
-    type
-  })
+  const setType = async newType => {
```
JF-Cozy, line 62: I think `if (newType === type)` part can be optimized (we hideMenu anyway, the only condition is to trigger the try/catch or no).

Besides, it seems to me we have a part of the action here, and in the makeActions part 🤔 it's not very clear. Do you think we can optimize a bit to gather action logic in only one place?

packages/cozy-sharing/src/SharingProvider.jsx
```diff
@@ -412,6 +413,73 @@ export class SharingProvider extends Component {
     }
   }
 
+  /**
+   * Update a sharing group's read_only field to change permission type
+   * @param {string} sharingId - The ID of the sharing
+   * @param {number} group - The index of the group in the sharing groups array
+   * @param {string} newType - 'one-way' for read-only, 'two-way' for read-write
+   */
+  updateSharingGroupType = async (sharingId, groupIndex, newType) => {
```
JF-Cozy, line 422: is it ok to disptach something... and then if it doesn't work, we dispatch a rollback? 🤔 sounds strange...

packages/cozy-sharing/src/components/Recipient/GroupRecipientPermissions.jsx
zatteo: I refactored it a bit to be simplier. What do you mean by _we have a part of the action here, and in the makeActions part 🤔_ ?

packages/cozy-sharing/src/SharingProvider.jsx
zatteo: It is usual optimistic update? I took the logic from the update member logic for the optimistic update here.

packages/cozy-sharing/src/components/Recipient/GroupRecipientPermissions.jsx
zatteo: I understand. Currently setReadOnlySharedPermission and setReadWriteSharedPermission are generic and are used to set permissions for members, groups and links. They only carry the UI of the action as you said.

JF-Cozy (approved):

PR #3098, opened 2026-07-24
feat(cozy-ui-plus): Support forced themes 

### LINKED ISSUE #4097: Allow File Picker caller to force theme in it

doubleface:
Allow intent callers to request light or dark through intent data. This
keeps the host dialog, loading surface, and service iframe visually
consistent.

Keep auto, invalid, and omitted values on the caller theme. Give the
themed iframe wrapper explicit dimensions instead of relying on the
caller's u-dc utility, which could collapse the iframe to its intrinsic
300x150 size.

## Summary by CodeRabbit

* **New Features**
  * Added theme overrides for intent dialogs and embedded intent views.
  * Supports forcing light or dark themes for loading surfaces, dialogs, and close buttons.
  * Invalid, missing, or automatic theme values preserve the caller’s theme.

* **Documentation**
  * Updated usage examples and guidance for configuring intent themes.

* **Tests**
  * Extended intent view test setup to support theme configuration.

### DIFF
--- a/packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
+++ b/packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
@@ -49,8 +49,9 @@ const IntentDialogOpener = props => {
   }
 
   const Tag = tag // React needs uppercase element names
-  const forcedTheme = ['light', 'dark'].includes(options?.theme)
-    ? options.theme
+  const themeType = options?.theme?.type
+  const forcedTheme = ['light', 'dark'].includes(themeType)
+    ? themeType
     : undefined
   const elements = [
     React.cloneElement(children, { key: 'opener', onClick: openModal })
@@ -101,7 +102,9 @@ IntentDialogOpener.propTypes = {
   action: PropTypes.string.isRequired,
   /** Options passed to the intent */
   options: PropTypes.shape({
-    theme: PropTypes.oneOf(['light', 'dark', 'auto'])
+    theme: PropTypes.shape({
+      type: PropTypes.oneOf(['light', 'dark'])
+    })
   }),
   /** Doctype on which you want to execute the action */
   doctype: PropTypes.string.isRequired,
--- a/packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
+++ b/packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
@@ -3,6 +3,7 @@ import React, { useState } from 'react'
 
 import { DialogCloseButton } from 'cozy-ui/transpiled/react/CozyDialogs'
 import Dialog from 'cozy-ui/transpiled/react/Dialog'
+import CozyTheme from 'cozy-ui/transpiled/react/providers/CozyTheme'
 
 import IntentIframe, { iframeProps } from '../IntentIframe'
 
@@ -48,33 +49,37 @@ const IntentDialogOpener = props => {
   }
 
   const Tag = tag // React needs uppercase element names
+  const forcedTheme = ['light', 'dark'].includes(options?.theme)
+    ? options.theme
+    : undefined
   const elements = [
     React.cloneElement(children, { key: 'opener', onClick: openModal })
   ]
 
   if (modalOpened) {
     elements.push(
-      <Component
-        key="intent-modal"
-        open={modalOpened}
-        onClose={closable && handleDismiss}
-        {...componentProps}
-      >
-        {closable && showCloseButton && (
-          <DialogCloseButton onClick={handleDismiss} />
-        )}
-        <IntentIframe
-          action={action}
-          type={doctype}
-          data={options}
-          create={create}
-          onCancel={handleDismiss}
-          onTerminate={handleComplete}
-          onReadyToUse={onReadyToUse}
-          waitForReadyToUse={waitForReadyToUse}
-          iframeProps={iframeProps}
-        />
-      </Component>
+      <CozyTheme key="intent-modal" type={forcedTheme}>
+        <Component
+          open={modalOpened}
+          onClose={closable && handleDismiss}
+          {...componentProps}
+        >
+          {closable && showCloseButton && (
+            <DialogCloseButton onClick={handleDismiss} />
+          )}
+          <IntentIframe
+            action={action}
+            type={doctype}
+            data={options}
+            create={create}
+            onCancel={handleDismiss}
+            onTerminate={handleComplete}
+            onReadyToUse={onReadyToUse}
+            waitForReadyToUse={waitForReadyToUse}
+            iframeProps={iframeProps}
+          />
+        </Component>
+      </CozyTheme>
     )
   }
 
@@ -94,6 +99,10 @@ IntentDialogOpener.propTypes = {
   waitForReadyToUse: PropTypes.bool,
   /** Action you want to execute */
   action: PropTypes.string.isRequired,
+  /** Options passed to the intent */
+  options: PropTypes.shape({
+    theme: PropTypes.oneOf(['light', 'dark', 'auto'])
+  }),
   /** Doctype on which you want to execute the action */
   doctype: PropTypes.string.isRequired,
   /** Whether the dialog is closable by itself (not by the Intent) with a button or by clicking outside of it */
--- a/packages/cozy-ui-plus/src/Intent/IntentIframe/Readme.md
+++ b/packages/cozy-ui-plus/src/Intent/IntentIframe/Readme.md
@@ -14,7 +14,7 @@ import utils from '../../docs/utils'
 <IntentIframe
   action="PICK"
   create={utils.fakeIntentCreate}
-  data={{folder: '840af7e7be6c41bed3c5b76d03f66328', theme: 'dark'}}
+  data={{folder: '840af7e7be6c41bed3c5b76d03f66328', theme: { type: 'dark' }}}
   type="io.cozy.files"
   onCancel={() => alert('intent cancelled')}
   onError={(error) => alert('intent has failed with error: ' + error.message)}
@@ -27,9 +27,9 @@ Set `waitForReadyToUse` to keep the intent hidden behind the loading spinner
 until its service sends the `readyToUse` signal. This option is disabled by
 default because older intent services do not send this signal.
 
-Set `data.theme` to `light` or `dark` to force that theme on the loading
-surface and pass it to the intent service. With `auto`, an invalid value or no
-value, the loading surface keeps the caller's theme.
+Set `data.theme.type` to `light` or `dark` to force that theme on the
+loading surface and pass it to the intent service. Any other value or no value
+keeps the caller's theme.
 
 ### Within an application side controled Dialog
 
--- a/packages/cozy-ui-plus/src/Intent/IntentIframe/Readme.md
+++ b/packages/cozy-ui-plus/src/Intent/IntentIframe/Readme.md
@@ -14,7 +14,7 @@ import utils from '../../docs/utils'
 <IntentIframe
   action="PICK"
   create={utils.fakeIntentCreate}
-  data={{folder: '840af7e7be6c41bed3c5b76d03f66328'}}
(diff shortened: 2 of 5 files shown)


zatteo:
It won't help an external app like tmail that do not use cozy-ui-plus to force a theme no? We need to pass the "theme": ... through postMessage to the service?

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
```diff
@@ -70,10 +75,21 @@ const IntentDialogOpener = props => {
           onCancel={handleDismiss}
           onTerminate={handleComplete}
           onReadyToUse={onReadyToUse}
+          waitForReadyToUse={waitForReadyToUse}
           iframeProps={iframeProps}
         />
       </Component>
     )
+
+    elements.push(
```
JF-Cozy, line 61: we can simplify and always use CozyTheme, just set type to undefined if no forcedTheme

JF-Cozy (commented):
Why do we have to add CozyTheme with forceTheme? Why, and for which apps, it's not working out of the box right now? 🤔

doubleface:
> It won't help an external app like tmail that do not use cozy-ui-plus to force a theme no? We need to pass the "theme": ... through postMessage to the service?

Yes but the file picker uses cozy-ui-plus. Here is the draft PR which will use this change : https://github.com/linagora/twake-drive/pull/4100

doubleface:
> Why do we have to add CozyTheme with forceTheme? Why, and for which apps, it's not working out of the box right now? 🤔

We need to allow mail application, which does not use cozy theme, to force the theme to match its own

JF-Cozy:
> > Why do we have to add CozyTheme with forceTheme? Why, and for which apps, it's not working out of the box right now? 🤔
> 
> We need to allow mail application, which does not use cozy theme, to force the theme to match its own

To be more precise, "it's working" out of the box right now, I mean we have light/dark theme in the intent according to the Twake settings. So no need to wrap anything here for this purpose.

But Mail app doesn't have dark theme, so in this configuration, we want to force the theme to be light in the intent to avoid having light app with dark intent. So here it's to force the light interface of the intent, meaning the loaders and modal stuff (close button etc.), isn't it?

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
```diff
@@ -48,12 +49,15 @@ const IntentDialogOpener = props => {
...
+  const forcedTheme = ['light', 'dark'].includes(options?.theme)
+    ? options.theme
+    : undefined
   const elements = [
     React.cloneElement(children, { key: 'opener', onClick: openModal })
   ]
 
   if (modalOpened) {
-    elements.push(
+    const intentModal = (
```
JF-Cozy: why creating intentModal and not just wrapping Component with CozyTheme?

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
```diff
@@ -48,12 +49,15 @@ const IntentDialogOpener = props => {
   }
 
   const Tag = tag // React needs uppercase element names
+  const forcedTheme = ['light', 'dark'].includes(options?.theme)
```
JF-Cozy: we can add this new prop in propTypes with shape prop

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.spec.jsx
```diff
@@ -3,12 +3,21 @@ import React from 'react'
 
 import IntentDialogOpener from './index'
 
+jest.mock('cozy-ui/transpiled/react/providers/CozyTheme', () => ({
```
JF-Cozy: too hard to test without mocking it? I don't think this test is still relevant with mocking everything, finally we just test if React pass a prop correctly... 🤷‍♂️

packages/cozy-ui-plus/src/Intent/IntentIframe/index.jsx
```diff
@@ -104,6 +107,18 @@ class IntentIframe extends React.Component {
         {/* intent iframe will be appended here */}
       </div>
     )
+
+    return forcedTheme ? (
```
JF-Cozy: same here, we can use always CozyTheme, around the div above, no need to create intentContainer

packages/cozy-ui-plus/src/Intent/IntentIframe/Readme.md
```diff
@@ -14,7 +14,7 @@ import utils from '../../docs/utils'
 <IntentIframe
   action="PICK"
   create={utils.fakeIntentCreate}
-  data={{folder: '840af7e7be6c41bed3c5b76d03f66328'}}
+  data={{folder: '840af7e7be6c41bed3c5b76d03f66328', theme: 'dark'}}
```
JF-Cozy: maybe we should rely on the cozy-ui api, so `theme: {type: 'dark'}` instead (I don't have strong good feeling about this 🤔 )

packages/cozy-ui-plus/src/Intent/IntentIframe/styles.styl
```diff
@@ -3,6 +3,7 @@ iframe
     height 100%
     border 0
 
+.intentContainerTheme,
```
JF-Cozy: still useful ?

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
doubleface: Right, intentModal was not really needed here

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.spec.jsx
doubleface: Right, I remove these tests

packages/cozy-ui-plus/src/Intent/IntentIframe/styles.styl
doubleface: Yes, it is still needed for forced themes. When ignoreItself is false, the CozyTheme wrapper participates in the layout and needs explicit width: 100% and height: 100%. It is not applied for auto, where ignoreItself remains true.

packages/cozy-ui-plus/src/Intent/IntentIframe/index.spec.jsx
```diff
@@ -3,11 +3,25 @@ import React from 'react'
 
 import IntentIframe from './index'
 
+jest.mock('cozy-ui/transpiled/react/providers/CozyTheme', () => ({
```
JF-Cozy: same question about test as before 🤔

packages/cozy-ui-plus/src/Intent/IntentIframe/index.jsx
```diff
@@ -79,30 +80,38 @@ class IntentIframe extends React.Component {
...
-          </div>
-        )}
-        {error && (
-          <div className={styles.intentContainer__error}>{error.message}</div>
-        )}
-        {/* intent iframe will be appended here */}
-      </div>
+        <div
+          ref={intentViewer => (this.intentViewer = intentViewer)}
+          className={styles.intentContainer}
```
JF-Cozy, line 97: we now have a wrapper above, is this style still relevant? Can we just use styles.intentContainer on CozyTheme and remove styles.intentContainerTheme?

packages/cozy-ui-plus/src/Intent/IntentIframe/index.jsx
```diff
@@ -79,30 +80,38 @@ class IntentIframe extends React.Component {
...
-        ref={intentViewer => (this.intentViewer = intentViewer)}
-        className={styles.intentContainer}
-        aria-busy={loading}
-        data-iframe-loaded={frameLoaded}
-        data-waiting-for-ready-to-use={waitForReadyToUse && !readyToUse}
-        {...get(iframeProps, 'wrapperProps')}
+      <CozyTheme
+        type={forcedTheme}
+        ignoreItself={forcedTheme === null}
+        className={forcedTheme ? styles.intentContainerTheme : null}
```
JF-Cozy: why a condition here? Even with no forced theme, we still have a wrapper, and so a div in the DOM

packages/cozy-ui-plus/src/Intent/IntentIframe/Readme.md
doubleface: If it's more coherent with cozy-ui, let's do it

packages/cozy-ui-plus/src/Intent/IntentIframe/index.jsx
doubleface: It was a layout problem but I understand better now. I removed it

packages/cozy-ui-plus/src/Intent/IntentIframe/index.jsx
doubleface: Right, done

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
```diff
@@ -94,6 +100,12 @@ IntentDialogOpener.propTypes = {
   waitForReadyToUse: PropTypes.bool,
   /** Action you want to execute */
   action: PropTypes.string.isRequired,
+  /** Options passed to the intent */
+  options: PropTypes.shape({
+    theme: PropTypes.shape({
+      type: PropTypes.oneOf(['light', 'dark', 'auto'])
```
JF-Cozy: you sure auto is a viable option?

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.spec.jsx
JF-Cozy, line 1: so no need to touch anything here, finally, right?

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
doubleface: Why would it not be viable ? You would prefer an empty value in this case ?

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.spec.jsx
doubleface: Almost nothing

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.jsx
JF-Cozy: it's not supported by cozy-ui, and as we do ['light', 'dark'].includes(type) everywhere, using auto is equal to use null or undefined, isn't it? 🤔

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.spec.jsx
JF-Cozy: Ok let me rephrase: why do we need to change CozyDialogs and Dialog mocks here?

packages/cozy-ui-plus/src/Intent/IntentIframe/styles.styl
JF-Cozy: We past that, so no need for this anymore. I close the comment 👍

packages/cozy-ui-plus/src/Intent/IntentDialogOpener/index.spec.jsx
doubleface: Ok, we don't need it anymore. Removed it

JF-Cozy (approved):
I approve to unlock.

For the `auto` forceTheme value, I just saw we have `type: PropTypes.oneOf(['light', 'dark', 'auto'])` in cozy-ui... BUT here it's to force something, so it's not relevant (it won't force anything with auto value), I think we should remove the `auto` value in the doc

doubleface:
I removed auto and we have the same effect with empty value. I also think it is better like that

PR #3101, opened 2026-07-28
Add guideline to build an intent

zatteo:
Explaining how we should handle:
- UI lifecycle
- errors

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for building intents, including client-led UI lifecycle management and display, closing, and hiding behavior.
  * Clarified how to handle business and technical errors.
  * Documented the recommended `ready` message timeout and fallback behavior when it is not received.

### DIFF
--- a/packages/cozy-interapp/README.md
+++ b/packages/cozy-interapp/README.md
@@ -30,6 +30,17 @@ See [cozy-ui-plus](https://github.com/linagora/cozy-libs/tree/master/packages/co
 
 ## Guideline to build an intent
 
+### UI lifecycle
+
+The client app is responsible of the UI lifecycle of the intent, e.g.:
+- opening the intent
+- hiding the intent
+- closing the intent
+
+As the service app can not know how client application will display the service, we let the client:
+- choose how to display the intent: inside a modal, inside a bottomshet, inside a card, etc
+- choose how the user can close it (with a cross button, with a keyboard shortcut, etc) or hide it (with a minimize button, etc), etc
+
 ### Error handling
 
 The rule:
--- a/packages/cozy-interapp/README.md
+++ b/packages/cozy-interapp/README.md
@@ -28,6 +28,16 @@ const intents = new Intents({ client })
 
 See [cozy-ui-plus](https://github.com/linagora/cozy-libs/tree/master/packages/cozy-ui-plus/src/Intent) for ready for use React components using cozy-interapp.
 
+## Guideline to build an intent
+
+### Error handling
+
+The rule:
+- all business errors are managed inside the intent. e.g. with a File Picker: a user selects a file but when the user click on "Share" the file has been deleted => the File Picker manages itself the error. It displays an error message and let the user try again or select another file or close the File Picker if it does not make sense anymore, etc. There is not interest to forward this kind of errors to the client app because it will not know what to do with these business errors.
+- technical errors are forwarded to the client in an `error` message. The client app can decide what it does: try again, display an error message saying the service is unavailable, ...
+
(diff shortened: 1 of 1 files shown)


packages/cozy-interapp/README.md
```diff
@@ -28,6 +28,27 @@ const intents = new Intents({ client })
...
+### UI lifecycle
+
+The client app is responsible of the UI lifecycle of the intent, e.g.:
+- opening the intent
+- hiding the intent
+- closing the intent
+
+As the service app can not know how client application will display the service, we let the client:
+- choose how to display the intent: inside a modal, inside a bottomshet, inside a card, etc
+- choose how the user can close it (with a cross button, with a keyboard shortcut, etc) or hide it (with a minimize button, etc), etc
```
zatteo, line 68: This means that this button must be inserted by Twake Mail/Twake Calendar/etc. If we want to display the cross in the same row than "Twake Drive", we can potentially formalize in the File Picker ADR that a blank space stays available at the top right corner. Or we display the cross somewhere else.

Image

shepilov: why the buttton should be inserted and not just option - should we disaply it or not with callback event if needed?

packages/cozy-interapp/README.md
```diff
@@ -28,6 +28,27 @@ const intents = new Intents({ client })
...
+
+### Error handling
+
+The rule:
+- all business errors are managed inside the intent. e.g. with a File Picker: a user selects a file but when the user click on "Share" the file has been deleted => the File Picker manages itself the error. It displays an error message and let the user try again or select another file or close the File Picker if it does not make sense anymore, etc. There is not interest to forward this kind of errors to the client app because it will not know what to do with these business errors.
```
shepilov, line 74: what is the defference between budiness and technical errors?
Everything that intent is know how to deal with is business?

zatteo: I don't know how to explain it more clearly for now.

>  Everything that intent is know how to deal with is business? 

Yeah maybe it's a good explanation. For a File Picker, every errors related to files.

packages/cozy-interapp/README.md
zatteo: Yes, it is not an obligation to insert it. But if the client app want a close button, it adds it.

packages/cozy-interapp/README.md
```diff
@@ -52,6 +52,30 @@ const intents = new Intents({ fetch: fetchJSON })
...
+
+### UI lifecycle
+
+The client app is responsible of the UI lifecycle of the intent, e.g.:
+
+- opening the intent
+- hiding the intent
+- closing the intent
+
+As the service app can not know how client application will display the service, we let the client:
```
zatteo, line 65: I think that's why we should not round intent corners inside the iframe. As I wrote below, for a given intent, we do not know:
- how the client will display it: it is not necessary a modal, or it can be a full screen modal
- what is the UI of the client: for example, on La Suite Visio, the modal may have a different border radius than on a genuine Twake app

@Crash--

Crash--: Since we already let the client app handle the close button and thing like that, ok, I'm on your side.

zatteo: Now I am unsure about this. If we let the app handle the close button, every app must implement this close button and position it correctly. 

> inside a modal, inside a bottomshet, inside a card, etc

Is it realistic? Or are we never gonna need this stuff in real life?

Same about 

> choose how the user can close it (with a cross button, with a keyboard shortcut, etc) or hide it (with a minimize button, etc), etc

PR #3110, opened 2026-08-03
fix(cozy-search): Add fallback message for empty assistant replies

ewan102:
When the RAG backend returns an empty or citation-only answer, display a translated fallback message instead of a blank bubble. Applied consistently across the live streaming path, conversation history reload, and the conversation list preview.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a localized fallback message when assistant responses are empty or contain only sanitized content.
  * Conversation previews now display the fallback message when no usable assistant response is available.
  * Improved handling of streamed responses to prevent blank answers.

* **Localization**
  * Added empty-response translations in English, French, Russian, and Vietnamese.


packages/cozy-search/src/components/adapters/CozyRealtimeChatAdapter.ts
```diff
@@ -129,8 +129,15 @@ export const createCozyRealtimeChatAdapter = (
       if (!wasAborted) {
         const finalText = sanitizeChatContent(fullText)
         const sources = streamBridge.getSources(conversationId)
+        if (!finalText.trim()) {
+          log.info(
+            `Empty content was received for conversation ${conversationId}, using fallback message`
+          )
+        }
```
zatteo: If you really want to log when the fallback message is displayed, you should do it directly in `withEmptyResponseFallback` function to avoid checking 2x with `trim()` no?

packages/cozy-search/src/components/CozyAssistantRuntimeProvider.tsx
```diff
@@ -73,19 +73,28 @@ interface CozyAssistantRuntimeProviderProps {
...
-      msg.role === 'assistant' && msg.sources
-        ? { custom: { sources: msg.sources } }
-        : undefined
-  }))
+  return messages.map((msg, idx) => {
+    const sanitized = sanitizeChatContent(msg.content)
+    const content =
+      msg.role === 'assistant'
+        ? withEmptyResponseFallback(sanitized, t)
+        : sanitized
```
zatteo: It seems that we have multiple time this pattern. Maybe we should create a `formatAnswer` function that does everything? cc @paultranvan

paultranvan: sounds like a good idea

packages/cozy-search/src/components/adapters/CozyRealtimeChatAdapter.ts
paultranvan: I'm not sure it's worth having this log, tough

packages/cozy-search/src/components/helpers.js
```diff
@@ -85,8 +98,11 @@ export const getNameOfConversation = conversation => {
  * Since we don't have rule for description of the conversation
  * So temporary we get the last answer from assistant as description of the conversation
  */
-export const getDescriptionOfConversation = conversation => {
-  const content =
-    conversation.messages?.[conversation.messages.length - 1]?.content
-  return content && sanitizeChatContent(content)
+export const getDescriptionOfConversation = (conversation, t) => {
+  const lastMessage = conversation.messages?.[conversation.messages.length - 1]
```
paultranvan: we should have a first check on conversation in case it's null / undefined

paultranvan (commented):
I'm thinking of an edge case that I'm unsure of the behaviour: 
- user sends query
- the user stops the generation before the first token is received
- is the fallback displayed in this case? I think it shouldn't
Can you test this case @ewan102 ?

packages/cozy-search/src/components/adapters/CozyRealtimeChatAdapter.ts
ewan102: Ok so I get rid of the log or do I keep it and put it in with EmptyResponseFallback ?

ewan102:
Tested @paultranvan 's case, it does not put the fallback answer in that case. Got rid of the log info

packages/cozy-search/src/components/helpers.js
paultranvan: @ewan102 Thank you for the check ; note you could have just added a `?` rather than a `if(!conversation)` : 
`  const lastMessage = conversation?.messages?.[conversation.messages.length - 1]`

packages/cozy-search/src/components/adapters/CozyRealtimeChatAdapter.ts
paultranvan: You can remove it

